Issue #56, opened 2022-10-24
Documentation website setup based on 11ty


DanielleRameau:
Setup documentation website setup based on 11ty.

- [ ] configuration
- [ ] setup structure

anna-lach:
Done by #63

PR #63, opened 2022-10-28
Documentation setup

anna-lach:
Documentation setup based on 11ty with basic content structure, example styles etc.

#56


docs/website/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@sanomalearning/slds-docs",
```
jpzwarte, line 2: This should probably also have `"private": true`? Because the module won't be published?

docs/website/.eslintrc.cjs
```diff
@@ -0,0 +1,25 @@
+module.exports = {
```
jpzwarte, line 1: Do we need a new config? Or can we extend the one in the root of the repo?

docs/website/.eleventy.cjs
```diff
@@ -0,0 +1,100 @@
...
+  eleventyConfig.setBrowserSyncConfig({
+    notify: true,
+    files: './public/css/**/*.css',
+  });
+
+  eleventyConfig.setLibrary('md', mdIt/*markdownIt(options)*/);
+
+  eleventyConfig.addPassthroughCopy('./src/assets');
+
+  const fs = require('fs');
```
jpzwarte, line 61: I prefer these at the top. That way you don't have to search where it is defined.

docs/website/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "description": "SLDS Documentation",
+  "version": "1.0.0",
+  "main": "index.js",
+  "type": "module",
+  "scripts": {
+    "watch:sass": "npx sass --watch src/styles:public/css",
```
jpzwarte, line 8: Please change all use of `npm` to `yarn`?

docs/website/public/404.html
```diff
@@ -0,0 +1,182 @@
...
+    <nav class="ds-sidebar ds-flow--xs">
+  <h1 class="ds-title">
+    <a href="/">Sanoma Learning Design System</a>
+  </h1>
+
+  <div>
+  <input class="ds-search" type="search" id="search" name="search" value="" />
+</div>
+
+
```
jpzwarte, line 23: Is this a generated file? If so, so we check that into git?

packages/core/src/button-bar/button-bar.ts
```diff
@@ -19,6 +19,7 @@ export type ButtonBarAlign = 'start' | 'center' | 'end' | 'space-between';
  */
 export class ButtonBar extends LitElement {
   /** @private */
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
jpzwarte, line 22: Why do you need this? Does this error happen when there is no `.scss.js` file yet?

jpzwarte (changes requested):
My main question is whether we want all the generated file in Git.

docs/website/public/404.html
anna-lach: Yes, this one is generated (all in 'public' directory). I think we don't need to have all the generated files in git, I can change it. :)

docs/website/package.json
anna-lach: right

packages/core/src/button-bar/button-bar.ts
anna-lach: I had a problem with committing merge main branch after solving conflicts, it was weird eslint error with `any` here.

jpzwarte: Yeah, I tried it myself: if you haven't run a `yarn build` yet, then there is no `*.scss.ts` file and it won't know the type of `styles`. I couldn't fix this other than just running the build.

PR #64, opened 2022-10-28
Add basic grid component

jpzwarte:


packages/grid/src/grid.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  @property() items: T[] = [];
+
+  @property({ type: Boolean, reflect: true, attribute: 'no-border' }) noBorder?: boolean;
+
+  @property({ type: Boolean, reflect: true, attribute: 'no-row-border' }) noRowBorder?: boolean;
+
+  render(): TemplateResult {
+    return html`
+      <slot @slotchange=${this.#onSlotchange} style="display:none"></slot>
```
anna-lach, line 22: Maybe we should add a class with `display: none` to omit inline styles?

packages/grid/src/grid.ts
```diff
@@ -0,0 +1,50 @@
...
+      <table>
+        <thead>
+          <tr>
+            ${this.columns.map(col => col.renderHeaderCell())}
+          </tr>
+        </thead>
+        <tbody>
+          ${virtualize({ items: this.items, renderItem: item => this.renderItem(item) })}
+        </tbody>
+        <tfoot></tfoot>
```
anna-lach, line 32: Is `tfoot` necessary here since we don't add anything inside?

jpzwarte: It's a placeholder for now.

PR #75, opened 2022-11-04
Add radio-group components

jpzwarte:


packages/core/src/radio-group/radio-group.ts
```diff
@@ -0,0 +1,122 @@
...
+  attributeOldValue: true,
+  subtree: true
+};
+
+export class RadioGroup extends FormControlMixin(LitElement) {
+  /** @private */
+  static override styles: CSSResultGroup = styles;
+
+  #onClick = (): void => {
+    console.log('click', this.radios, this.radios?.[0]);
```
anna-lach, line 19: Do we still need this console.log? :)

packages/core/src/radio-group/radio-group.ts
```diff
@@ -0,0 +1,122 @@
...
+import { property, queryAssignedElements } from 'lit/decorators.js';
+import styles from './radio-group.scss.js';
+
+const OBSERVER_OPTIONS: MutationObserverInit = {
+  attributeFilter: ['checked'],
+  attributeOldValue: true,
+  subtree: true
+};
+
+export class RadioGroup extends FormControlMixin(LitElement) {
```
anna-lach, line 14: Maybe radiogroup should get role `radiogroup`?

https://www.w3.org/WAI/ARIA/apg/patterns/radiobutton/ 
"The radio buttons are contained in or owned by an element with role [radiogroup](https://w3c.github.io/aria/#radiogroup)."

packages/core/src/radio-group/radio.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    super.disconnectedCallback();
+  }
+
+  updated(changes: PropertyValues<this>): void {
+    super.updated(changes);
+
+    if (changes.has('checked')) {
+      this.internals.ariaChecked = this.checked ? 'true' : 'false';
+
```
anna-lach, line 57: Maybe checked radio should also have attribute `aria-checked = true`?

https://www.w3.org/WAI/ARIA/apg/patterns/radiobutton/
"If a radio button is checked, the radio element has [aria-checked](https://w3c.github.io/aria/#aria-checked) set to true. If it is not checked, it has [aria-checked](https://w3c.github.io/aria/#aria-checked) set to false."

packages/core/src/radio-group/radio-group.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    const activeElement = (this.getRootNode() as Document).activeElement as Radio;
+
+    let index = this.radios.indexOf(activeElement);
+    if (index === -1) {
+      return;
+    }
+
+    switch (event.key) {
+      case 'ArrowDown':
```
anna-lach, line 43: Maybe we should improve our keyboard navigation as described here https://www.w3.org/WAI/ARIA/apg/patterns/radiobutton/#keyboard-interaction-15 ?

packages/core/src/radio-group/radio.ts
jpzwarte: What do you mean? Using `ElementInternals`, it set's the ARIA attribute directly in the a11y tree. So there is no need to do `this.setAttribute('aria-checked', this.checked ? 'true' : 'false');`.

Issue #37, opened 2022-10-20
[POC] What are the possibilities in 11ty regarding search


DanielleRameau:
#36 

Find out what the possibilities are in 11ty regarding search functionality in navigation.

- [ ] What are the possibities
- [ ] Test

PR #77, opened 2022-11-07
docs website search POC

anna-lach:
Search based on elasticlunr

#37


docs/website/src/styles/categories/components.scss
```diff
@@ -0,0 +1,10 @@
+.wrapper {
+  column-gap: 32px;
```
jpzwarte, line 2: I believe we can now just use `gap: 0 32px;` (32px horizontal gap)

docs/website/src/styles/categories/components/button.scss
```diff
@@ -0,0 +1,11 @@
+h3 {
+  margin-bottom: 16px;
+}
+
+.ds-buttons-wrapper {
+  column-gap: 16px;
```
jpzwarte, line 6: Same here

docs/website/src/styles/categories/components/button.scss
```diff
@@ -0,0 +1,11 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  margin: 16px 0;
+  row-gap: 16px;
```
jpzwarte, line 10: So replace `column-gap` and `row-gap` with just `gap: 16px;`

packages/core/src/button-bar/button-bar.ts
```diff
@@ -19,6 +19,7 @@ export type ButtonBarAlign = 'start' | 'center' | 'end' | 'space-between';
  */
 export class ButtonBar extends LitElement {
   /** @private */
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
jpzwarte, line 22: You can remove this; there's a timing issue where if you haven't built `button-bar.scss.ts`, then it doesn't know the type of `styles`.

jpzwarte (approved):

packages/core/src/button-bar/button-bar.ts
anna-lach: right, I just forgot about it :)

PR #92, opened 2022-11-26
Add support for sticky columns to grid

jpzwarte:


packages/grid/src/grid.ts
```diff
@@ -37,14 +62,57 @@ export class Grid<T extends { [x: string]: unknown } = Record<string, unknown>>
...
+          cur.style.flexGrow = cur.style.width = '';
+
+          return Math.max(acc, width);
+        }, 0);
+      });
+
+    this.requestUpdate('columns');
+  }
+
   #onSlotchange(event: Event & { target: HTMLSlotElement }): void {
```
anna-lach, line 96: minor change :) maybe `#onSlotChange` instead of `#onSlotchange`

packages/grid/src/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+};
+
+export const getValueByPath = (object: unknown, path = ''): unknown => {
+  const parts = path?.split('.');
+
+  let result = object;
+  for (const part of parts) {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
+    result = (result as any)[part];
```
anna-lach, line 19: Can we get rid of this `any` here?

packages/grid/src/grid.ts
jpzwarte: Well the name of the DOM event is `slotchange`, so that's why `#onSlotchange`

packages/grid/src/utils.ts
jpzwarte: If you know how I could do the types here, sure. But I don't know how to fix the types so I don't need `any`.

PR #105, opened 2022-12-04
Add selection column to grid

jpzwarte:


packages/core/src/utils/decorators/event.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  emit(value: T, options?: EventOptions): boolean {
+    options = { bubbles: true, composed: true, ...this.options, ...options };
+
+    return this.target.dispatchEvent(new CustomEvent<T>(this.eventName, { detail: value, ...options }));
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function event(options?: EventOptions): any {
```
anna-lach, line 23: Maybe `unknown` here will be better?

anna-lach (approved):

packages/core/src/utils/decorators/event.ts
jpzwarte: `unknown` causes compiler errors when you use this decorator. Hopefully in 2023 when typescript supports the official ecmascript decorators, we can use the types from typescript itself.

PR #109, opened 2022-12-07
Add sortable columns to grid

jpzwarte:
- Add `@sanomalearning/utils/data-source` with `ArrayDataSource`
- Add `dataSource` property to `Grid` and use that for filtering & sorting
- Add `<sl-grid-sort-column>` for sortable columns
- Fix dialog backdrop blur in Safari


anna-lach:
Maybe we should also add `aria-sort`? https://www.w3.org/WAI/ARIA/apg/example-index/table/sortable-table

PR #110, opened 2022-12-08
Add button styling

jpzwarte:


anna-lach:
I'm thinking about disabled buttons, it looks like they're not visible by screen readers with 'disabled' attribute.

https://a11y-101.com/development/aria-disabled
https://www.smashingmagazine.com/2021/08/frustrating-design-patterns-disabled-buttons/#making-disabled-buttons-more-inclusive

anna-lach (approved):

PR #111, opened 2022-12-09
Add rich text editor component

jpzwarte:
- New `@sanomalearning/slds-editor` NPM package
- With `<sl-editor>` web component using prosemirror.net

### DIFF
--- a/packages/editor/src/editor.ts
+++ b/packages/editor/src/editor.ts
@@ -7,7 +7,7 @@ import { EditorView } from 'prosemirror-view';
 import { LitElement, html } from 'lit';
 import { property } from 'lit/decorators.js';
 import styles from './editor.scss.js';
-import { createContentNode } from './utils.js';
+import { createContentNode, getHTML } from './utils.js';
 import { marks, nodes } from './schema.js';
 import { setHTML } from './commands.js';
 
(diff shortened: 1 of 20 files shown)


packages/editor/src/editor.ts
```diff
@@ -0,0 +1,132 @@
...
+      state = this.createState();
+
+    const editor = new EditorView(
+      { mount },
+      {
+        state,
+        dispatchTransaction: function (tr) {
+          // `this` is bound to the view instance.
+          (this as unknown as EditorView).updateState(this.state.apply(tr));
+          // that.toolBarChanges.next(getToolBarState(this.state));
```
anna-lach, line 96: Do we still need this comment? :)

packages/editor/src/schema.ts
```diff
@@ -0,0 +1,371 @@
...
+export declare type EditorMarks =
+  | 'link'
+  | 'em'
+  | 'strong'
+  | 'code'
+  | 'underline'
+  | 'strikethrough'
+  | 'subscript'
+  | 'superscript';
+// | 'style';
```
anna-lach, line 12: Do we need this comment?

packages/editor/src/schema.ts
```diff
@@ -0,0 +1,371 @@
...
+  | 'image'
+  | 'hardBreak'
+  | 'listItem'
+  | 'orderedList'
+  | 'bulletList';
+
+const SLOT = 0; // https://prosemirror.net/docs/guide/#schema.serialization_and_parsing
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export const isEmpty = (obj: Object): boolean => Object.keys(obj).length === 0;
```
anna-lach, line 31: Maybe we can use `(obj: Record<string, unknown>)` instead of `(obj: Object)` here?

packages/editor/src/schema.ts
```diff
@@ -0,0 +1,371 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const removeEntries = (obj: any, predicate: (key: string) => boolean): Record<string, string> => {
```
anna-lach, line 34: Maybe here instead of `obj: any` we can use `obj: Record<string, unknown>` as well?

packages/editor/src/schema.ts
```diff
@@ -0,0 +1,371 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+    Object.keys(obj)
+      .filter(key => predicate(key))
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
+      .reduce((acc, curr: any) => Object.assign(acc, { [curr]: obj[curr] }), {})
+  );
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const removeEmptyEntries = (obj: any): Record<string, string> => {
```
anna-lach, line 45: Maybe here instead of `obj: any` we can use `obj: Record<string, unknown>` as well?

PR #134, opened 2023-01-05
Feature/drawer

Diaan:


packages/core/src/drawer/drawer.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+      dialog?.dispatchEvent(new Event('close'));
+      expect(document.documentElement.style.overflow).to.equal('');
+
+    });
+
+    it('should not close the drawer when the cancel event is fired but close is disabled', async () => {
+      const dialog = el.renderRoot.querySelector('dialog');
+      const cancelEvent = new Event('cancel');
+      const cancelEventSpy = spy(cancelEvent,'preventDefault')
+
```
jpzwarte, line 65: please remove superfluous newline

packages/core/src/drawer/drawer.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+      
+      expect(cancelEventSpy).to.have.been.called;
+    });
+
+    it('should close the drawer when the cancel event is fired and close isn`t disabled', async () => {
+      const dialog = el.renderRoot.querySelector('dialog');
+      const cancelEvent = new Event('cancel');
+      const cancelEventSpy = spy(cancelEvent,'preventDefault')
+
+
```
jpzwarte, line 83: please remove superfluous newline

packages/core/src/drawer/drawer.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+      
+      dialog?.dispatchEvent(cancelEvent);
+      
+      expect(cancelEventSpy).to.have.been.called;
+    });
+
+    it('should close the drawer when the cancel event is fired and close isn`t disabled', async () => {
+      const dialog = el.renderRoot.querySelector('dialog');
+      const cancelEvent = new Event('cancel');
+      const cancelEventSpy = spy(cancelEvent,'preventDefault')
```
jpzwarte, line 81: Sigh, I need to figure out how to enable eslint for `*.spec.ts` files; you're missing a semicolon here :)

packages/core/src/drawer/drawer.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+};
+
+const onClick = (event: Event & { target: HTMLElement }): void => {
+  (event.target.nextElementSibling as Drawer).showModal();
+};
+
+export const API: StoryObj = {
+  render: ({ attachment, buttonSize }) => html`
+    <sl-button @click=${onClick}>Show Drawer</sl-button>
+    <sl-drawer .attachment=${attachment} .closeButtonSize=${buttonSize}>
```
jpzwarte, line 32: I'm not sure if `closeButtonSize` is a valid public property. Perhaps it should be something internal? So you could have a design token with the size of the close button?

packages/core/src/drawer/drawer.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+      <h2 slot="title">In this sidepanel you can find a lot of info</h2>
```
jpzwarte, line 33: Can this also use a `<div slot="title">`? Often using an `<h2>` automatically applies styling? And that may conflict with the drawer styling?

packages/core/src/drawer/drawer.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+      <p>
+        Apple pie chocolate jelly-o carrot cake gummi bears halvah cake cheesecake. Sesame snaps macaroon shortbread
```
jpzwarte, line 35: Apple pie..... 🤤

packages/core/src/drawer/drawer.stories.ts
```diff
@@ -0,0 +1,89 @@
...
+      </p>
+    </sl-drawer>
+  `
+};
+
+export const DisableClose: StoryObj = {
+  render: () => html`
+    <sl-button @click=${onClick}>Show Drawer</sl-button>
+    <sl-drawer disable-close>
+      <span slot="title" id="title">Drawer title</span>
```
jpzwarte, line 48: Can we add the `id="title"` automatically perhaps? Otherwise people may forget to set it?

packages/core/src/drawer/drawer.ts
```diff
@@ -0,0 +1,110 @@
+import type { TemplateResult } from 'lit-html';
+import type { CSSResultGroup } from 'lit';
+import type { ButtonSize } from '../button/button.js';
+import { LitElement, html } from 'lit';
+import { property, query } from 'lit/decorators.js';
+import styles from './drawer.scss.js';
+
+export type DrawerAttachement = 'right' | 'left' | 'top' | 'bottom';
```
jpzwarte, line 8: `DrawerAttachment` (1 "e" teveel :) )

packages/core/src/drawer/drawer.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * A dialog component for displaying modal UI.
```
jpzwarte, line 11: "dialog" :)

packages/core/src/drawer/drawer.ts
```diff
@@ -0,0 +1,110 @@
...
+  /** @private */
+  static override styles: CSSResultGroup = styles;
+
+  @query('dialog') dialog?: HTMLDialogElement;
+
+  /** Disables the ability to close the dialog using the Escape key. */
+  @property({ type: Boolean, attribute: 'disable-close' }) disableClose = false;
+
+  /** The ARIA role of the dialog. */
+  @property() override role: 'dialog' | 'section' = 'dialog';
```
jpzwarte, line 27: Is `section` a valid role here? I thought only `dialog` and `alertdialog` were?

packages/core/src/drawer/drawer.ts
```diff
@@ -0,0 +1,110 @@
...
+        .role=${this.role}
+        aria-labelledby="title"
+        part="dialog"
+      >
+        <div>
+          <sl-button sl-dialog-close .size="${this.closeButtonSize}" aria-label="back to page" title="close"
+            >x</sl-button
+          >
+          <slot name="title" id="title"></slot>
+          <slot name="actions"></slot>
```
jpzwarte, line 56: So in DNA, I usually wrap the `<slot name="actions">` in a button-bar (`<sl-button-bar>`); that way the button-bar automatically spaces the buttons if they are next to each other.

packages/core/src/drawer/drawer.ts
```diff
@@ -0,0 +1,110 @@
...
+export type DrawerAttachement = 'right' | 'left' | 'top' | 'bottom';
+
+/**
+ * A dialog component for displaying modal UI.
+ *
+ * @slot default - Body content for the dialog
+ * @slot header - Header content for the dialog
+ * @slot title - The title of the dialog
+ */
+export class Drawer extends LitElement {
```
jpzwarte, line 17: So since you're using other SLDS web components inside the drawer, you should use `ScopedElementsMixin`; otherwise you may run into conflicts when 2 components try to define the same child web component.

packages/core/src/drawer/drawer.ts
```diff
@@ -0,0 +1,110 @@
...
+        part="dialog"
+      >
+        <div>
+          <sl-button sl-dialog-close .size="${this.closeButtonSize}" aria-label="back to page" title="close"
+            >x</sl-button
+          >
+          <slot name="title" id="title"></slot>
+          <slot name="actions"></slot>
+        </div>
+        <slot name="body">
```
jpzwarte, line 58: You could remove `name="body"` here and just have it be the default slot?

jpzwarte (approved):

packages/core/src/drawer/drawer.stories.ts
Diaan: But isn't this the freedom we want to give users so they can put in the drawer what they like? We just provide the slot for a title. Whether that needs a heading it is up to the context.

jpzwarte: Yeah, you're right :)

packages/core/src/drawer/drawer.scss
```diff
@@ -0,0 +1,95 @@
...
+
+dialog[open] {
+  transform: translate(0, 0) scale(1, 1);
+}
+
+div {
+  align-items: center;
+  display: grid;
+  gap: var(--_gap);
+  grid-auto-flow: column;
```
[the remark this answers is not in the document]
Diaan: Yes, because of the grid-column-start for the close button, otherwise it would make 2 rows and the title and other action buttons would be on the second row

packages/core/src/drawer/drawer.spec.ts
Diaan: being too reliant on linting makes you a lazy developer 🫣

packages/core/src/drawer/drawer.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  // it('should have a dialog role', async () => {
+  //   const { role } = await a11ySnapshot({ selector: 'dialog' }) as any;
+
+  //   expect(role).to.equal('dialog');
+  // });
+
+  describe('positioning', () => {
+    it('should attach the drawer to the right by default', () => {
+      expect(el).to.have.attribute('attachment', 'right');
```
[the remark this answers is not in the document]
Diaan: I don't know how much of the position of a drawer is based on the writing direction and orientation of the text and how much it's based on taste/randomness.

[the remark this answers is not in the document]
jpzwarte: Let's stick with `left`, `right` etc. That's more clear than the alternative.

packages/core/src/drawer/drawer.stories.ts
jpzwarte: Perhaps have a `no-close-button` attribute that causes the component to not render a close button? For example when you have a form inside the drawer that has a "Cancel" button?

jpzwarte: But perhaps this is beyond the scope of the PoC :)

PR #135, opened 2023-01-06
Create custom `FormControlMixin`

jpzwarte:


packages/core/src/checkbox/checkbox-group.ts
```diff
@@ -1,42 +1,73 @@
...
+        .join(', ');
+
+      this.#validation.validate(value);
+    },
+    config: { attributeFilter: ['checked'], attributeOldValue: true, subtree: true }
+  });
+
+  /** Manage the keyboard navigation. */
   #rovingTabindexController = new RovingTabindexController<Checkbox>(this, {
     focusInIndex: (elements: Checkbox[]) => elements.findIndex(el => !el.disabled),
```
anna-lach, line 42: What do you think about focusing all the checkbox element with text just like in this example (not only checkbox element) https://www.w3.org/WAI/ARIA/apg/example-index/checkbox/checkbox?

packages/core/src/radio-group/radio-group.ts
```diff
@@ -86,27 +106,22 @@ export class RadioGroup extends FormControlMixin(HintMixin(LitElement)) {
...
-      button.checked = button.value === value;
-    });
-
-    this.#observer?.observe(this, OBSERVER_OPTIONS);
+  #onFocusout(event: FocusEvent): void {
+    if (!event.relatedTarget || !this.buttons.includes(event.relatedTarget as Radio)) {
+      // This component is weird in that it doesn't actually contain the form controls,
+      // those are the `<sl-radio>` custom elements in the light DOM. So run the validation
+      // manually from here.
+      // this.#validation.validate();
```
anna-lach, line 124: Do we want to keep it for now?

packages/core/src/utils/controllers/validation.ts
```diff
@@ -0,0 +1,363 @@
...
+        return this.validationMessage;
+      case 'patternMismatch':
+        return msg('Please match the specified pattern.');
+      case 'rangeOverflow':
+        return msg('The value is outside the minimum or maximum allowed range.');
+      case 'stepMismatch':
+        return msg('Step mismatch');
+      case 'tooLong':
+        if ('maxLength' in this) {
+          const length = /*this.value?.toString().length ??*/ 0,
```
anna-lach, line 324: Do we still need this comment?

packages/core/src/utils/mixins/form-control.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    setValidity(flags?: ValidityStateFlags, message?: string, anchor?: HTMLElement): void {
+      console.log('setValidity', { flags, message, anchor });
+    }
+
+    setFormValue(value?: FormControlValue): void {
+      this.#cachedValue = value;
+
+      const valueShouldUpdate = true, //this.shouldFormValueUpdate(),
```
anna-lach, line 142: Do we need this comment?

packages/core/src/utils/validators.ts
```diff
@@ -0,0 +1,153 @@
+export type ValidationValue = File | FormData | string | undefined;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ValidationMessageCallback = (instance: any, value: ValidationValue) => string;
```
anna-lach, line 4: Can we use `unknown` here instead of `any`?

packages/core/src/utils/validators.ts
```diff
@@ -0,0 +1,153 @@
...
+    const _value = (value as string) || '';
+    return `Please use at least ${instance.minLength} characters (you are currently using ${_value.length} characters).`;
+  },
+  isValid(instance: HTMLElement & { minLength: number }, value: string): boolean {
+    /** If no value is provided, this validator should return true */
+    if (!value) {
+      return true;
+    }
+
+    if (!!value && instance.minLength > value.length) {
```
anna-lach, line 111: Maybe this part, lines from 111 to 115 can be simplified to sth like
`return !(!!value && instance.minLength > value.length);` ?

packages/core/src/utils/validators.ts
```diff
@@ -0,0 +1,153 @@
...
+    const _value = (value as string) || '';
+    return `Please use no more than ${instance.maxLength} characters (you are currently using ${_value.length} characters).`;
+  },
+  isValid(instance: HTMLElement & { maxLength: number }, value: string): boolean {
+    /** If maxLength isn't set, this is valid */
+    if (!instance.maxLength) {
+      return true;
+    }
+
+    if (!!value && instance.maxLength < value.length) {
```
anna-lach, line 132: Here we can also simplify it to `return !(!!value && instance.maxLength < value.length);`?

anna-lach (approved):

packages/core/src/utils/validators.ts
```diff
@@ -0,0 +1,153 @@
...
+    return true;
+  }
+};
+
+export const maxLengthValidator: Validator = {
+  attribute: 'maxlength',
+  key: 'rangeOverflow',
+  message(instance: { maxLength: number }, value: ValidationValue): string {
+    const _value = (value as string) || '';
+    return `Please use no more than ${instance.maxLength} characters (you are currently using ${_value.length} characters).`;
```
Diaan, line 124: How is this different than what happens in controllers/validation? Feels like a duplication.

packages/editor/src/editor.ts
```diff
@@ -43,7 +54,7 @@ export class Editor extends FormControlMixin(LitElement) {
     this.#value = value;
 
     if (this.#view) {
-      setHTML(value || '')(this.#view.state, this.#view.dispatch, this.#view);
+      setHTML(value?.toString() || '')(this.#view.state, this.#view.dispatch, this.#view);
```
Diaan, line 57: Why do you need `toString` here? It either is a string or it it undefined, according to the typing, and when it's undefined it becomes an empty string.

packages/core/src/utils/controllers/validation.ts
jpzwarte: I want to leave this here for now.

packages/core/src/utils/validators.ts
jpzwarte: That will break all the locations where `instance` is passed.

packages/core/src/utils/validators.ts
jpzwarte: I would like to change this only after there are unit tests for the validators.

packages/core/src/utils/validators.ts
jpzwarte: Agreed, but there are different use cases here:
1. For custom validators in combination with form associated custom elements
2. For built-in validators in combination with `<input>` or `<textarea>`

But I agree that we should definitely look at refactoring the messages (in the future).

packages/core/src/utils/validators.ts
jpzwarte: I would like to change this only after there are unit tests for the validators.

packages/core/src/checkbox/checkbox-group.ts
jpzwarte: I think that's more of a UI design question. Since checkbox is still being designed, let's wait for that :)

PR #216, opened 2023-01-24
Feature/tabs

Diaan:


.vscode/settings.json
```diff
@@ -7,5 +7,8 @@
   "eslint.format.enable": true,
   "[typescriptreact]": {
     "editor.defaultFormatter": "dbaeumer.vscode-eslint"
+  },
+  "files.exclude": {
+    "**/*.scss.ts": true
```
Diaan, line 12: I've added this in combination with the [_Make Hidden_ plugin](https://marketplace.visualstudio.com/items?itemName=devine-davies.make-hidden) so you don't see them in the in Explorer tab in VS Code, since you don't need to edit those files any way. I guess if you don't have the plugin this doesn't matter, but I found it really nice

packages/core/src/tabs/tab-group.ts
```diff
@@ -0,0 +1,248 @@
+import type { CSSResultGroup, TemplateResult } from 'lit';
+import type { EventEmitter } from '../utils/decorators/event.js';
+import type { ScopedElementsMap } from '@open-wc/scoped-elements';
+import { ScopedElementsMixin } from '@open-wc/scoped-elements';
+import { LitElement, html } from 'lit';
+import { property, queryAssignedElements, state } from 'lit/decorators.js';
+import { event } from '../utils/decorators/event.js';
+import { RovingTabindexController } from '../utils/controllers/roving-tabindex.js';
```
jpzwarte, line 8: This should be from `controllers/index.js`

packages/core/src/tabs/tab-group.ts
```diff
@@ -0,0 +1,248 @@
+import type { CSSResultGroup, TemplateResult } from 'lit';
+import type { EventEmitter } from '../utils/decorators/event.js';
+import type { ScopedElementsMap } from '@open-wc/scoped-elements';
+import { ScopedElementsMixin } from '@open-wc/scoped-elements';
+import { LitElement, html } from 'lit';
+import { property, queryAssignedElements, state } from 'lit/decorators.js';
+import { event } from '../utils/decorators/event.js';
```
jpzwarte, line 7: Also from `decorators/index.js`

packages/core/src/tabs/tab-group.ts
```diff
@@ -0,0 +1,248 @@
...
+
+  /** @private */
+  static override styles: CSSResultGroup = styles;
+
+  /**
+   * Unique ID for each tab group component present.
+   */
+  private tabGroupId = `sl-tab-group-${tabGroupCount++}`;
+
+  private observer?: MutationObserver;
```
jpzwarte, line 32: Please use `#observer` instead of the typescript `private` keyword. The first is pure JS, the latter is TS-only :)

packages/core/src/tabs/tab-group.ts
```diff
@@ -0,0 +1,248 @@
...
+  });
+
+  /**
+   * The current tab node selected in the tab group.
+   */
+  @state() private selectedTab: Tab | null = this.initialSelectedTab;
+
+  @event() tabChange!: EventEmitter<number>;
+
+  @property({ reflect: true }) vertical = false;
```
jpzwarte, line 57: Please add a comment for each `@property`; that is then used for documentation later on.

packages/core/src/tabs/tab-group.ts
```diff
@@ -0,0 +1,248 @@
...
+    this.updateSlots();
+  }
+
+  override firstUpdated(): void {
+    this.observer = new MutationObserver(this.handleMutation);
+    this.observer?.observe(this, TabGroup.observerOptions);
+    setTimeout(() => this.#updateSelectionIndicator(), 100);
+  }
+
+  private updateSlots(): void {
```
jpzwarte, line 85: Please use `#updateSlots()` instead of `private` TS keyword here as well (and below).

packages/core/src/tabs/tab-group.ts
```diff
@@ -0,0 +1,248 @@
...
+    this.tabChange.emit(tabIndex);
+
+    this.#updateSelectionIndicator();
+  }
+
+  /**
+   * Handle keyboard accessible controls.
+   */
+  private handleKeydown(event: KeyboardEvent): void {
+    switch (event.key) {
```
jpzwarte, line 198: Isn't an `if (['Enter', ' '].includes(event.key) { ...` shorter? Switch seems slightly overkill :)

jpzwarte (approved):

.vscode/settings.json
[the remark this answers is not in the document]
Diaan: 😍

packages/core/src/tabs/tab-group.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  /**
+   * Get the selected tab button, or the first tab button.
+   */
+  private get initialSelectedTab(): Tab | null {
+    return this.querySelector('sl-tab[selected]') || this.querySelector('sl-tab:not([disabled])');
+  }
+
+  override render(): TemplateResult {
+    return html`<div @click=${this.handleTabChange} role="tablist" @keydown=${this.handleKeydown} part="tab-list">
```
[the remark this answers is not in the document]
Diaan: auto formatting doesn't allow that 😉

packages/core/src/tabs/tab-group.ts
Diaan: I guess this goes for all private functions and variables?

PR #218, opened 2023-01-25
Docs/component documentation instructions

anna-lach:
Here you can find the information about adding component documentation:

https://github.com/sanomalearning/design-system/blob/docs/component-documentation-instructions/docs/website/README.md


docs/website/README.md
```diff
@@ -0,0 +1,428 @@
...

+
+## Component's files structure
+
+Every component directory consists of **7 files**, like for a **button** directory there will be:
+
+![image](https://user-images.githubusercontent.com/111562742/214264654-af3891e9-8814-45e0-b550-fc57d2f5af16.png)
+
+
+- `accessibility.md` - file keeps content for the **“Accessibility”** tab for a button component
```
Diaan, line 36: i think "contains" is a better word than "keeps" in this sentence (and all the following ;) )

docs/website/README.md
```diff
@@ -0,0 +1,428 @@
...
+
+More information about component's files content you can find below.
+
+## Component json file
+
+### `button.json` file
+
+![image](https://user-images.githubusercontent.com/111562742/214271396-14edcc44-d487-4fb4-822b-cef63e7da0d5.png)
+
+`button.json` file consists of **layout** and **tags** elements. Layout should be the same for every component `"components/components.njk"` and tags should contain component name like `"button"` in this example.
```
Diaan, line 52: when making the documentation for the _button bar_ for example, does this need to be camel case or kebab case?
And is this always the same as the filename of the actual component?

docs/website/README.md
```diff
@@ -0,0 +1,428 @@
...
+
+## Component markdown files
+
+> Please, keep in mind that identation in markdown files is an important thing.
```
Diaan, line 56: identation => indentation

docs/website/README.md
```diff
@@ -0,0 +1,428 @@
...
+eleventyNavigation:
+  parent: Components
+  key: Button
+  order: 2
+---
+```
+  - `title` contains a component name, which we can find in the output file:
+    <img width="482" alt="image" src="https://user-images.githubusercontent.com/111562742/214508369-2bd45963-1013-4214-ba01-3c4e6a58a461.png">
+    
+    - `description` contains short description of the component, which we can find in the output file:
```
Diaan, line 78: the indentation of `description` and all the following properties is incorrect, it should have the same weight as `title`

docs/website/README.md
```diff
@@ -0,0 +1,428 @@
...
+    <img width="482" alt="image" src="https://user-images.githubusercontent.com/111562742/214509804-c19ff048-cd43-4cf7-8044-3451b84ad6d9.png">
+
+    - `layout` for components should always contain `"components/components.njk"`
+    - `tags` should always contain `component` for `component.md` files
+    - `eleventyNavigation` contains options used for navigation on our website, `parent` should be `Components` and `order` represents order of showing components in our navigation (Button order is 2, so we can find it on the second place under `Components` navigation item):
```
Diaan, line 83: so if we have 20 components and we want to add one starting with and `A` we need to update 20 files and increase the order with 1?

jpzwarte (approved):

Diaan (changes requested):

docs/website/README.md
anna-lach: right 😀

docs/website/README.md
anna-lach: Good point :) kebab case as far as I know :) I'll add this information.

docs/website/README.md
Diaan: Is there really no easier way to do this?

anna-lach: Maybe we can add sth like custom sorting in the future.

arecuenco-dsgn (approved):
For me, it's a perfect start. Maybe, when I create the first Docs for the buttons, I can suggest some additions.  👍

DanielleRameau (approved):

Diaan (approved):

PR #242, opened 2023-01-30
Feature/137 select component

Diaan:


packages/core/src/avatar/avatar.ts
```diff
@@ -0,0 +1,59 @@
...
+    super.connectedCallback();
+
+    this.user = await this._getUserDetails(this.uniqueProfileId);
+  }
+
+  async _getUserDetails(id: string): Promise<UserProfile | undefined> {
+    try {
+      const response = await fetch(`https://randomuser.me/api/?inc=picture,name&seed=slds-${id}`);
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
+      const json: { results: UserProfile[]; info: any } = await response?.json();
```
anna-lach, line 52: Maybe we can use `unknown` here instead of `any`?

packages/core/src/select/select-option.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  override connectedCallback(): void {
+    super.connectedCallback();
+    this.setAttribute('role', 'option');
+  }
+
+  override firstUpdated(): void {
+    this.#observer = new ResizeObserver(m => this.#handleResize(m));
+    this.#observer?.observe(this);
```
anna-lach, line 50: Maybe we should add `this.#observer?.disconnect()` on `disconnectedCallback()`?

packages/core/src/select/select-overlay.scss
```diff
@@ -0,0 +1,21 @@
...
+  --_max-overlay-height: var(--max-overlay-height, 80vh);
+}
+
+:host {
+  background-color: #fff;
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
+  max-height: var(--_max-overlay-height);
+  overflow: auto;
```
anna-lach, line 11: Maybe we should remove it because we have `overflow-x` and `overflow-y` below?

anna-lach:
It looks like exports for select and avatar components is missing in `package.json`?

packages/core/src/avatar/avatar.ts
```diff
@@ -0,0 +1,59 @@
+import type { CSSResultGroup, TemplateResult } from 'lit';
+import { LitElement, html } from 'lit';
+import { property, state } from 'lit/decorators.js';
+import styles from './avatar.scss.js';
+
+export interface UserProfile {
+  name: {
+    first: string;
```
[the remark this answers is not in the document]
Diaan: This is actually a (very) temporary component, just to have something to render in the select, so it's not really very important, but I've changed it anyway

anna-lach:
> It looks like exports for select and avatar components are missing in `package.json`?

Perhaps we also need to add exports for select option and select option group?

Issue #143, opened 2023-01-09
[PoC] Add Angular form control integration


jpzwarte:
Add initial Angular form control integration.

Requirements:
- Add a `CustomValueAccessor` for `<sl-input>`, `<sl-textarea>` etc.
- Support template forms as well as .. other one
- Add some examples in Storybook
- This story specifically does *not* include validation

PR #243, opened 2023-01-30
[POC] Feature/Angular form control integration

anna-lach:
#143 

- Angular form control integration (POC),
- Unnecessary Angular app removed


packages/angular/projects/angular-core/src/forms/core-forms.module.ts
```diff
@@ -0,0 +1,29 @@
...
+  exports: [
+    CheckboxDirective,
+    InputDirective,
+    TextareaDirective,
+    RadioGroupDirective,
+    RadioDirective
+  ],
+  schemas: [CUSTOM_ELEMENTS_SCHEMA]
+})
+export class CoreFormsModule { }
```
jpzwarte, line 29: This should probably be renamed to `FormsModule`

jpzwarte (approved):
There's probably some overlap in directives here. With some refactor, we should be able to serve multiple web components with a single directive (similar to DNA). But for now this is fine.

Issue #182, opened 2023-01-18
Development of explore section


DanielleRameau:
## Development of hero block
**This task is part of https://github.com/sanomalearning/design-system/issues/180**

## Description
Based on the outcome of the design ideation of the homepage of sanomalearning.com we created a UI design of the explore section for all breakpoints and is available in light and dark mode. This explore section has to be developed.

## Design 
See issue https://github.com/sanomalearning/design-system/issues/181

https://github.com/sanomalearning/design-system/issues/182#tasklist-block-86650002-ae4b-444a-94ef-8f67fd5592a8

PR #250, opened 2023-02-07
Feature/website homepage explore section

anna-lach:
#182


docs/website/src/site/index.md
```diff
@@ -51,4 +51,120 @@ eleventyNavigation:
...
+<path class="ds-explore__components-part" d="M118.209 105.558L121.568 111.638" stroke="#5985FF" stroke-width="5.25962" stroke-miterlimit="10" stroke-linecap="round"/>
+<path class="ds-explore__components-part" d="M125.346 84.7039L131.64 81.7661" stroke="#5985FF" stroke-width="5.25962" stroke-miterlimit="10" stroke-linecap="round"/>
+<path class="ds-explore__components-part" d="M124.954 95.9535L133.071 98.6218" stroke="#5985FF" stroke-width="5.25962" stroke-miterlimit="10" stroke-linecap="round"/>
+</svg>
+
+
+</div>
+
+<div class="ds-heading-2">
+<a href="/categories/components/overview/" aria-label="Components. Sanoma Learning’s UI components are built to best practices to ensure usability, performance and accessibility.">
```
Diaan, line 107: should the label be the same as the link text? doesn't that read the same text aloud twice?

docs/website/src/site/index.md
```diff
@@ -51,4 +51,120 @@ eleventyNavigation:
...
+Components
+</a>
+</div>
+
+Sanoma Learning’s UI components are built to best practices to ensure usability, performance and accessibility.
```
Diaan, line 112: I'm not sure this sentence is correct... i think you mean 
> Sanoma Learning’s UI components are built *according* to best practices to ensure usability, performance and accessibility.

But then still i'm not sure what message you are trying to get across...

docs/website/src/site/index.md
```diff
@@ -51,4 +51,120 @@ eleventyNavigation:
...
+
+</div>
+
+<div class="ds-heading-2">
+<a href="/categories/design-tokens/" aria-label="Design tokens. Tokens are platform-agnostic variables that represent the look and feel of our brands.">
+Design tokens
+</a>
+</div>
+
+Tokens are platform-agnostic variables that represent the look and feel of our brands.
```
Diaan, line 135: maybe "define" or "make up" instead of represent?
That way it's clearer that the tokens are the input for the DS, so it looks the way you want.

Also maybe, use "your brand" instead of "our brands" that way you get the reader more involved

docs/website/src/site/index.md
```diff
@@ -51,4 +51,120 @@ eleventyNavigation:
...
+
+</div>
+
+<div class="ds-heading-2">
+<a href="/categories/guidelines/" aria-label="Guidelines. Practical guides to get started with designing and developing experiences with Sanoma Learning’s Design System.">
+Guidelines
+</a>
+</div>
+
+Practical guides to get started with designing and developing experiences with Sanoma Learning’s Design System.
```
Diaan, line 164: "your experience" instead of "experience", again, to involve the reader

docs/website/src/site/styles/homepage/explore.scss
```diff
@@ -0,0 +1,205 @@
...
+  @media screen and (min-width: $md-breakpoint) {
+    margin-top: 4rem;
+  }
+
+  @media screen and (min-width: $xl-breakpoint) {
+    margin-top: 4.8rem;
+  }
+
+  h2 {
+    font-size: 2.4rem;
```
Diaan, line 17: why not use variables for this? so you can make sure the headings are the same all over the pages?

Diaan (changes requested):

docs/website/src/site/styles/homepage/explore.scss
anna-lach: in other places h1 has 2.4rem (in general), but we already have h1 on the homepage, maybe I should use in this example `ds-heading-2` class instead. :)

docs/website/src/site/index.md
DanielleRameau: The sentence is correct, 'are built to' should be ok. I want to say that Sanoma Learning's components are made with good accessibility and UX in mind. I believe that the message should be clear with this sentence.

docs/website/src/site/index.md
DanielleRameau: @Diaan thanks for your input
@anna-lach can we change it to: 'Tokens are platform-agnostic variables that define the look and feel of your brand.

docs/website/src/site/index.md
DanielleRameau: @Diaan, Thanks!
@anna-lach, Can we change it to:'Practical guides to get started with designing and developing your experience with Sanoma Learning's Design System'

Diaan (approved):

docs/website/src/site/index.md
Diaan: when i use the arrowkeys to navigate over all the elements (and don't click the link) i get the same text twice; once from the aria-label and once from the paragraph below it

Issue #189, opened 2023-01-19
Development of call to action block


DanielleRameau:
## Description
Based on the outcome of the design ideation of the homepage we created a UI design of the call to action block for all breakpoints and is available in light and dark mode. This explore section has to be developed.

## Design 
See issue https://github.com/sanomalearning/design-system/issues/188

Create hyperlink components in different states
-idle
-hover
-active
-focus

## Content

The content in the CTA block is created manually

https://github.com/sanomalearning/design-system/issues/189#tasklist-block-08abca7c-c82f-42a6-a437-38699a68eacd

PR #270, opened 2023-02-13
Feature/website homepage call to action block

anna-lach:
#189


docs/website/src/site/styles/homepage/call-to-action.scss
```diff
@@ -0,0 +1,311 @@
...
+
+  @media screen and (min-width: $md-lg-breakpoint) {
+    direction: rtl;
+    grid-template-columns: repeat(2, 1fr);
+    grid-template-rows: none;
+    padding: 4rem;
+  }
+
+  @media screen and (min-width: $lg-breakpoint) {
+    direction: ltr;
```
Diaan, line 26: why? this is the writing direction, that shouldn't change, right?

docs/website/src/site/styles/homepage/call-to-action.scss
```diff
@@ -0,0 +1,311 @@
...
+
+.ds-call-to-action__smile {
+  animation: smileAnimation 3s infinite;
+  animation-delay: $transition-timing;
+  backface-visibility: hidden;
+  transform: translate3d(0, 0, 0);
+  transform-box: fill-box;
+  transform-origin: center;
+
+  @media screen and (prefers-reduced-motion: reduce) {
```
Diaan, line 98: you can also group all the animations together in `@media (prefers-reduced-motion: no-preference)` instead of overwriting it.

docs/website/src/site/styles/homepage/call-to-action.scss
```diff
@@ -0,0 +1,311 @@
...
+  &:active {
+    color: var(--cta-link-color--active);
+
+    &::after {
+      background-image: var(--chevron-icon--active);
+    }
+  }
+}
+
+@keyframes blink {
```
Diaan, line 205: 😉

docs/website/src/site/styles/variables.scss
```diff
@@ -17,6 +18,7 @@
   --ds-color-blue-rgb: 51, 102, 255;
   --ds-color-baby-blue: #9ebaff;
   --ds-color-blueberry: #5985ff;
+  --ds-color0blueberry-light: #5f87ff;
```
Diaan, line 21: typo?

docs/website/src/site/styles/homepage/call-to-action.scss
anna-lach: It was a kind of workaround to change order in css grid, but yeah I'll use `order` instead :)

docs/website/src/site/index.md
```diff
@@ -266,7 +266,7 @@ Guidelines
 
 <section class="ds-call-to-action">
 
-<svg width="296" height="308" viewBox="0 0 296 308" fill="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="ctaTitle ctaDesc" role="img">
+<svg width="296" height="308" viewBox="0 0 296 308" fill="none" preserveAspectRatio="none" xmlns="http://www.w3.org/2000/svg" aria-labelledby="ctaTitle ctaDesc" role="img">
```
Diaan, line 269: why did you add this? if the width and height are set explicitly it shouldn't be necessary, and also i don't think `none` would be the best value, since that would cause the image to distort.

docs/website/src/site/styles/homepage/call-to-action.scss
```diff
@@ -296,6 +323,34 @@ $transition-timing: 200ms;
...
+  }
+}
+
+@keyframes smileBottomAnimationSafari {
+  30% {
+    transform: translateY(0);
+  }
+
+  60% {
+    transform: translateY(-6px);
```
Diaan, line 346: can't you make the number of pixels a css variable and set that according to whether it's safari? It's a shame to have so many different animations when the only difference is 1px ;)

Diaan (approved):

docs/website/src/site/index.md
anna-lach: yes, it's not necessary for now, I was just trying different solutions for Safari and I left it by chance. :)

docs/website/src/site/styles/homepage/call-to-action.scss
```diff
@@ -69,65 +69,73 @@ $transition-timing: 200ms;
...
+
+    .ds-call-to-action__smile,
+    .ds-call-to-action__smile-left {
+      --element-transform: -5px;
+      animation: translateAnimation 3s infinite;
+    }
+
+    .ds-call-to-action__smile-bottom {
+      --element-transform: -7px;
+      animation: translateAnimation 3s infinite;
```
Diaan, line 132: this is the same as in the non-safari animation, right?

anna-lach: yes, changed ;)

PR #292, opened 2023-02-14
Feature/237 checkbox component

Diaan:


packages/core/src/checkbox/checkbox-group.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+      await sendKeys({ press: 'Enter' });
+
+      expect(el.boxes?.[0].checked).to.equal(true);
+      expect(el.boxes?.[1].checked).to.equal(true);
+    });
+
+    it('should be valid when no option is chosen but the group is not required', async () =>{
+      el.boxes?.forEach(b=>b.blur());
+      // this doesn't work, can this even be tested here?
+      // expect(el.boxes?.[0].tabIndex).to.equal(-1);
```
anna-lach, line 61: Do we still want to keep these comments here?

packages/core/src/checkbox/checkbox.scss
```diff
@@ -52,25 +47,100 @@
...
+ */
+
+@each $size in $sizes {
+  :host([size='#{$size}']) {
+    --_line-height: var(--sl-text-typeset-line-height-#{$size});
+    --_padding: var(--sl-space-input-block-#{$size})
+      var(--sl-space-input-inline-#{$size})
+      var(--sl-space-input-block-#{$size})
+      0;
+    // --_path-length: 22.910259;
```
anna-lach, line 96: Do we need to keep this comment? :)

packages/core/src/checkbox/checkbox.scss
```diff
@@ -1,16 +1,18 @@
+$variants: default, valid, invalid;
+$variantselectors: ':not([valid],[invalid])', ':where([valid])', ':where([invalid])';
+$states: unchecked, checked;
+$stateselectors: '', ':where(:--checked, :--indeterminate, [state--checked], [state--indeterminate])';
```
anna-lach, line 4: Maybe kebab case for variable naming here as well?

packages/core/src/checkbox/checkbox.scss
```diff
@@ -1,16 +1,18 @@
+$variants: default, valid, invalid;
+$variantselectors: ':not([valid],[invalid])', ':where([valid])', ':where([invalid])';
```
anna-lach, line 2: Maybe we should use kebab case for variables naming?

packages/core/src/checkbox/checkbox.stories.ts
```diff
@@ -32,11 +125,27 @@ export const NoText: StoryObj = {
...
+      div {
+        display: flex;
+        flex-direction: column;
+        gap: 16px;
+      }
+      .wrapper {
+        border: 2px solid rgb(var(--sl-color-palette-accent-base));
+      }
+    </style>
+    <em>Borders are added to show the allignment in the container</em>
```
anna-lach, line 138: Small typo :) `allignment` => `alignment`

packages/core/src/radio-group/radio.scss
```diff
@@ -1,58 +1,191 @@
+// :host {
+//   --_background: #fff;
+//   --_foreground: #fff;
```
anna-lach, line 3: Do we want to keep those comments?

packages/core/src/radio-group/radio.scss
```diff
@@ -1,58 +1,191 @@
...
+//     border-radius: 100%;
+//     content: '';
+//     height: 100%;
+//     transition: background 0.2s ease-in-out;
+//     width: 100%;
+//   }
+// }
+
+$variants: default, valid, invalid;
+$variantselectors: ':not([valid],[invalid])', ':where([valid])', ':where([invalid])';
```
anna-lach, line 61: Kebab case for variables names?

anna-lach:
Disabled state doesn't look properly on Safari?

image

packages/core/src/checkbox/checkbox-group.spec.ts
Diaan: I don't really know how to get the test working :D Maybe Jeroen had an idea about this?

packages/core/src/radio-group/radio.scss
Diaan: This was just a work in progress, the rest was done in a separate branch :D

packages/core/src/checkbox/checkbox-group.ts
```diff
@@ -76,4 +86,14 @@ export class CheckboxGroup extends HintMixin(LitElement) {
       this.#rovingTabindexController.focus();
     }
   }
+
+  async #onSlotchange(): Promise<void> {
```
jpzwarte, line 90: Why `async`? I don't seen anything async in this method?

packages/core/src/checkbox/checkbox-group.ts
```diff
@@ -57,6 +66,7 @@ export class CheckboxGroup extends HintMixin(LitElement) {
 
   /** Custom validators. */
   @property({ attribute: false }) validators?: Validator[];
+  @property() name?: string;
```
jpzwarte, line 69: Please add a comment for each property. That way, those comments are automatically used as API documentation.

packages/core/src/checkbox/checkbox-group.ts
```diff
@@ -12,9 +13,17 @@ import {
...
 import { requiredValidator } from '../utils/index.js';
+import { Checkbox } from './checkbox.js';
 import styles from './checkbox-group.scss.js';
 
-export class CheckboxGroup extends HintMixin(LitElement) {
+export class CheckboxGroup extends HintMixin(ScopedElementsMixin(LitElement)) {
+  /** @private */
+  static get scopedElements(): ScopedElementsMap {
+    return {
+      'sl-checkbox': Checkbox
```
jpzwarte, line 23: `scopedElements` only refer to the custom elements that are rendered by *this* component (in the `render()` method). The checkbox isn't one of those right?

packages/core/src/checkbox/checkbox-group.spec.ts
jpzwarte: I will take a look

packages/core/src/checkbox/checkbox.scss
```diff
@@ -1,16 +1,18 @@
...
-  --_gap: 0.25rem;
-  --_line-height: 1.5rem;
-  --_path-length: 22.910259;
-  --_size: calc(1rem - var(--_border-size) * 2);
-  --_stroke: #fff;
+  --_line-height: var(--sl-text-typeset-line-height-md);
+  --_path-length: 23;
   --_stroke-dashoffset: var(--_path-length);
   --_transition: 200ms ease-in-out;
+  --_icon-size: 16px;
```
jpzwarte, line 15: This should come from a token (eventually), right?

packages/core/src/checkbox/checkbox.scss
```diff
@@ -20,27 +22,20 @@
   gap: var(--_gap);
   line-height: var(--_line-height);
   outline: none;
+  padding: var(--_padding);
 }
 
-:host(:disabled) {
+:host([disabled]) {
```
jpzwarte, line 28: I thought with `ElementInternals` we could use `:disabled`? 

From `button.scss`:

```sass
:host(:where(:disabled, [internals-disabled])) {
  --_cursor: default;
  pointer-events: none;
}
```

jpzwarte: `[internals-disabled]` is for the element internals polyfill.

packages/core/src/checkbox/checkbox.scss
```diff
@@ -52,25 +47,99 @@
...
   transition-property: background, border, box-shadow, fill;
-  width: var(--_size);
+  width: var(--_icon-size);
 }
 
 path {
   stroke: var(--_stroke);
   stroke-dasharray: var(--_path-length);
   stroke-dashoffset: var(--_stroke-dashoffset);
+  stroke-linecap: round;
```
jpzwarte, line 75: Shouldn't this be a css variable? Bingel = playful = round; Kampus = business = sharp?

packages/core/src/checkbox/checkbox.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+import './register.js';
+
+describe('sl-checkbox', () => {
+  let el: Checkbox;
+
+  describe('defaults', () => {
+    beforeEach(async () => {
+      el = await fixture(html`<sl-checkbox>Hello world</sl-checkbox>`);
+
+      el.disabled = false;
```
jpzwarte, line 14: Why?

packages/core/src/checkbox/checkbox.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+      await sendKeys({ press: 'Enter' });
+
+      expect(el.checked).to.equal(true);
+    });
+
+  });
+  
+  describe('disabled', () => {
+    beforeEach(async ()=>{
+      el = await fixture(html`<sl-checkbox>Hello world</sl-checkbox>`);
```
jpzwarte, line 54: Just do `<sl-checkbox disabled>Hello world</sl-checkbox>`?

packages/core/src/checkbox/checkbox.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+
+      el.disabled = true;
+      el.setAttribute('disabled', '');
+      await el.updateComplete;
+    });
+    it('should be disabled if set', async () => {
```
jpzwarte, line 60: Please add newline above; and remove below :)

packages/core/src/checkbox/checkbox.ts
```diff
@@ -8,6 +8,8 @@ import { event } from '../utils/decorators/index.js';
 import { FormControlMixin, HintMixin } from '../utils/mixins/index.js';
 import styles from './checkbox.scss.js';
 
+export type CheckboxSize = 'md' | 'lg';
```
jpzwarte, line 11: Just discussed this with @DanielleRameau : we probably need an `sm` size (eventually) as well.

packages/core/src/checkbox/checkbox.ts
```diff
@@ -33,9 +36,15 @@ export class Checkbox extends FormControlMixin(HintMixin(LitElement)) {
   /** Whether the checkbox is checked. */
   @property({ type: Boolean, reflect: true }) checked = false;
 
+  /** Whether the checkbox is invalid. */
+  @property({ type: Boolean, reflect: true }) invalid = false;
```
jpzwarte, line 40: So if you set an initial value like this, it will trigger the Lit change detection for this property. If you instead do `invalid?: boolean;` it will still be falsy, but it won't trigger change detection. That can be useful if changing the property has side-effects. So I usually go with the `?: boolean;` version.

packages/core/src/radio-group/radio.scss
```diff
@@ -1,58 +1,132 @@
...
+  --_stroke-dashoffset: var(--_path-length);
+  --_transition: 200ms ease-in-out;
+  --_icon-size: 16px;
 }
 
 :host {
   align-items: start;
-  cursor: pointer;
-  display: inline-flex;
+  display: flex;
```
jpzwarte, line 22: Same as with checkbox.

packages/core/src/radio-group/radio.scss
```diff
@@ -1,58 +1,132 @@
...
+  flex-direction: column;
   gap: var(--_gap);
   line-height: var(--_line-height);
   outline: none;
+  padding: var(--_padding);
```
jpzwarte, line 27: So recently I've been using `padding` less and more `padding-inline` & `padding-block` (same for `margin`). By using the logical versions of those properties, it's easier also to customise them since you can have separate variables for the inline & block versions. That was always a (potential) problem with a single css variable.

jpzwarte: Not saying you need to change anything in this case :)

packages/core/src/radio-group/radio.scss
```diff
@@ -1,58 +1,132 @@
...
+      var(--sl-space-input-inline-#{$size})
+      var(--sl-space-input-block-#{$size})
+      0;
+    --_size: var(--sl-size-input-#{$size});
+  }
+}
+
+@each $variant in $variants {
+  $i: index($variants, $variant);
+  $variantselector: nth($variant-selectors, $i);
```
jpzwarte, line 90: `$variant-selector` is a slightly better name?

packages/core/src/radio-group/radio.scss
```diff
@@ -1,58 +1,132 @@
...
+
+  @each $state in $states {
+    $j: index($states, $state);
+    $stateselector: nth($state-selectors, $j);
```
jpzwarte, line 94: Same

jpzwarte (approved):

packages/core/src/checkbox/checkbox.scss
Diaan: yes, but it's not available yet. we decided that based on how far/definite the icon component is we'll fix this in a separate story so this can be merged

packages/core/src/checkbox/checkbox.scss
Diaan: :disabled doesn't work in safari

jpzwarte: That's what the `[internals-disabled]` is for :)

Diaan: right, that does work... :D 
I only tried it with :disabled i think

packages/core/src/checkbox/checkbox.scss
Diaan: I've made it a variable, but there isn't a token for it. We're also not 100% sure how we're going to make the icon; svg in place or use sl-icon

packages/core/src/checkbox/checkbox.spec.ts
Diaan: that is quite an existential question to answer in a code review...

packages/core/src/checkbox/checkbox.ts
Diaan: Will add that when it's relevant. Right now we also don't have the tokens yet

packages/core/src/checkbox/checkbox-group.spec.ts
Diaan: we'll pick this up in a later story

PR #302, opened 2023-02-28
Feature/236 radio group component

Diaan:


anna-lach:
Disabled state doesn't look and work properly on Safari? We are able to check disabled radio:

image

packages/core/src/radio-group/radio.scss
```diff
@@ -1,62 +1,3 @@
...
-//     border-radius: 100%;
-//     content: '';
-//     height: 100%;
-//     transition: background 0.2s ease-in-out;
-//     width: 100%;
-//   }
-// }
-
 $variants: default, valid, invalid;
 $variantselectors: ':not([valid],[invalid])', ':where([valid])', ':where([invalid])';
```
anna-lach, line 2: Maybe we should use kebab case for variables naming?

PR #305, opened 2023-03-06
Feature/button tokens

Diaan:


packages/core/src/button/button.stories.ts
```diff
@@ -1,146 +1,130 @@
 import type { ButtonFill, ButtonSize, ButtonVariant } from './button.js';
 import type { Meta, StoryObj } from '@storybook/web-components';
 import '../button-bar/register.js';
+import '../icon/register.js';
 import './register.js';
 import { html } from 'lit';
 
 interface Props {
```
jpzwarte, line 8:
```ts
interface Props extends Pick<Button, 'disabled' | 'fill' | 'size' | 'variant'> {
  icon: string;
  text: string;
}
```

packages/core/src/button/button.stories.ts
```diff
@@ -1,146 +1,130 @@
...
   fill: ButtonFill;
   size: ButtonSize;
   text: string;
+  icon: string;
   variant: ButtonVariant;
 }
 
 type Story = StoryObj<Props>;
+const fills = ['default', 'outline', 'link'];
+const variants = ['default', 'primary', 'success', 'warning', 'danger'];
```
jpzwarte, line 18: I would at least add typing to these variables, so there no errors in the list: `const variants: ButtonVariant[] = ...`

packages/core/src/button/button.ts
```diff
@@ -87,4 +95,41 @@ export class Button extends LitElement {
       event.stopPropagation();
     }
   }
+
+  #onSlotChange(event: Event): void {
```
jpzwarte, line 99:
```ts
#onSlotchange(event: Event & { target: HTMLSlotElement }): void {
  const assignedNodes = event.target.assignedNodes(....);
  ...
```

packages/core/src/icon/icon.scss
```diff
@@ -1,8 +1,21 @@
...
+
 :host {
   --fill-default: currentColor;
   --fill-accent: rgb(var(--sl-color-palette-accent-300));
-  
-  display: inline-flex;
-  height: 2rem;
-  width: 2rem;
+   
+  display: grid;
```
jpzwarte, line 7: Not sure about this one; this will cause the icon component so have the full width of the parent? Shouldn't it be `inline-grid`?

jpzwarte (approved):

PR #324, opened 2023-03-13
Feature/301 website generating token tables for base tokens

anna-lach:


docs/website/site.eleventy.cjs
```diff
@@ -32,6 +32,54 @@ module.exports = function(eleventyConfig) {
 
   eleventyConfig.addFilter('search', searchFilter);
 
+  eleventyConfig.addFilter('replaceString', (value) => {
```
Diaan, line 35: is this the best name? i'm guessing all "addLiquidFilter" replace things in strings ;)

Diaan: plus,  i don't see it used anywhere :D

Diaan (approved):

docs/website/site.eleventy.cjs
anna-lach: yeah, right, this one I left by chance, it was an old one :)
And about liquid filters... I was trying to write it in a better way, but they were not working with more than one parameter.

PR #326, opened 2023-03-14
Feature/icons

Diaan:


packages/core/package.json
```diff
@@ -185,6 +185,7 @@
   },
   "dependencies": {
     "@floating-ui/dom": "^1.2.5",
+    "@fortawesome/fontawesome-svg-core": "^6.4.0",
```
jpzwarte, line 188: Do we need this as a runtime dependency? Would be nice if we didn't :)

packages/core/src/icon/icon.scss
```diff
@@ -2,4 +2,6 @@
   display: inline-flex;
   height: 2rem;
   width: 2rem;
+  --fill-default: currentColor;
```
jpzwarte, line 5: I usually put the CSS custom properties *before* the regular CSS properties. That's also the order if you sort the lines alphabetically :)

packages/core/src/icon/icon.stories.ts
```diff
@@ -1,11 +1,56 @@
...
+import type { IconStyle } from '@fortawesome/fontawesome-common-types';
 import { html } from 'lit';
 import './register.js';
+import { faPinata as falPinata } from '@fortawesome/pro-light-svg-icons';
+import { far } from '@fortawesome/pro-regular-svg-icons';
+import { faPinata as fasPinata } from '@fortawesome/pro-solid-svg-icons';
+import { faPinata as fatPinata } from '@fortawesome/pro-thin-svg-icons';
+import { Icon } from './icon.js';
+
+interface Props {
```
jpzwarte, line 11: So kind of nitpicking, but what I do is this:

```ts
interface Props<T = unknown> extends Pick<PageList<T>, 'items' | 'renderItem' | 'scopedElements'> {
  height?: string;
  styles?: string;
}
```

So pick the properties from the web component and then add any story-specific props you need.

packages/core/src/icon/icon.ts
```diff
@@ -1,36 +1,116 @@
 import type { CSSResultGroup, PropertyValues, TemplateResult } from 'lit';
+import type { IconPrefix, IconStyle } from '@fortawesome/fontawesome-common-types';
+import type { IconDefinition, IconName } from '@fortawesome/fontawesome-svg-core';
```
jpzwarte, line 3: What I did in DNA was to copy the types I needed into DNA. That way I did not have a runtime dependency on font awesome. Perhaps we should do the same? As long as the types match, typescript is fine.

packages/core/src/icon/icon.ts
```diff
@@ -1,36 +1,116 @@
...
+  [key: string]: SLIconDefinition | CustomIconDefinition;
+}
+
+interface SLIconDefinition {
+  value?: string;
+  type?: string;
+  style?: string;
+  description?: string;
+}
+interface CustomIconDefinition extends SLIconDefinition {
```
jpzwarte, line 22: Please add newlines between blocks.

packages/core/src/icon/icon.ts
```diff
@@ -1,36 +1,116 @@
...
+}
+
+declare global {
+  interface Window {
+    SLDS: {
+      icons: IconLibrary;
+    };
+  }
+}
+if (!window?.SLDS) {
```
jpzwarte, line 33: `window.SLDS ||= { icons: {}};` :)

packages/core/src/icon/icon.ts
```diff
@@ -1,36 +1,116 @@
...
+
+  /**
+   * Add icon(s) to the icon registry
+   *
+   * @param {IconDefinition | IconDefinition[] } faIcons One or more IconDefinition that have been imported from FontAwesome
+   */
+  static registerIcon(...faIcons: IconDefinition[]): void {
+    faIcons.forEach(icon => {
+      if (window.SLDS.icons[`${icon.prefix}-${icon.iconName}`]) {
+        console.warn(`Icon ${icon.prefix}-${icon.iconName} is already in the registry`);
```
jpzwarte, line 53: So what Lit does and Shoelace as well is that they have development/production environments. In development, the web components log various warnings etc to the console. In production they don't. It can be worth looking into how Shoelace did this and copy that :)

jpzwarte (approved):

PR #373, opened 2023-04-25
fix missing custom icons

Diaan:


packages/tokens/src/themes/bingel/icons.d.ts
```diff
@@ -41,4 +41,19 @@ export declare const icons: {
         description: string;
         svg: string;
     };
+    cirlce: {
```
jpzwarte, line 44: typo

.github/workflows/clean.yml
```diff
@@ -6,6 +6,7 @@ on:
 
 env:
   FONT_AWESOME_AUTH_TOKEN: ${{ secrets.FONT_AWESOME_AUTH_TOKEN }}
+  FIGMA_TOKEN: ${{ secrets.FIGMA_AUTH_TOKEN }}
```
jpzwarte, line 9: This line is unnecessary for the `clean` workflow?

jpzwarte (approved):

packages/tokens/src/themes/bingel/icons.d.ts
Diaan: image
It's like that in figma :D

.github/workflows/clean.yml
Diaan: for font awesome as well then?

jpzwarte: Yeah, good point.

jpzwarte: Or we need that token for `yarn install` to work?

packages/tokens/src/themes/bingel/icons.d.ts
Diaan: asked the designers to fix this :D

PR #412, opened 2023-05-10
Fix/393 icon issues

Diaan:


packages/components/icon/src/icon.ts
```diff
@@ -87,18 +121,22 @@ export class Icon extends LitElement {
...
-      return html`${unsafeHTML(this.#resolve(this.name))}`;
-    } else {
-      return html`${unsafeHTML(this.iconNotDef)}`;
-    }
+    return html`${unsafeHTML(this.iconHTML)}`;
   }
 
-  #resolve(name: string): string {
-    return this.icons[name] ? this.icons[name].svg : this.iconNotDef;
+  async waitForWindowProperty(): Promise<void> {
```
jpzwarte, line 130: Can you document this method please? For future reference :)

jpzwarte (approved):

PR #422, opened 2023-05-13
Add proper sort column icons

jpzwarte:


packages/components/grid/src/sorter.scss
```diff
@@ -7,3 +10,16 @@
   display: inline-flex;
   gap: var(--_gap);
 }
+
+:host([direction]) {
```
Diaan, line 14: in an other PR you moved declarations like this to the top of the file, ahead of `:host{}` should that be done here as well?

Diaan (approved):

packages/components/grid/src/sorter.scss
jpzwarte: Hmm, good question!

jpzwarte: So in this case you are absolutely right; but in some other cases it may not be as straightforward. Like if you have non-CSS custom properties in there. For example:

```css
:host(:where(:disabled, [internals-disabled])) {
  --_cursor: default;
  pointer-events: none;
}
```

jpzwarte: For this specific instance, I moved the declaration to the top.

PR #423, opened 2023-05-13
Fix transitions and other minor styling issues

jpzwarte:
- Only transition on `:active`, `:focus-visible` or `:hover`. This fixes a bug where components transition from their undefined state to their defined state, causing animation. You can see this on `main` where the initial button, checkbox and radio are black and then transition to their theme colors
- Fix checkbox and radio groups to align their children to `start`; fixes checkboxes & radios having an unnecessary large click area
- Fix `:hover` changes applying when hovering over the checkbox hint
- Fix magisters theme having incorrect colors for input focus state
- Clean up unnecessary styling rules


anna-lach:
I know that we have set transition ease-in-out, but does it look like now when hovering that transition is working like ease-in only? Like no ease-out? :)

packages/components/radio-group/src/radio.scss
```diff
@@ -125,3 +73,49 @@ circle {
     }
   }
 }
+
```
Diaan, line 76: do you put the loops at the top because they're only setting variables?

Diaan (approved):

packages/components/radio-group/src/radio.scss
jpzwarte: Yes, variables at the top, then the styles that use the variables.

jpzwarte: Otherwise you have some variables at the top and some at the bottom, and it's kinda not clear where you have to look.

jpzwarte:
> I know that we have set transition ease-in-out, but does it look like now when hovering that transition is working like ease-in only? Like no ease-out? :)

I did not actually change anything to the existing `transition` value. We have a separate issue though for the checkbox animations: https://github.com/sl-design-system/components/issues/325

anna-lach (approved):

PR #447, opened 2023-05-18
Fix linting being broken since the rename

jpzwarte:


package.json
```diff
@@ -162,9 +162,13 @@
     },
     "lint": {
       "dependencies": [
-        "./packages/core:lint"
+        "lint:components",
```
anna-lach, line 165: Maybe we can also add linting styles?

jpzwarte: Good idea! I'd like to do that in a separate PR though. The latest `stylelint` major has some pretty breaking changes.

anna-lach: ok, sure :)

PR #450, opened 2023-05-19
Add support for options in the grid filter column

jpzwarte:
- Add ability to filter on a column either by text or by selecting from a list of options
- Fix incorrectly inherited box-shadow in `<sl-checkbox>`
- Fix inability to type in an `<sl-popover>` due to `keydown` events being default prevented


anna-lach:
Maybe it will be nice when the icon for filtering has also box-shadow on focus? :)

anna-lach (approved):

jpzwarte:
> Maybe it will be nice when the icon for filtering has also box-shadow on focus? :)

I'm thinking of making the filter column header an actual `<sl-button>` (link button probably). But that requires a bit more visual design. But I'd like to delay that until Roald or Danielle have had time to do that.

PR #453, opened 2023-05-22
Add support for initial filtered/sorted columns

jpzwarte:


packages/components/grid/src/filter.ts
```diff
@@ -66,10 +68,17 @@ export class GridFilter extends ScopedElementsMixin(LitElement) {
...
+    } else {
+      this.#value = value;
+    }
+
+    this.active = Array.isArray(this.#value) ? this.#value.length > 0 : !!this.#value;
+    this.requestUpdate('value');
   }
 
-  @property()
+  @property({ type: String })
```
Diaan, line 81: I assume the support for filtering on numbers will come later?

packages/components/grid/src/filter.ts
```diff
@@ -51,7 +50,10 @@ export class GridFilter extends ScopedElementsMixin(LitElement) {
   /** The filter value(s). */
   #value?: string | string[];
 
-  /** Whether the grid is currently being filtered by this column. */
+  /**
+   * Whether the grid is currently being filtered by this column.
+   * @private
```
Diaan, line 55: can a property be private?

packages/components/grid/src/filter.ts
jpzwarte: Good point, but yes :)

PR #459, opened 2023-06-01
Feature/234 textarea component

anna-lach:
Styled `sl-textarea` component


packages/components/checkbox/src/checkbox.ts
```diff
@@ -121,7 +121,7 @@ export class Checkbox extends FormControlMixin(HintMixin(LitElement)) {
           <slot @slotchange=${() => this.#updateNoLabel()}></slot>
         </span>
       </div>
-      ${this.#validation.render()} ${this.#validation.render() ? this.#validation.render() : this.renderHint()}
+      ${this.#validation.render() ? this.#validation.render() : this.renderHint()}
```
jpzwarte, line 124: This looks weird; you're effectively rendering the validation *twice*? (the old code also looks weird)

packages/components/shared/src/mixins/form-control.ts
```diff
@@ -86,13 +86,24 @@ export function FormControlMixin<T extends Constructor<ReactiveElement>>(
       }
     }
 
+    override shouldUpdate(changedProperties: PropertyValues<this>): boolean {
```
jpzwarte, line 89: Please rename `changedProperties` to `changes`; that's what we use everywhere

jpzwarte (approved):

packages/components/checkbox/src/checkbox.ts
anna-lach: right... I need to find a better solution

packages/components/textarea/src/textarea.spec.ts
```diff
@@ -0,0 +1,351 @@
...
+    });
+
+    it('should have a wrap property of soft by default', () => {
+      expect(el).to.have.attribute('wrap', 'soft');
+    });
+
+    it('should have a wrap property of hard when set', () => {
+      el.setAttribute('wrap', 'hard');
+
+      expect(el).to.have.attribute('wrap', 'hard');
```
Diaan, line 67: does this actually test something? you only check if the attribute is what you just set it, not anything it affects.
If I do 
```
el.setAttribute('blablabla', 'hard');
expect(el).to.have.attribute('blablabla', 'hard');
```
it should pass as well, but i haven't really tested anything

packages/components/textarea/src/textarea.spec.ts
```diff
@@ -0,0 +1,351 @@
...
+      el.placeholder = 'my placeholder';
+      await el.updateComplete;
+
+      const textarea = el.querySelector('textarea');
+
+      expect(textarea).to.have.attribute('placeholder', 'my placeholder');
+    });
+
+    it('should not have a minLength by default', () => {
+      expect(el).not.to.have.attribute('minLength');
```
Diaan, line 108: does this check if there's an actual value that is being used in the component or just if the attribute is set?

packages/components/textarea/src/textarea.spec.ts
```diff
@@ -0,0 +1,351 @@
...
+
+      expect(el.textarea.getAttribute('maxlength')).to.be.null;
+    });
+
+    it('should not have minlength attribute when minLength property is not provided', async() => {
+      el.minLength = 2;
+      await el.updateComplete;
+
+      el.minLength = undefined;
+      await el.updateComplete;
```
Diaan, line 172: why set it to `2` and then to undefined without checking if the `2` is being set properly?

packages/components/textarea/src/textarea.spec.ts
```diff
@@ -0,0 +1,351 @@
...
+      expect(el.textarea.getAttribute('readonly')).to.be.null;
+    });
+  });
+
+  describe('resize of textarea', () => {
+    beforeEach(async () => {
+      el = await fixture(html`<sl-textarea></sl-textarea>`);
+    });
+
+    it('should set the textarea height to 100px if resize is auto and 100px height is set', async() => {
```
Diaan, line 204:
```suggestion
    it('should set the textarea to the value of the scrollHeight', async() => {
```
you don't actually test it will (in the real world) be 100px. The 100px is just a random value, it shouldn't be in the name of the test

Diaan (changes requested):

packages/components/textarea/src/textarea.ts
```diff
@@ -93,6 +148,40 @@ export class Textarea extends FormControlMixin(HintMixin(LitElement)) {
...
+        this.textarea.setAttribute('resize', this.resize);
+      } else {
+        this.resize = 'vertical';
+        this.textarea.setAttribute('resize', this.resize);
+      }
+    }
+
+    if (changes.has('readonly')) {
+      if (this.readonly) {
+        this.textarea.readOnly = this.readonly;
```
Diaan, line 170: when you set `this.textarea.readOnly` to `false` it removes the attribute, so this check shouldn't be necessary; 
```
if (changes.has('readonly')) {
    this.textarea.readOnly = this.readonly;
}
```
should work as well

packages/components/textarea/src/textarea.ts
```diff
@@ -93,6 +148,40 @@ export class Textarea extends FormControlMixin(HintMixin(LitElement)) {
...
+    }
+
+    if (changes.has('resize')) {
+      if (this.resize) {
+        this.textarea.setAttribute('resize', this.resize);
+      } else {
+        this.resize = 'vertical';
+        this.textarea.setAttribute('resize', this.resize);
+      }
+    }
```
Diaan, line 166:
```suggestion
    if (changes.has('resize')) {
      if (!this.resize) {
        this.resize = 'vertical';
      }
      this.textarea.setAttribute('resize', this.resize);
    }
```

Diaan (changes requested):

packages/components/textarea/src/textarea.spec.ts
anna-lach: If the attribute is not set by default.

packages/components/textarea/src/textarea.scss
```diff
@@ -1,9 +1,149 @@
...
-  --_border-color: #d3d3d3;
-  --_border-radius: 3px;
-  --_box-shadow: none;
-  --_gap: 0.25rem;
+  --_border-size: var(--sl-border-width-button-default);
+  --_border: var(--_border-size) solid var(--_border-color);
+  --_border-radius: var(--sl-border-radius-default);
+  --_box-shadow: var(--sl-box-shadow-none);
+  --_transition: 200ms ease-in-out;
+  --sl-textarea-min-height-md: 72px;
```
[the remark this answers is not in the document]
anna-lach: I guess that there can be some cases in the future, where someone wants to use a smaller one, that's why I wanted to add here CSS custom properties, but I set as default the values (height) we currently have in Figma.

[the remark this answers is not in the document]
jpzwarte: So there is a new CSS unit we could use here: `lh` (`line-height`)? See https://caniuse.com/mdn-css_types_length_lh
And perhaps with a fallback for Firefox?

packages/components/textarea/src/textarea.spec.ts
jpzwarte: The attribute name should be probably be `min-length` instead of `minLength`?

anna-lach: We don't want to have it consistent with the native textarea? So actually it should be `minlength` instead of `minLength ` :)

packages/components/textarea/src/textarea.scss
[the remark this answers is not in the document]
anna-lach: I wanted to add `rows` attribute first, but when we use rows, the height is inconsistent between different browsers, that's why I added here these custom properties for the min-height.

So should I try with `lh` instead of pixels the in the custom properties? Or there will be also inconsistent height between Firefox and other browsers?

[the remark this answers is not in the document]
jpzwarte: I would try something like this: `--_min-height: calc(var(--sl-textarea-rows) * 1lh);`
And a fallback for FF using perhaps `@supports`???? Not sure how that would work exactly atm.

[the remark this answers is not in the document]
anna-lach: Ok, I'll try this way :)

packages/components/checkbox/src/checkbox.ts
anna-lach: I've added some changes

packages/components/textarea/src/textarea.scss
```diff
@@ -4,13 +4,17 @@ $variant-selectors: ':not([valid],[invalid])', ':where([valid])', ':where([inval
...
   --_border: var(--_border-size) solid var(--_border-color);
   --_border-radius: var(--sl-border-radius-default);
   --_box-shadow: var(--sl-box-shadow-none);
   --_transition: 200ms ease-in-out;
-  --sl-textarea-min-height-md: 72px;
-  --sl-textarea-min-height-lg: 88px;
+  --_min-height: calc(var(--sl-textarea-rows) * 1lh);
+  // TODO: needs to be removed in the future when lh unit will be supported for Firefox as well
+  @supports (-moz-appearance: none) {
+    --_min-height: calc(var(--sl-textarea-rows) * 1.4em);
```
Diaan, line 16: isn't there a token that we can use for this? are you sure it's always 1.4em?

anna-lach: This is just an approximate value, I hope we will be able to remove it in the future, when the `lh` will be also supported by Firefox.
We have a line-hight and font-size in `px` which come from tokens, I'm not sure if it's possible for now to calculate it fast in css? Or maybe you have a better idea how to implement it in a better way?  Will be appreciated :)

anna-lach: Or maybe we should calculate it just with line-height which comes from token, but it's in px?
Sth like:
```
  @supports not (min-height: 1lh) {
    --_min-height: calc(var(--sl-textarea-rows) * var(--sl-text-typeset-line-height-md));
  }
  ```
  and of course `--sl-text-typeset-line-height-lg` for the large version?

anna-lach: I've just changed it this way.

Diaan (approved):

Diaan (approved):

PR #490, opened 2023-06-27
Update dependencies

jpzwarte:


jpzwarte:
@Diaan I noticed that `chromatic/package.json` includes `eslint` & `@typescript-eslint`, but I don't see it used anywhere. Are those packages necessary?

anna-lach (approved):

Diaan:
> @Diaan I noticed that `chromatic/package.json` includes `eslint` & `@typescript-eslint`, but I don't see it used anywhere. Are those packages necessary?

I haven't looked at everything the chromatic install added very critically, so if you say they can be removed I'm ok with that. (assuming it still works after 😉 )

Diaan (approved):

PR #499, opened 2023-07-05
Feature/272 select

Diaan:


packages/components/select/src/select.stories.ts
```diff
@@ -88,6 +198,9 @@ export const InForm: StoryObj = {
       sl-button-bar {
         grid-column: 1 / 3;
       }
+      sl-avatar {
+        margin: 0 4px;
+      }
     </style>
```
anna-lach, line 204: Maybe we can change layout of the example a little bit to have the error message under the select element?
image
And also labels above select elements to have it consistent with other form examples? :)

packages/components/select/src/select.ts
```diff
@@ -30,19 +37,26 @@ export class Select extends FormControlMixin(LitElement) {
     attributeOldValue: true
   };
 
-  @query('sl-select-overlay') overlay?: SelectOverlay;
   @query('#selectedOption') selectedOptionPlaceholder?: HTMLElement;
 
-  /** The slotted options. */
-  @queryAssignedElements({ selector: 'sl-select-option', flatten: false }) options?: SelectOption[];
+  /** The slotted options and option groups. */
+  // @queryAssignedElements({ selector: 'sl-select-option', flatten: false }) options?: SelectOption[];
```
anna-lach, line 43: Do we still need this comment?

packages/components/select/src/select.ts
```diff
@@ -30,19 +37,26 @@ export class Select extends FormControlMixin(LitElement) {
...
 
+  /** Select size. */
+  @property({ reflect: true }) size: SelectSize = 'md';
+
+  /** Whether the select is invalid. */
+  @property({ type: Boolean, reflect: true }) invalid?: boolean;
+
   #rovingTabindexController = new RovingTabindexController<SelectOption>(this, {
-    focusInIndex: (elements: SelectOption[]) => elements.findIndex(el => el.selected && !!isPopoverOpen(this.overlay)),
+    focusInIndex: (elements: SelectOption[]) => elements.findIndex(el => el.selected && !!isPopoverOpen(this.dialog)),
```
anna-lach, line 59: Maybe
`... && isPopoverOpen(this.dialog))` will be enough?

packages/components/select/src/select.ts
```diff
@@ -52,50 +66,75 @@ export class Select extends FormControlMixin(LitElement) {
   });
 
   #observer?: MutationObserver;
+  #anchor = new AnchorController(this);
```
anna-lach, line 69: Do we use this private anchor?

packages/components/select/src/select.stories.ts
```diff
@@ -16,42 +19,149 @@ const onSubmit = (event: Event & { target: HTMLFormElement }): void => {
...
+                      ${disabledState ? 'Disabled' : 'Enabled'}
+                    </td>`
+                )}
+              </tr>
+            </thead>
+            <tbody>
+              ${states.map(
+                state => html` <tr>
+                  <th>${state}</th>
+                  ${disabledStates.map(
```
anna-lach, line 123: Selected option disabled has a blue or grey border on the left?
In the story we have a version with blue:
image


But in Figma with grey?
image

packages/components/select/src/select.ts
```diff
@@ -52,50 +66,75 @@ export class Select extends FormControlMixin(LitElement) {
...
+        popover
+        ${anchor({ position: 'bottom' })}
         aria-labelledby=${this.#selectId}
-        style="--max-overlay-height:${this.maxOverlayHeight || 'unset'}"
+        role="listbox"
       >
         <slot @slotchange=${this.#handleOptionsSlotChange}></slot>
-      </sl-select-overlay>
+      </dialog>
       ${this.#validation.render()}
```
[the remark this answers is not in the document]
Diaan: I don't think we do; there was a discussion about that right? that you can't see the hint anymore when you open the dropdown. And it shouldn't be needed, since the fact that you have a list of option is already guiding the user on what needs to be filled in.
@RoaldBoerema do you agree with that / did i remember it correctly?

.changeset/fresh-donkeys-beam.md
```diff
@@ -0,0 +1,19 @@
...
+'@sl-design-system/magister': patch
+'@sl-design-system/myvanin': patch
+'@sl-design-system/bingel': patch
+'@sl-design-system/kampus': patch
+'@sl-design-system/neon': patch
+'@sl-design-system/teas': patch
+'@sl-design-system/max': patch
+---
+
+added select component
```
jpzwarte, line 19: Should this also contain the theme packages? Aren't they already listed below?

package.json
```diff
@@ -324,5 +324,8 @@
     "typescript": "^5.2.0-beta",
     "vite": "^4.3.9",
     "wireit": "^0.9.5"
+  },
+  "dependencies": {
+    "@floating-ui/dom": "^1.3.0"
```
jpzwarte, line 329: Why here? Also the root package.json should only contain `devDependencies`

packages/components/popover/src/popover.stories.ts
```diff
@@ -73,8 +73,55 @@ export const Dialog: Story = {
...
+          }
+
+          [anchor]:is(:popover-open) {
+            inset: auto;
+          }
+        }
+      </style>
       <sl-button @click=${onClick} id="button">Toggle popover</sl-button>
-      <dialog anchor="button" popover ${anchor()}>I'm a popover</dialog>
+      <dialog anchor="button" popover ${anchor({ position: 'bottom' })} position="bottom">I'm a popover</dialog>
```
jpzwarte, line 124: Why the double `position`?

packages/components/select/src/select-option-group.ts
```diff
@@ -7,11 +8,14 @@ export class SelectOptionGroup extends LitElement {
   /** @private */
   static override styles: CSSResultGroup = styles;
 
+  /** Select size. */
+  @property({ reflect: true }) size: SelectSize = 'md';
+
   @property({ attribute: 'group-title' }) groupTitle?: string;
```
jpzwarte, line 14: Wouldn't it be more consistent if this were called `group-heading`/`groupHeading`? In other places I try to avoid `title` since that triggers a native tooltip.

packages/components/select/src/select-option-group.ts
```diff
@@ -21,4 +25,10 @@ export class SelectOptionGroup extends LitElement {
     this.setAttribute('role', 'group');
     this.setAttribute('aria-label', this.groupTitle || '');
   }
+
+  #groupHeading(): TemplateResult | void {
```
jpzwarte, line 29: This tiny a method could just be inlined in the main `render()` method imo.

packages/components/select/src/select.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+      `Finland`,
+      `The Netherlands`,
+      `Somewhere else`,
+    ];
+    describe('nothing selected', () => {
+      beforeEach(async () => {
+        el = await fixture(html`
+          <sl-select>
+            <sl-select-option>Won't say</sl-select-option>
+            <sl-select-option-group title="europe">
```
jpzwarte, line 24: Doesn't `title` here cause a native tooltip?

packages/components/select/src/select.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+              <sl-select-option>Finland</sl-select-option>
+              <sl-select-option>The Netherlands</sl-select-option>
+            </sl-select-option-group>
+            <sl-select-option>Somewhere else</sl-select-option>
+          </sl-select>`);
+      });
+      afterEach(async () => {
```
jpzwarte, line 31: There is nothing `async` in this method?

packages/components/select/src/select.stories.ts
```diff
@@ -16,42 +19,149 @@ const onSubmit = (event: Event & { target: HTMLFormElement }): void => {
...
+    <style>
+      sl-select {
+        width: 400px;
+        max-width: 90vw;
+        display: inline-flex;
+      }
+    </style>
     <sl-button>To focus</sl-button>
-    <sl-select maxOverlayHeight=${maxOverlayHeight}>
+    <sl-select maxOverlayHeight=${maxOverlayHeight} .size="${size}">
```
jpzwarte, line 53: what is `maxOverlayHeight`? There is no dot in front of it, so it isn't a property? And attributes names should be snake-case

packages/components/select/src/select.ts
```diff
@@ -30,19 +37,26 @@ export class Select extends FormControlMixin(LitElement) {
...
+  /** The slotted options and option groups. */
+  // @queryAssignedElements({ selector: 'sl-select-option', flatten: false }) options?: SelectOption[];
   @queryAssignedElements({ selector: 'sl-select-option-group', flatten: false }) optionGroups?: SelectOptionGroup[];
 
+  @query('button') button?: HTMLButtonElement;
+  @query('dialog') dialog?: HTMLDialogElement;
+
   /** render helpers */
-  @property() size?: { width: string; height: string } = { width: '500px', height: '32px' };
   @property() maxOverlayHeight?: string;
```
jpzwarte, line 50: This should at least have `{ attribute: 'max-overlay-height' }`

jpzwarte (approved):

anna-lach (approved):

Issue #530, opened 2023-07-18
[POC] Dev of skeleton


anna-lach:
Skeleton is an animated placeholder which simulates the layout while content is being loaded.

```[tasklist]
### Acceptance criteria
- [x] Should be available in one of effects (or none): shimmer / pulse / sheen
- [x] Should be accessible
- [x] Should have some basic styling
- [x] Should be presented in stories examples in the Storybook
```

DanielleRameau:
@anna-lach please describe the goal of the POC

PR #542, opened 2023-07-21
Feature/530 PoC of skeleton component

anna-lach:
#530 
POC of skeleton component


packages/components/skeleton/src/skeleton.scss
```diff
@@ -0,0 +1,70 @@
...
+  0% {
+    background-position: -2000px 0;
+  }
+
+  100% {
+    background-position: 2000px 0;
+  }
+}
+
+@keyframes sheen-effect {
```
Diaan, line 48: this animation makes me a bit nervous :D it's so fast compared to the other ones, even though they share the "animation-duration".

Diaan (approved):

packages/components/skeleton/register.ts
```diff
@@ -0,0 +1,9 @@
+import { Skeleton } from './src/skeleton';
```
jpzwarte, line 1: Add `.js` extension plz. This helps when you run the code directly in the browser (without a bundler).

packages/components/skeleton/src/skeleton.scss
```diff
@@ -0,0 +1,70 @@
+:host {
+  --background: rgb(var(--sl-color-palette-neutral-150));
```
jpzwarte, line 2: Please use `--_` notation since these variables are "private" to this file

packages/components/skeleton/src/skeleton.ts
```diff
@@ -0,0 +1,15 @@
+import type { CSSResultGroup } from 'lit';
+import { LitElement } from 'lit';
+import { property } from 'lit/decorators.js';
+import styles from './skeleton.scss.js';
+
+export class Skeleton extends LitElement {
+  static override styles: CSSResultGroup = styles;
```
jpzwarte, line 7: Please add a `/** @private */` comment so this isn't marked as public API.

packages/components/skeleton/src/skeleton.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  /** Skeleton's effect. */
+  @property({ reflect: true }) effect: 'none' | 'shimmer' | 'pulse' | 'sheen' = 'shimmer';
```
jpzwarte, line 10: Please extract these types into an `SkeletonEffect` type. Users of the component can then use that type themselves.

packages/components/skeleton/src/skeleton.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  override firstUpdated(): void {
+    this.setAttribute('aria-busy', 'true');
```
jpzwarte, line 13: Please move this to the `constructor()` (after `super()`). No reason to wait until `firstUpdated`?

jpzwarte (approved):

PR #555, opened 2023-08-02
Feature/531 website components pages structure update

Diaan:


anna-lach:
Maybe we should show explicitly hintSize values as well instead of `HintSize` type?

image

packages/components/shared/src/mixins/form-control.ts
```diff
@@ -78,6 +80,7 @@ export function FormControlMixin<T extends Constructor<ReactiveElement>>(
       }
     }
 
+    /** Native labels property. */
```
anna-lach, line 83: Maybe we should add here `@type NodeListOf<HTMLLabelElement> | null ...` as well?
Because it's not rendering properly:

image

packages/components/icon/src/icon.ts
```diff
@@ -54,29 +52,32 @@ export class Icon extends LitElement {
   }
 
   /**
-   * store all icons from the IconLibrary of the theme in the icon registry for easy access
+   * Store all icons from the IconLibrary of the theme in the icon registry for easy access.
```
anna-lach, line 55: Do we want to show here for the documentation purposes what exactly `IconLibrary` interface contains?

packages/components/radio-group/src/radio-group.ts
```diff
@@ -65,21 +78,22 @@ export class RadioGroup extends FormControlMixin(HintMixin(LitElement)) {
...
-  /** Element internals. */
+  /** @private Element internals. */
   readonly internals = this.attachInternals();
 
-  /** The assigned nodes. */
+  /** @ignore The assigned nodes. */
   @queryAssignedNodes() defaultNodes?: Node[];
 
-  /** If true, displays the radio buttons next to each other instead of below. */
+  /** The orientation of the radio options; when true, the radio buttons are desplayed next to each other instead of below eachother. */
```
anna-lach, line 87: small typo, `displayed`

anna-lach: and `each other`

packages/components/select/src/select-option-group.ts
```diff
@@ -4,13 +4,29 @@ import { LitElement, html, nothing } from 'lit';
...
+ *  <sl-select>
+      <sl-select-option-group group-heading="Europ">
+      <sl-select-option>Netherlands</sl-select-option>
+      <sl-select-option>Spain</sl-select-option>
+        <sl-select-option>Poland</sl-select-option>
+      </sl-select-option-group>
+    </sl-select>
+ * ```
+ *
+ * @slot default - List of sl-options
```
anna-lach, line 20: Maybe `<code>sl-options</code>`? :)

packages/components/select/src/select.ts
```diff
@@ -36,20 +37,26 @@ export class Select extends FormControlMixin(LitElement) {
...
   @property({ attribute: 'max-overlay-height', reflect: true }) maxOverlayHeight?: string;
 
-  /** Select size. */
+  /** The size of the select.
+   *  @type {'md' | 'lg'}
+   */
   @property({ reflect: true }) size: SelectSize = 'md';
 
-  /** Placeholder to show when no options is chosen. */
+  /** The placeholder text to show when no options is chosen. */
```
anna-lach, line 59: `no options are chosen`?

packages/components/icon/src/icon.ts
Diaan: I think the whole process of setting up/loading the icons will need to be described separately. I don't think we can get the full explanation in this column

Issue #511, opened 2023-07-11
Development of component overview page


DanielleRameau:
```[tasklist]
### Tasks
- [x] make cards for components (with image)
- [x] make sections (with heading for a groups of components)
- [x] available in light and dark mode
- [x] animation on hover: zoom in similar to the one on the homepage
- [x] make sure the page is responsive in the same way as the design
- [x] make the card that are still in PoC/coming state not clickable and use a placeholder image
- [x] exclude 'Guidelines' from the navigation for now
- [x] improve code sections styling in the dark mode version of 'Guidelines'
- [x] Paragraphs should have max-width of 640px
```

Out of scope:
- footer
- side menu

Design: 
https://www.figma.com/file/fpvv1aBzMNVTbJaFHhraB9/SL---Web-Generator?type=design&node-id=1230%3A2500&mode=dev

PR #583, opened 2023-08-28
Feature/511 website component overview page

anna-lach:
#511 (website) component overview page, excluded 'guidelines' from the navigation for now


website/src/site/styles/categories/components.scss
```diff
@@ -1,40 +1,167 @@
...
+    /* stylelint-disable-next-line max-nesting-depth */
+    .component-card__picture {
+      display: none;
+    }
+  }
 }
 
 .ds-components__content {
+  display: grid;
+  grid-column-gap: 7.2rem;
```
jpzwarte, line 113: You can just use `gap` for both grid & flexbox.

website/src/site/styles/components/table.scss
```diff
@@ -34,7 +34,8 @@
       background-color: var(--table-code-background);
       border-radius: 0.6rem;
       box-decoration-break: clone;
-      box-decoration-break: clone;
+      /* stylelint-disable-next-line property-no-vendor-prefix */
+      -webkit-box-decoration-break: clone;
```
jpzwarte, line 38: I think you should have the vendor prefixes before the regular properties? So:

```css
-webkit-box-decoration-break: clone;
box-decoration-break: clone;
```

So it will will always use the non-prefixed one when available.

jpzwarte (approved):

PR #642, opened 2023-09-11
Feature/288 tooltip component

Diaan:


packages/components/shared/src/controllers/anchor.ts
```diff
@@ -24,7 +25,8 @@ export class AnchorController implements ReactiveController {
       if (anchorElement) {
         this.#cleanup = positionPopover(host, anchorElement, {
           arrow: this.#config?.arrow,
-          position: this.position ?? this.#config.position ?? 'top'
+          position: this.position ?? this.#config.position ?? 'top',
+          maxWidth: this.maxWidth ?? this.#config.maxWidth ?? undefined
```
jpzwarte, line 29: This feels superfluous. If `this.maxWidth` is `null` or `undefined`, it will be `undefined`? I think you could just do `maxWidth: this.maxWidth`?

packages/components/shared/src/popover.ts
```diff
@@ -137,7 +138,7 @@ export const positionPopover = (
           isConstrained = actualHeight < initialHeight || maxHeight <= actualHeight;
           const appliedHeight = isConstrained ? `${maxHeight}px` : '';
           Object.assign(element.style, {
-            maxWidth: `${Math.floor(availableWidth)}px`,
+            maxWidth: `${options.maxWidth ? options.maxWidth : Math.floor(availableWidth)}px`,
```
jpzwarte, line 141: `options.maxWidth ?? Math.floor(availableWidth)`

packages/components/tooltip/src/tooltip.scss
```diff
@@ -1,28 +1,133 @@
...
+  --_arrow-height: 6px;
+  --_arrow-width: 16px;
+  /* the distance from the edge when the arrow is positionend at start or end */
+  --_arrow-start-end-offset: var(--sl-space-tooltip-arrow-offset);
 }
 
 :host {
-  // This has to be in pixels until we can use @property
-  --sl-popover-offset: 8px;
+  /** This has to be in pixels until we can use @property.
```
jpzwarte, line 20: Perhaps check if we can use `@property` with a fallback for FF/older browsers? https://caniuse.com/mdn-css_at-rules_property

packages/components/tooltip/src/tooltip.scss
```diff
@@ -1,28 +1,133 @@
...
+
+:host([actual-placement='bottom-start']) .arrow {
+  inset-inline-start: var(--_arrow-start-end-offset);
+}
+
+:host([actual-placement='bottom-end']) .arrow {
+  inset-inline-end: var(--_arrow-start-end-offset);
+}
+
+// Left
```
jpzwarte, line 62: Kind of unnecessary :)

jpzwarte (approved):

packages/components/tooltip/src/tooltip.scss
Diaan: I thought all the blocks looked so similar I added a heading to provide some clarity when scanning over the document.

jpzwarte: On the next line it also says `left`. But I guess this is personal :)

packages/components/tooltip/src/tooltip.scss
jpzwarte: Should probably be done in a different PR.

Diaan: Created a story for this: #656

PR #654, opened 2023-09-20
getting started page

Diaan:


website/src/site/_includes/components/components.njk
```diff
@@ -8,7 +8,7 @@ layout: base.njk
     <p class="ds-component__heading-description">{{ description }}</p>
   </div>
 </header>
-<section>
+<section class="for-collections-component">
```
anna-lach, line 11: Do we still need that class? :)

website/src/site/categories/getting-started/designers.md
```diff
@@ -0,0 +1,70 @@
...
+</div>
+</header>
+
+<section class="ds-subpage-section">
+
+<div class="ds-subpage-section__wrapper">
+
+<section>
+
+## Loading the SL Design System Library
```
anna-lach, line 23: We have a little bit different spacing in Figma than on our website:

image


Do we want to change it in that task, for that subpage only for now or maybe later globally?

anna-lach:
On the homepage we had before a link to the guidelines which we will not have for now, maybe it should be a link to getting started page?

image

website/src/site/categories/getting-started/designers.md
Diaan: I know, i discussed with Danielle that we will keep it the same as the spacing and headings (h2 and not h1 as in Figma) as we now have on the component detail page.

Diaan:
> On the homepage we had before a link to the guidelines which we will not have for now, maybe it should be a link to getting started page?
> 
> image

Good idea, but the problem is that there's no "overview" page for "Getting started" so not sure to which page it should link.
@DanielleRameau any thoughts on this?

Diaan:
We will change the "guidelines" button in a different story

PR #663, opened 2023-09-22
removed tabs + fixed link colours

Diaan:
Left the source files for all components because we will be using them in the future. They don't get served/loaded for end-users so they don't impact the website

### DIFF
--- a/website/src/site/styles/layout/content.scss
+++ b/website/src/site/styles/layout/content.scss
@@ -50,6 +50,7 @@ $transition-timing: 300ms;
 }
 
 a.header-anchor {
+  color: inherit;
   font-size: inherit;
   text-decoration: none;
 }
--- a/website/src/site/styles/layout/content.scss
+++ b/website/src/site/styles/layout/content.scss
@@ -26,24 +26,7 @@ $transition-timing: 300ms;
   }
 }
 
-.ds-subpage-section {
-  display: grid;
-  grid-column-gap: 7.2rem;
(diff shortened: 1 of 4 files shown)


website/src/site/styles/layout/content.scss
```diff
@@ -26,24 +26,7 @@ $transition-timing: 300ms;
...
-
-  @media screen and (min-width: $xxl-breakpoint) {
-    grid-template-columns: 88rem 1fr;
-  }
-
-  h2 {
-    margin-top: 2.4rem;
-  }
-
+*:where(.ds-subpage-section, .ds-tabs__tab-container) {
```
anna-lach, line 29: One minor thing, this `header-anchor` should not have text-decoration as well :) actually all `header-anchor` elements should have no text-decoration:

image

anna-lach: So these ones as well (all `header-anchor` elements):

image

website/src/site/styles/layout/content.scss
```diff
@@ -63,8 +46,30 @@ $transition-timing: 300ms;
 
   a.header-anchor {
     color: var(--font-color);
-    font-size: inherit;
-    text-decoration: none;
+  }
+}
+
+a.header-anchor {
+  font-size: inherit;
```
anna-lach, line 53: I think that the colour should be inherit as well:

image

anna-lach (approved):

PR #673, opened 2023-10-02
Feature/171 launch button component version api

Diaan:


packages/components/button/src/button.stories.ts
```diff
@@ -118,6 +125,7 @@ export const All: Story = {
                               .fill=${fill}
                               .size=${size}
                               ?disabled=${disabledState}
+                              aria-disabled="true"
```
jpzwarte, line 128: Should this be here?

jpzwarte (approved):

PR #679, opened 2023-10-09
Feature/275 dialog component

anna-lach:
Dialog component


packages/components/dialog/src/dialog.ts
```diff
@@ -14,50 +18,89 @@ import styles from './dialog.scss.js';
...
 
   override render(): TemplateResult {
     return html`
       <dialog
         @cancel=${this.#onCancel}
         @click=${this.#onClick}
+        @keydown=${this.#onKeydown}
         @close=${this.#onClose}
         .role=${this.role}
+        aria-labelledby="titleId"
```
jpzwarte, line 78: Why `titleId` and not just `title`? Having an `id` with name `titleId` seems kind of excessive :)

packages/components/dialog/src/dialog.ts
```diff
@@ -14,50 +18,89 @@ import styles from './dialog.scss.js';
...
+
+    super.disconnectedCallback();
   }
 
   override render(): TemplateResult {
     return html`
       <dialog
         @cancel=${this.#onCancel}
         @click=${this.#onClick}
+        @keydown=${this.#onKeydown}
```
jpzwarte, line 75: Personally I try to keep things sort alphabetically: so `@keydown` *after* `@close` :)

packages/components/dialog/src/dialog.ts
```diff
@@ -14,50 +18,89 @@ import styles from './dialog.scss.js';
...
         </slot>
         <slot name="body">
-          <slot></slot>
+          <div class="body-wrapper">
+            <slot></slot>
+          </div>
         </slot>
         <slot name="footer">
-          <sl-button-bar align="end"><slot name="action"></slot></sl-button-bar>
+          <sl-button-bar class="footer-buttons" align=${this.buttonsAlign}><slot name="action"></slot></sl-button-bar>
```
jpzwarte, line 103: Do you need the class name here? Afaics there is no other `<sl-button-bar>` here? So the class name is not really necessary.

jpzwarte: Also `header-buttons` is a slot name. So having a class name called `footer-buttons` can be confusing.

packages/components/dialog/src/dialog.ts
```diff
@@ -14,50 +18,89 @@ import styles from './dialog.scss.js';
...
         <slot name="header">
-          <slot name="title"></slot>
+          <slot name="titles">
+            <slot name="subtitle"></slot>
+            <slot name="title" id="titleId"></slot>
+          </slot>
+          <slot name="header-actions">
+            <slot name="header-buttons"></slot>
+            ${this.closingButton
+              ? html`<slot name="close" @click=${this.#onCloseClick}>
```
jpzwarte, line 89: If you "help" eslint/prettier along by having a newline just before `<slot>` and after the closing backtick, then prettier will put all the markup on new lines.

packages/components/dialog/src/dialog.ts
```diff
@@ -14,50 +18,89 @@ import styles from './dialog.scss.js';
...
 
+  /** Determines whether closing button (default one) should be shown in the top right corner. */
+  @property({ type: Boolean, attribute: 'closing-button' }) closingButton?: boolean;
+
   /** The ARIA role of the dialog. */
   @property() override role: 'dialog' | 'alertdialog' = 'dialog';
 
+  /** Alignment of the action buttons in the dialog footer
+   * @type {'start' | 'end'} */
+  @property({ reflect: true, attribute: 'buttons-align' }) buttonsAlign: Extract<ButtonBarAlign, 'start' | 'end'> =
```
jpzwarte, line 53: Not sure if `buttons-align` is correct, or if it should be `align-buttons`. The latter feels more natural.

packages/components/dialog/src/dialog.ts
```diff
@@ -83,25 +141,57 @@ export class Dialog extends ScopedElementsMixin(LitElement) {
     }
   }
 
+  #handleAnimationEnd = (event: AnimationEvent, target?: HTMLElement): void => {
```
jpzwarte, line 144: Why `handle` instead of `on`? The convention I use everywhere is to name event handler with the prefix `#on`; so `#onAnimationEnd`

jpzwarte (approved):

packages/components/dialog/src/dialog.ts
jpzwarte: What's the idea behind this property btw? Why would a developer change this setting? Or is this a theme variable? If the latter is the case, then having (`static`) dialog defaults would be better I think.

You could also remove the property and simply have the developer slot his/her own `<sl-button-bar>` in the `footer` slot?

anna-lach: We have the acceptance criteria
image

that in the footer we can have buttons on the left and on the right side. It doesn't come from the token. So do we want to keep it as it is or do you suggest that we should let the developer using `sl-button-bar` on his/her own in the `footer` slot?

anna-lach: The possibilities of the footer usage we can find in Figma: https://www.figma.com/file/1oIVcD9VMbOtLXYp48ehsA/Sanoma-Learning-Light-mode?type=design&node-id=4801-22057&mode=dev

We have now two slots inside `sl-button-bar` .

packages/components/dialog/src/dialog.ts
```diff
@@ -14,31 +17,48 @@ import styles from './dialog.scss.js';
...
+
   /** The ARIA role of the dialog. */
   @property() override role: 'dialog' | 'alertdialog' = 'dialog';
 
   override connectedCallback(): void {
     super.connectedCallback();
 
     this.inert = true;
+
+    this.dialog?.setAttribute('closing', 'false');
```
jpzwarte, line 55: I'm not sure if `this.dialog` already exists at this point. Perhaps not until after `firstUpdated`?

packages/components/dialog/src/dialog.ts
```diff
@@ -14,31 +17,48 @@ import styles from './dialog.scss.js';
...
+  }
+
+  override disconnectedCallback(): void {
+    this.dialog?.removeEventListener('animationend', event => this.#onAnimationEnd(event));
```
jpzwarte, line 59: This won't work: the callback you pass to `removeEventListener` must be the same callback (same function) you passed to `addEventListener`. You can achieve that by doing:

```ts
#onAnimationEnd = (event: AnimationEvent): void => {
  ...
}
```

And then using `this.#onAnimationEnd` as the second argument in `add/removeEventListener`.

packages/components/dialog/src/dialog.ts
```diff
@@ -47,17 +67,38 @@ export class Dialog extends ScopedElementsMixin(LitElement) {
...
         @click=${this.#onClick}
         @close=${this.#onClose}
+        @keydown=${this.#onKeydown}
         .role=${this.role}
+        aria-labelledby="title"
         part="dialog"
       >
         <slot name="header">
-          <slot name="title"></slot>
+          <slot name="titles">
```
jpzwarte, line 76: `aria-labelledby` will no longer work if you slot something in the `titles` slot. Not sure if/how we should handle that....

packages/components/dialog/src/dialog.ts
```diff
@@ -47,17 +67,38 @@ export class Dialog extends ScopedElementsMixin(LitElement) {
...
+                      <sl-icon name="xmark"></sl-icon>
+                    </sl-button>
+                  </slot>
+                `
+              : nothing}
+          </slot>
         </slot>
         <slot name="body">
-          <slot></slot>
+          <div class="body-wrapper">
```
jpzwarte, line 94: Should this also have a `part`?

packages/components/dialog/src/dialog.ts
```diff
@@ -69,6 +110,21 @@ export class Dialog extends ScopedElementsMixin(LitElement) {
 
     // Disable scrolling while the dialog is open
     document.documentElement.style.overflow = 'hidden';
+
+    this.dialog?.setAttribute('closing', 'false');
+
+    /** Workaround for the backdrop background */
```
jpzwarte, line 116: Please provide a little more explanation why this is necessary for our future selves ;)

jpzwarte (approved):

packages/components/dialog/src/dialog.ts
anna-lach: Maybe we don't nedd `titles` slot, maybe slot `title` and `subtitle` would be enough.

This is how it looks like in Figma:

image

jpzwarte: I agree: I think a `header` slot is enough.

Diaan (changes requested):

packages/components/dialog/src/dialog.ts
```diff
@@ -47,17 +59,38 @@ export class Dialog extends ScopedElementsMixin(LitElement) {
...
         <slot name="body">
-          <slot></slot>
+          <div class="body-content" part="body-content">
+            <slot></slot>
+          </div>
         </slot>
         <slot name="footer">
-          <sl-button-bar align="end"><slot name="action"></slot></sl-button-bar>
+          <sl-button-bar>
+            <slot name="action"></slot>
```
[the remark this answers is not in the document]
anna-lach: Actually it is already there, in line 16. :) But yes, I'm going to rename it to `actions` :)

packages/components/dialog/src/dialog.scss
```diff
@@ -5,26 +5,36 @@ $md-breakpoint: 600px;
...
   --_max-inline-size-mobile: var(--sl-size-dialog-full);
-  --_min-inline-size-desktop: var(--sl-size-dialog-min);
+  --_min-inline-size: var(--sl-size-dialog-min);
   --_max-inline-size-desktop: var(--sl-size-dialog-max);
+  --_max-inline-size: var(--_max-inline-size-mobile);
   --_padding-block-mobile: var(--sl-space-dialog-mobile-body-block);
   --_padding-inline-mobile: var(--sl-space-dialog-mobile-body-inline);
   --_padding-block-desktop: var(--sl-space-dialog-desktop-body-block);
   --_padding-inline-desktop: var(--sl-space-dialog-desktop-body-inline);
+  --_padding-block: var(--_padding-block-mobile);
```
Diaan, line 27: i meant actually remove all the specific mobile and desktop variables, just use the token directly.

packages/components/dialog/src/dialog.scss
```diff
@@ -118,48 +127,48 @@ slot[name='title']::slotted(*) {
...
-    padding-inline: var(--_padding-inline-desktop);
+    --_border-radius: var(--_border-radius-desktop);
+    --_padding-block: var(--_padding-block-desktop);
+    --_padding-inline: var(--_padding-inline-desktop);
+    --_margin-bottom-container: var(--_margin-bottom-container-desktop);
+    --_margin-top-container: var(--_margin-top-container-desktop);
+    --_max-inline-size: var(--_max-inline-size-desktop);
+    --_margin-container-inline: var(--_margin-container-inline-desktop);
+    --_body-gap: var(--_body-gap-desktop);
+    --_footer-direction: var(--_footer-direction-desktop);
```
Diaan, line 153: move this to the top, just underneath the rest of the variables

Diaan (approved):

PR #689, opened 2023-10-16
Feature/681 update homepage

Diaan:


jpzwarte (approved):

website/src/styles/layout/content.scss
```diff
@@ -3,8 +3,11 @@
 $transition-timing: 300ms;
```
anna-lach, line 4: layout on the component detail page is a little bit broken on a smaller resolution (< 600px)

image

website/src/styles/tokens-core.scss
```diff
@@ -69,19 +69,19 @@
...
-  --scale-400-scale: 24px;
-  --scale-500-scale: 28px;
-  --scale-600-scale: 32px;
-  --scale-700-scale: 40px;
-  --scale-800-scale: 48px;
-  --scale-900-scale: 48px;
-  --scale-1000-scale: 56px;
-  --scale-1200-scale: 64px;
-  --scale-1300-scale: 80px;
+  --scale-none-scale: 0rem;
```
anna-lach, line 72: just `0` should be enough?

Diaan: these are copy pasted from figma, i don't think we should change too much in that, because it will be overwritten next time we update the tokens. and i don't think it's a very big problem...

PR #693, opened 2023-10-19
Remove teaser website

jpzwarte:
- Remove the teaser site
- Move the main site from `src/site` to `src`
- Remove the separate DEV/PROD builds and always just output to dist (and build)
- Remove the unused SSR stuff from the build
- Simplify the wireit scripts
- Properly specify files and output for the wireit tasks, so the tasks do not loop when run with --watch
- Fix `scripts/build-cem.js` to incorrectly also analyze `*.d.ts` files
- Cleanup


website/esbuild.config.js
```diff
@@ -63,26 +63,27 @@ if (DEV) {
...
-  .catch(() => process.exit(1));
+// // this code is inlined into the HTML because it is performance-sensitive
+// const inlineBuild = esbuild
+//   .build({
+//     ...config,
+//     format: 'iife',
+//     splitting: false,
+//     entryPoints: ['src/ts/ssr-utils/dsd-polyfill.ts'],
+//   })
+//   .catch(() => process.exit(1));
```
Diaan, line 84: can this be removed?

Diaan (approved):

anna-lach (approved):

Issue #686, opened 2023-10-13
Create a PoC for the Lit API


DanielleRameau:
Goal of PoC: can we easily create complex forms?

PR #705, opened 2023-10-30
Adds new `@sl-design-system/form` package

jpzwarte:
- New `FormControlMixin` (the old is still there so we can migratie incrementally)
- New `<sl-error>` and `<sl-hint>` so we stop copy/pasting the markup & styling to all form controls
- Remove hint & validation mixins and move them into `FormControlMixin`
- Move `required` & `disabled` out of the new mixin; the control itself should manage these
- Rename `hint` property to `hintText`
- Update `<sl-text-input>` to use the new package

Fixes part of #686


packages/components/text-input/src/text-input.stories.ts
```diff
@@ -12,13 +12,12 @@ export default {
 };
 
 const sizes: InputSize[] = ['md', 'lg'],
-  labelSizes: LabelSize[] = ['sm', 'md', 'lg'],
-  hintSizes: HintSize[] = ['sm', 'md', 'lg'];
+  labelSizes: LabelSize[] = ['sm', 'md', 'lg'];
 
 export const API: StoryObj = {
```
anna-lach, line 17: Can you please update all stories like `ValidInput`, `All` etc.? :)

packages/components/text-input/src/text-input.ts
```diff
@@ -19,33 +14,29 @@ let nextUniqueId = 0;
 /**
  * Single line text input component.
  *
+ * @csspart wrapper - The input's wrapper
  * @slot prefix - Content shown before the input
  * @slot input - The slot for the input element
  * @slot suffix - Content shown after the input
+ * @slot error-text - The error text to display
+ * @slot hint-text - The hint text to display
```
anna-lach, line 22: There is no `hint-text` slot in the code?

packages/components/text-input/src/text-input.ts
```diff
@@ -19,33 +14,29 @@ let nextUniqueId = 0;
 /**
  * Single line text input component.
  *
+ * @csspart wrapper - The input's wrapper
  * @slot prefix - Content shown before the input
  * @slot input - The slot for the input element
  * @slot suffix - Content shown after the input
+ * @slot error-text - The error text to display
```
anna-lach, line 21: There is no `error-text` slot in the code?

packages/components/text-input/src/text-input.ts
```diff
@@ -19,33 +14,29 @@ let nextUniqueId = 0;
...
-    }
-  };
-  #events = new EventsController(this, {
-    click: this.#onClick
-  });
+  static get scopedElements(): ScopedElementsMap {
+    return {
+      'sl-error': Error,
+      'sl-hint': Hint,
+      'sl-icon': Icon
```
anna-lach, line 30: We should also have added these components to the text-input's `package.json` file as dependencies?

packages/components/text-input/src/text-input.stories.ts
```diff
@@ -440,7 +439,7 @@ export const CustomValidation: StoryObj = {
           margin-bottom: 8px;
         }
       </style>
-      <sl-text-input required="true" .validators=${[validator]} error-size="sm"></sl-text-input>
+      <sl-text-input required .validators=${[validator]} error-size="sm"></sl-text-input>
```
anna-lach, line 442: We don't want to always show that the input is valid (only on demand), eg. in this example we want to show only invalid state, but here when the input is valid, there is a background for `show-validity=valid` set for the wrapper, but it shouldn't be. :)

packages/components/form/src/form-control-mixin.ts
```diff
@@ -0,0 +1,286 @@
...
+     */
+    get labels(): NodeListOf<HTMLLabelElement> | null {
+      if (isNative(this.formControlElement)) {
+        return this.formControlElement.labels;
+      } else {
+        return this.formControlElement.internals.labels as NodeListOf<HTMLLabelElement>;
+      }
+    }
+
+    /** Returns whether the from control is valid or not. */
```
anna-lach, line 119: Small typo `form` :)

packages/components/form/src/form-control-mixin.ts
```diff
@@ -0,0 +1,286 @@
...
+     * NOTE: This method updates the `showValidity` property and therefore should be called from
+     * `willUpdate`, never from `updated` or you will trigger a new lifecycle update.
+     */
+    updateValidity(): void {
+      this.showValidity = this.report ? (this.valid ? 'valid' : 'invalid') : undefined;
+      this.#setErrorText(this.errorText ?? this.validationMessage);
+    }
+
+    /**
+     * Set's a custom validation message for the form control. If the message
```
Diaan, line 211:
```suggestion
     * Sets a custom validation message for the form control. If the message
```

packages/components/text-input/src/text-input.ts
jpzwarte: They are part of the `FormControlMixin`. Should I list them here, or in the mixin?

anna-lach: In the mixin I guess.

Diaan: does it also show up in the documentation page when it's inherited from a mixin?

Diaan: are they really slots though? aren't they properties now? and depending on the value of the property/state you show the elements? because i see `<sl-error>` in the render function here

Diaan (approved):

packages/components/text-input/src/text-input.stories.ts
jpzwarte: Fixed!

PR #715, opened 2023-11-01
Use the new `FormControlMixin` for checkbox components

### LINKED ISSUE #665: Checkbox: clickable area is too big
**Describe the bug**
Width of the clickable area of the checkbox increases to the same width as the hint.

**To Reproduce**
1. Go to https://kind-plant-088739303.1.azurestaticapps.net/?path=/story/checkbox--api&globals=theme:magister
2. Click on the area *after* the `.wrapper` element
3. Notice the checkbox toggles

**Expected behavior**
It should not be clickable outside of the `.wrapper` element

**Screenshots**
CleanShot 2023-09-26 at 10 28 37@2x
CleanShot 2023-09-26 at 10 28 00@2x

jpzwarte:
Fixes #665


anna-lach:
Can you please update Validate in Form story to have it working like it was before?
image

anna-lach:
Are we going to update those stories?
image

If not I guess checkbox and checkbox group should also be added to the new form stories examples?
So somewhere here:
image

anna-lach:
We already have checkbox documentation live on our documentation website so changes there also need to be done I think.
So example here need to be improved (for now it's not clickable)
image

And also here:
image


And it looks like some documentation is missing for `UpdateValidityEvent `
image

packages/components/form/src/form-control-mixin.ts
```diff
@@ -1,7 +1,6 @@
 import type { PropertyValues, ReactiveElement } from 'lit';
 import type { Constructor } from '@sl-design-system/shared';
 import { property, state } from 'lit/decorators.js';
-import styles from './form-control-mixin.scss.js';
 import { UpdateValidityEvent } from './update-validity-event.js';
```
anna-lach, line 4: It looks like there is a description missing for the documentation part of `UpdateValidityEvent`

jpzwarte:
> Can you please update Validate in Form story to have it working like it was before?

Since this functionality is now part of form-field, I've added examples to form-field for this.
CleanShot 2023-12-12 at 13 09 05@2x

packages/components/form/src/form-control-mixin.ts
jpzwarte: I managed to add a description for the event, but the event name remains blank. I think we need a new story to "fix" the cem tool to output event metadata properly.

anna-lach:
It looks like those examples on the website of checkboxes are not clickable?
image
also in the `usage` tab

anna-lach:
Is it possible to add missing description for the `UpdateValidityEvent ` somehow?
image

packages/components/form/src/form-control-mixin.ts
anna-lach: Maybe it's possible to add a name somehow in the comment? :) We have it already live on the website so in my opinion it would be better to have the value somehow added here?

anna-lach: Maybe it's possible to add a name somehow in the comment? :) We have it already live on the website so in my opinion it would be better to have the value somehow added here?

Diaan (approved):

PR #747, opened 2023-11-15
Feature/740 findings in themes button component

Diaan:


jpzwarte (approved):

packages/components/button/src/button.scss
```diff
@@ -59,6 +59,9 @@ $variants: default, primary, success, warning, danger;
     --_font: var(--sl-text-button-text-#{$size});
     --_label-gap: var(--sl-space-button-gap-#{$size});
   }
```
anna-lach, line 61: Solid and outline button in Figma have the same padding, but in our components we have different padding for solid and for outline, which one is the proper one?

image

Diaan: The problem is that all buttons use the "solid" padding in Figma and of course are different in the code. @DanielleRameau I think it's best you fix this so you can decide if it should indeed be the value of the solid button, or whether it should be bigger/smaller for other buttons.

Diaan: This problem will be fixed in a different issue: #749

PR #748, opened 2023-11-15
Feature/590 selectbox all themes

RoaldBoerema:
Added tokens for select box to all themes (except Magister Nova Era en Myvanin)


Diaan (approved):

Diaan (changes requested):
The boxshadow doesn't work anymore.
image
With this change it does. Is this an option according to Figma?

This is an example from sanoma-learning/light.json

Diaan (approved):

PR #805, opened 2023-12-08
Cleanup text-field

jpzwarte:
Since hint & error are now rendered by `<sl-form-field>`, the *only* thing the `<sl-text-field>` now contains is the input. This PR cleans up the component to remove the wrapper element (no longer necessary) and uses `delegatesFocus` for handling the focus.


packages/components/text-field/src/text-field.stories.ts
```diff
@@ -189,62 +186,6 @@ export const Size: StoryObj = {
...
-      <sl-text-field id="input">
-        <div slot="hint-text">
-          Hint is an accessible way to provide <strong>additional information</strong> that might help the user
-        </div>
-      </sl-text-field>
-    </form>
-  `
-};
-
 export const ErrorMessageSizes: StoryObj = {
```
anna-lach, line 189: Can you please update this story?
Error messages are not visible right now. :)

image

packages/components/text-field/src/text-field.stories.ts
```diff
@@ -15,7 +15,6 @@ const sizes: TextFieldSize[] = ['md', 'lg'];
 export const API: StoryObj = {
```
anna-lach, line 15: Can you please also improve min max length story? Validation message is not shown after validation clicked
image

anna-lach: Also Pattern story:

image

anna-lach: And also that story with Custom validation?
image

anna-lach, line 15: If we want to keep those form stories, could you please update text fields there as well? :)

image

Diaan (approved):

packages/components/text-field/src/text-field.stories.ts
jpzwarte: I've updated the stories

jpzwarte: I'll refactor the "Forms" stories later (as I mentioned in the checkbox PR).

PR #807, opened 2023-12-11
Use the new `FormControlMixin` for textarea

jpzwarte:


packages/components/textarea/src/textarea.stories.ts
```diff
@@ -32,20 +29,19 @@ export const API: StoryObj = {
...
         .placeholder=${placeholder}
         .required=${required}
+        .resize=${resize}
         .size=${size}
         .value=${value}
-        .resize=${resize}
-        .hint=${hint}
       ></sl-textarea>
     `
 };
```
anna-lach, line 47: Can you please also update `All` story?
image
There should be an icon when the textarea is invalid and also there are some examples that should have visually shown that the textarea is valid (with green icon)

packages/components/textarea/src/textarea.stories.ts
```diff
@@ -134,68 +130,6 @@ export const Label: StoryObj = {
...
-      <sl-textarea id="textarea">
-        <div slot="hint">
-          Hint is an accessible way to provide <strong>additional information</strong> that might help the user
-        </div>
-      </sl-textarea>
-    </form>
-  `
-};
-
 export const MinMaxLength: StoryObj = {
```
anna-lach, line 133: Can you please update the example?
image

When the textarea is invalid it should have red border and red icon etc.

anna-lach: In this example as well :)

image

packages/components/textarea/src/textarea.ts
```diff
@@ -25,27 +22,35 @@ let nextUniqueId = 0;
  *
  * @slot textarea - The slot for the textarea element
  */
-export class Textarea extends FormControlMixin(HintMixin(LitElement)) {
+export class Textarea extends FormControlMixin(ScopedElementsMixin(LitElement)) {
   /** @private */
-  static override styles: CSSResultGroup = [validationStyles, hintStyles, styles];
+  static get scopedElements(): ScopedElementsMap {
+    return {
+      'sl-icon': Icon
```
anna-lach, line 29: Icon should be also added as dependency in the textarea's package.json file?

packages/components/textarea/src/textarea.scss
```diff
@@ -99,13 +99,24 @@ $resize-types: none, both, horizontal, vertical, auto;
...
   gap: var(--_gap);
+  outline-color: transparent;
+  position: relative;
+  transition: var(--_transition);
+  transition-property: background, border, color, filter, outline-color;
   vertical-align: top;
 }
 
-:host(:disabled) {
+:host(:focus-visible) {
```
anna-lach, line 114: `:focus-within` is missing? There is no outline, when textarea is focused

Diaan (approved):

packages/components/textarea/src/textarea.ts
jpzwarte: Good catch! I've fixed this for text-field as well.

Issue #277, opened 2023-02-13
[V1] Inline Message


DanielleRameau:
## User Story
As a Sanoma Learning Design System user, I want a versatile and user-friendly Inline Message Component that can effectively handle various use cases across our educational products. This Inline Message Component should be adaptable to convey:

## Definition
Inline messages are strategically positioned cues in user interfaces, providing contextual information and notifying users of ongoing actions. Displaying updates or warnings related to system aspects, ensuring users are promptly informed of necessary changes for a seamless experience.

## Key Points
- Appear near their related items.
- Fill the width of the container.
- Does not dismiss automatically.
- Do use to inform the user about ongoing events in flow.
- Do use for confirmation and low-priority messages.

## Acceptance Criteria

**Elements**
- Icon
- Title
- Description
  - Plain text
  - Summary List (Error list)
- Close Button

**Status**
The status defines how the design system visually displays the components inside of the message. **For example, a warning message with an inline link would automatically use the warning link variant.** Not possible to do this within the component, explain in documentation.
- Info (Default)
- Success
- Warning
- Danger

**Sizes** 
For version 1, we only focus on one size. It’s important to remember that the content area must define the width of the Inline Message.
- Medium (Default)

**Position**
The inline message must be positioned close to the workflow area and needs to be visible when it appears. In a form, the message will appear inline at the bottom of the form, close to the submit button. In a table view, the messages will appear at the top of the page to make it more visible the first time.
- Inline Top
- Inline Botton

**A11y**
- Screen readers will announce the text inside the Inline Message without moving the focus to the Inline Message.
- Using an Inline Message for long messages because they aren’t timed, making it easier for people with various disabilities to get the entire message.

**Tokens Notes**
- Left Border - Some products have a much thicker border.

## Figma Links
- [Figma Inline Message Component Research](https://www.figma.com/file/PGb66fVN325tu0IngRka66/Component-Research?type=design&node-id=52-422&mode=design&t=zM963NoDdSV01FB0-4)

## Reference Links
- [How to make inline error messages accessible](https://hidde.blog/how-to-make-inline-error-messages-accessible/), by Hidde de Vries
- [Code Library of Accessibility](https://dequeuniversity.com/library/aria/alert#:~:text=An%20alert%20is%20a%20special,them%20obvious%20when%20they%20appear.)

```[tasklist]
### User stories
- [ ] https://github.com/sl-design-system/components/issues/636
- [ ] https://github.com/sl-design-system/components/issues/717
[...]

PR #823, opened 2023-12-15
Feature/608 inline message component

anna-lach:


packages/components/inline-message/src/inline-message.ts
```diff
@@ -0,0 +1,129 @@
...
+import { Icon } from '@sl-design-system/icon';
+import { Button } from '@sl-design-system/button';
+import { breakpoints } from '@sl-design-system/shared';
+import { localized, msg } from '@lit/localize';
+import { faOctagonExclamation } from '@fortawesome/pro-solid-svg-icons';
+import { LitElement, html, nothing } from 'lit';
+import { property, query } from 'lit/decorators.js';
+import styles from './inline-message.scss.js';
+
+Icon.registerIcon(faOctagonExclamation);
```
jpzwarte, line 14: Should this not be part of the theme?

packages/components/inline-message/src/inline-message.ts
```diff
@@ -0,0 +1,129 @@
...
+    if (this.status === 'danger' || this.status === 'warning') {
+      this.setAttribute('role', 'alert');
+    } else {
+      this.setAttribute('role', 'status');
+    }
+  }
+
+  override render(): TemplateResult {
+    return html`
+      <div class="inline-message-wrapper" open>
```
jpzwarte, line 81: Since we're in the Shadow DOM here, there is no need for the `inline-message` prefix. It's fine to just call this `class="wrapper"`

packages/components/inline-message/src/inline-message.ts
```diff
@@ -0,0 +1,129 @@
...
+        return 'triangle-exclamation-solid';
+      default:
+        return 'info';
+    }
+  }
+
+  override async connectedCallback(): Promise<void> {
+    super.connectedCallback();
+
+    if (this.status === 'danger' || this.status === 'warning') {
```
jpzwarte, line 72: Should this not be in `updated()` instead? What if somebody changes the `status` property after `connectedCallback`?

jpzwarte: You could also 1-line this: `this.setAttribute('role', ['danger', 'warning'].includes(this.status) ? 'alert' : 'status');`

packages/components/inline-message/src/inline-message.ts
```diff
@@ -0,0 +1,129 @@
...
+    } else {
+      this.setAttribute('role', 'status');
+    }
+  }
+
+  override render(): TemplateResult {
+    return html`
+      <div class="inline-message-wrapper" open>
+        <div class="content">
+          ${!this.noIcon
```
jpzwarte, line 83: This could be simpler:

```
${this.noIcon ? nothing : html`....`}
```

packages/components/inline-message/src/inline-message.ts
```diff
@@ -0,0 +1,129 @@
...
+            ? html`<slot name="icon" part="icon">
+                <sl-icon name=${this.iconName} size="md"></sl-icon>
+              </slot>`
+            : nothing}
+          <div class="content-details">
+            <slot name="title" part="title">
```
jpzwarte, line 89: Why have a nested slot without any additional elements? Afaics the title slot doesn't add anything that the default slot already has?

packages/components/inline-message/src/inline-message.ts
```diff
@@ -0,0 +1,129 @@
...
+                  <sl-icon name="xmark"></sl-icon>
+                </sl-button>
+              </slot>
+            `
+          : nothing}
+      </div>
+    `;
+  }
+
+  onClose(): void {
```
jpzwarte, line 109: Please make the event handlers private: `#onClose(): void {`

packages/components/inline-message/src/inline-message.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.wrapper?.removeAttribute('open');
+    this.#closeOnAnimationend();
+  }
+
+  #closeOnAnimationend(): void {
+    console.log('closeOnAnimationend', this.wrapper);
```
jpzwarte, line 115: Remove plz

packages/components/inline-message/src/inline-message.ts
```diff
@@ -0,0 +1,129 @@
...
+      </div>
+    `;
+  }
+
+  onClose(): void {
+    this.wrapper?.removeAttribute('open');
+    this.#closeOnAnimationend();
+  }
+
+  #closeOnAnimationend(): void {
```
jpzwarte, line 114: Is there a reason why we can't have the `animationend` event handler directly defined on the `.wrapper` div in the `render()` method?

jpzwarte:
One last personal comment: I don't feel like the open/close animation adds anything.

jpzwarte (approved):

packages/components/inline-message/src/inline-message.ts
anna-lach: Icon value comes from the token, but I'm not sure if we can use it in a different way than now?
image

jpzwarte: No, what I meant is that icons we use inside the SLDS web components should be defined in the theme. That way the icon is theme specific.

anna-lach: So I guess this is a question to @arecuenco-dsgn and @RoaldBoerema - the icons would be the same for all themes or different per theme?

jpzwarte: Also a question for Diana I think. I think by design all icons used in the web components should be defined in the theme.

packages/components/inline-message/src/inline-message.ts
anna-lach: yeah, not necessary - removed :)

packages/components/inline-message/src/inline-message.ts
anna-lach: Unnecessary icon registration is removed :)

packages/components/inline-message/src/inline-message.scss
```diff
@@ -0,0 +1,150 @@
+$statuses: info, success, warning, danger;
+
+:host {
+  --_background: var(--_background-color);
+  --_border: var(--sl-border-width-inline-message-default) solid var(--_border-color);
+  --_border-left: var(--sl-border-width-inline-message-start) solid var(--_border-color);
+  --_border-radius: var(--sl-border-radius-inline-message-default);
+  --_padding-block: var(--sl-space-inline-message-block);
+  --_padding-inline-start: calc(var(--sl-space-inline-message-inline-start));
+  --_padding-inline-end: var(--sl-space-inline-message-inline-end);
```
Diaan, line 10:
```suggestion
  --_padding-inline: var(--sl-space-inline-message-inline-start) var(--sl-space-inline-message-inline-end);
```
so you online have 1 variable you need to set later on.

packages/components/inline-message/src/inline-message.scss
```diff
@@ -0,0 +1,150 @@
...
+  margin: auto auto 0;
+  outline: none;
+  overflow: hidden;
+  padding-block: var(--_padding-block);
+  padding-inline: var(--_padding-inline-start) var(--_padding-inline-end);
+  position: relative;
+  width: auto;
+}
+
+.wrapper[close]:not([open]) {
```
Diaan, line 54: do you need the `:not` part?

packages/components/inline-message/src/inline-message.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  /** Determines whether closing button (default one) should be shown in the top right corner. */
+  @property({ type: Boolean, reflect: true }) dismissible = true;
+
+  /** Determines whether the icon should be shown on the left side of the component. */
+  @property({ type: Boolean, attribute: 'no-icon' }) noIcon?: boolean;
+
+  /** The status of the inline message.
+   * @type {'info' | 'success' | 'warning' | 'danger'} */
+  @property({ reflect: true }) status: InlineMessageStatus = 'info';
```
Diaan, line 48: I'm not sure "status" is the best term. I'd use "variant" just like we do for the buttons.

packages/components/inline-message/src/inline-message.ts
```diff
@@ -0,0 +1,119 @@
...
+    `;
+  }
+
+  #closeOnAnimationend(event: AnimationEvent): void {
+    if (event.animationName !== 'slide-in-up') {
+      this.wrapper?.removeAttribute('open');
+      this.wrapper?.addEventListener(
+        'animationend',
+        () => {
+          this.wrapper?.removeAttribute('open');
```
Diaan, line 108: Isn't this already removed a couple of lines ago?

packages/components/inline-message/src/inline-message.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+      </sl-inline-message> `);
+      const closeButton = msg.shadowRoot?.querySelector('slot[name="close-button"] sl-button') as Button;
+
+      setTimeout(() => closeButton.click());
+      await new Promise(resolve => setTimeout(resolve, 500));
+
+      expect(document.querySelectorAll('sl-inline-message')).not.to.exist;
+    });
+
+    it('should close the inline message when onClose is called', async () => {
```
Diaan, line 67:
```suggestion
    it('should close the inline message when remove is called', async () => {
```

Diaan (commented):
Is the colour of the hover button correct? 
image

packages/components/inline-message/src/inline-message.ts
anna-lach: I used `status` to have it the same like in figma
image

anna-lach:
> Is the colour of the hover button correct? image

good one, button variant needs to be the same as inline message status. I need to change it :)

packages/components/inline-message/src/inline-message.ts
anna-lach: We also have status in acceptance criteria, so maybe we should leave `status` for now?
https://github.com/sl-design-system/components/issues/277

Diaan (approved):

PR #828, opened 2023-12-20
Refactor `<sl-select>`

jpzwarte:


packages/components/select/package.json
```diff
@@ -38,8 +38,7 @@
     "test": "echo \"Error: run tests from monorepo root.\" && exit 1"
   },
   "dependencies": {
-    "@floating-ui/dom": "^1.5.3",
-    "@sl-design-system/popover": "0.1.0",
+    "@sl-design-system/form": "0.0.2",
     "@sl-design-system/shared": "0.2.0",
```
anna-lach, line 42: Icon dependency should be added here as well?

packages/components/form/src/form-field.stories.ts
```diff
@@ -81,6 +82,18 @@ export const RadioGroup: Story = {
   }
 };
 
+export const Select: Story = {
```
anna-lach, line 85: it looks like that `invalid` state needs to look differently comparing to Figma:
image

And currently in the Storybook
image

packages/components/select/src/select.stories.ts
```diff
@@ -78,24 +146,68 @@ export const Basic: StoryObj = {
...
+  }
 };
 
-const options = html`
-  <sl-select-option>🐷 Pig</sl-select-option>
-  <sl-select-option selected>🐨 Koala</sl-select-option>
-  <sl-select-option>🐼 Panda</sl-select-option>
-  <sl-select-option>🦊 Fox</sl-select-option>
-`;
 export const All: StoryObj = {
```
anna-lach, line 195: Can you please update `All` story to show `valid` and `invalid` state as well (with the right styling)? :)

image

packages/components/select/src/select-option-group.spec.ts
```diff
@@ -6,25 +6,28 @@ import { SelectOptionGroup } from './select-option-group.js';
...
+      </sl-select-option-group>
+    `);
+  });
+
+  it('should render correctly', () => {
+    expect(el).shadowDom.to.equalSnapshot();
+  });
+
+  it('should have a group role', () => {
+    expect(el).to.have.attribute('role');
```
Diaan, line 23: you don't check what the role is

Diaan (approved):

packages/components/form/src/form-field.stories.ts
jpzwarte: Fixed!

anna-lach (approved):

PR #829, opened 2023-12-20
Feature/619 skeleton component

anna-lach:


packages/components/skeleton/src/skeleton.scss
```diff
@@ -1,26 +1,32 @@
...
+  --_min-height: var(--sl-size-skeleton-min-height);
+  --_shine-color: var(--sl-color-skeleton-effect);
   --_animation-duration: var(--sl-skeleton-animation-duration, 3000ms);
 }
 
 :host {
   background: var(--_background);
-  border-radius: var(--_border-radius);
   display: block;
   min-height: var(--_min-height);
```
jpzwarte, line 13: `min-block-size` plz

packages/components/skeleton/src/skeleton.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+  it('should have an aria busy', () => {
+    expect(el).to.have.attribute('aria-busy', 'true');
+  });
+
+  it('should have a shimmer effect by default', () => {
+    expect(el).to.have.attribute('effect', 'shimmer');
+  });
+
+  it('should have a shimmer effect when set', async () => {
+    el.setAttribute('effect', 'shimmer');
```
jpzwarte, line 26: You're testing if the attribute is set, *after you've set the attribute* :)

Better to do `el.effect = 'shimmer';` and then test the value of the attribute :)

packages/components/skeleton/src/skeleton.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    await el.updateComplete;
+
+    const effect = el.getAttribute('effect');
+    expect(effect).to.equal('shimmer');
+  });
+
+  it('should have a pulse effect when set', async () => {
+    el.setAttribute('effect', 'pulse');
```
jpzwarte, line 34: same here

packages/components/skeleton/src/skeleton.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    await el.updateComplete;
+
+    const effect = el.getAttribute('effect');
+    expect(effect).to.equal('pulse');
+  });
+
+  it('should have a sheen effect when set', async () => {
+    el.setAttribute('effect', 'sheen');
```
jpzwarte, line 42: etc.

packages/components/skeleton/src/skeleton.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+    await el.updateComplete;
+
+    const effect = el.getAttribute('effect');
+    expect(effect).to.equal('sheen');
+  });
+
+  it('should have no effect when set to none', async () => {
+    el.setAttribute('effect', 'none');
```
jpzwarte, line 50: etc.

packages/components/skeleton/src/skeleton.stories.ts
```diff
@@ -60,6 +58,7 @@ export const API: Story = {
         .body sl-skeleton {
           block-size: 0.5rem;
           inline-size: 80%;
+          height: 1rem;
```
jpzwarte, line 61: ehh, `block-size` === `height` :)

packages/components/skeleton/src/skeleton.ts
```diff
@@ -5,13 +5,22 @@ import styles from './skeleton.scss.js';
...
 
+export type SkeletonVariant = 'circle' | 'default';
+
 export class Skeleton extends LitElement {
   /** @private */
   static override styles: CSSResultGroup = styles;
 
-  /** Skeleton's effect. */
+  /** Skeleton's effect.
+   *  @type {'none' | 'shimmer' | 'pulse' | 'sheen'}
```
jpzwarte, line 15: I'm always wondering if the cem analyser needs this, or if it can read typescript and figure out the types from there?

jpzwarte (approved):

packages/components/skeleton/src/skeleton.stories.ts
anna-lach: right 😅

packages/components/skeleton/src/skeleton.ts
anna-lach: Maybe we can test this plugin in a separated task?

packages/components/skeleton/src/skeleton.stories.ts
```diff
@@ -80,115 +78,74 @@ export const API: Story = {
...
           flex-direction: column;
           gap: 0.5rem;
         }
+
+        sl-skeleton {
+          height: 1rem;
+        }
       </style>
-      <section>
+      <section aria-label="Loading">
```
Diaan, line 125: I don't think this is really necessary. :D Since there is already some text in this container, technically, it isn't loading :D 
I think the aria-label that has a value with "loading" in it should not have any text at all in it. If you have a mix of loaded and still loading elements you should make a deeper wrapper so that it contains only "loading" elements.

(but of course this is a weird example that no CFA will ever use in this way.)

packages/components/skeleton/src/skeleton.stories.ts
```diff
@@ -214,29 +171,24 @@ export const LoadingImage: Story = {
...
           width: 100%;
           height: 100%;
-          border-radius: 0.5rem;
         }
 
         img {
           width: 100%;
           aspect-ratio: 1 / 1;
-          border-radius: 0.5rem;
+          border-radius: var(--sl-border-radius-skeleton-default); //0.5rem;
```
Diaan, line 179: comment?

Diaan (approved):

packages/components/skeleton/src/skeleton.stories.ts
anna-lach: so I'll just remove `aria-label` from this story :)

packages/components/skeleton/src/skeleton.ts
anna-lach: New task created: https://github.com/sl-design-system/components/issues/832

PR #857, opened 2024-01-02
Development of spinner

Diaan:


packages/components/spinner/src/spinner.scss
```diff
@@ -0,0 +1,49 @@
+@use 'sass:map';
```
anna-lach, line 1: map is not used here? So we don't need it here? :)

packages/components/spinner/src/spinner.stories.ts
```diff
@@ -0,0 +1,103 @@
...
+    variant: {
+      control: 'radio',
+      options: variants
+    }
+  },
+  render: ({ size, variant }) => html` <sl-spinner .size=${size} .variant=${variant}></sl-spinner> `
+} satisfies Meta<Props>;
+
+export const Basic: Story = {};
+
```
anna-lach, line 55: Maybe it would be nice to have a story/example with spinner inside button?

packages/components/spinner/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
```
anna-lach, line 1: spinner should be also added to `tsconfig.all.json` file in `references` array:

`{ "path": "./packages/components/spinner" },`

packages/components/spinner/src/spinner.ts
```diff
@@ -0,0 +1,51 @@
...
+export type SpinnerVariant = 'accent' | 'info' | 'danger' | 'success' | 'warning';
+
+/**
+ * Show totals at a glance or labels contents with a tag.
+ *
+ * ```html
+ * <sl-spinner> 99+ </sl-spinner>
+ * ```
+ *
+ * @slot default - Contents of the spinner
```
anna-lach, line 16: just one small thing - there is no slot, right? :)

packages/components/spinner/src/spinner.ts
```diff
@@ -0,0 +1,51 @@
...
+import styles from './spinner.scss.js';
+
+export type SpinnerSize = 'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl';
+export type SpinnerVariant = 'accent' | 'info' | 'danger' | 'success' | 'warning';
+
+/**
+ * Show totals at a glance or labels contents with a tag.
+ *
+ * ```html
+ * <sl-spinner> 99+ </sl-spinner>
```
anna-lach, line 13: so the example should be without `99+` inside? :)

anna-lach (approved):

PR #861, opened 2024-01-03
Fix builtin vs custom validation behavior for form controls

jpzwarte:


anna-lach:
Maybe when there is a custom error shown the form field should be automatically set to invalid?

image

anna-lach (approved):

Diaan (approved):

PR #871, opened 2024-01-08
Rename `Icon.registerIcon` to `Icon.register`

jpzwarte:


packages/components/icon/src/icon.ts
```diff
@@ -39,40 +43,47 @@ export class Icon extends LitElement {
...
    * Is run in the setup method of each theme.
+   *
+   * @param {IconLibrary} icons The IconLibrary of the theme
    */
-  static registerIcons(icons: IconLibrary): void {
-    window.SLDS.icons = { ...window.SLDS.icons, ...icons };
+  static register(icons: IconLibrary): void;
+
+  static register(icon: IconDefinition | IconLibrary, ...icons: IconDefinition[]): void {
+    // TODO: find a better (and more universal) way to only log these kind of warnings in dev mode
```
jpzwarte, line 59: So lit does this: `if (process.env.NODE_ENV === 'development') { console.warn(....) }` which I've seen elsewhere as well. But I'm not sure about the implications. Lit also *ships* separate development and production bundles. And those are specified in `package.json`. Something to think about.

Diaan (approved):

PR #883, opened 2024-01-16
Typography of avatar

Diaan:


packages/components/avatar/src/avatar.ts
```diff
@@ -86,9 +88,14 @@ export class Avatar extends LitElement {
   /** @private */
```
jpzwarte, line 88: If you're rendering `<sl-tooltip>` inside this component, then this needs to use the `ScopedElementsMixin`, right?

packages/components/avatar/src/avatar.ts
```diff
@@ -161,18 +175,21 @@ export class Avatar extends LitElement {
...
       }
+
+      this.#checkOverflow();
     }
   }
 
   override render(): TemplateResult {
     return html`
       ${this.image ? this.#renderPicture() : nothing}
+      <sl-tooltip id="avatar-tooltip-${this.#avatarId}">${this.profileName}</sl-tooltip>
```
jpzwarte, line 186: Since we're in Shadow DOM, there is no need to use a unique number, right? The id will already be unique *inside the Shadow DOM*?

packages/components/avatar/src/avatar.ts
```diff
@@ -280,6 +297,8 @@ export class Avatar extends LitElement {
     `;
   }
 
+  #hasOverflow = false;
```
jpzwarte, line 300: Please move (private) variables at the top of the class.

packages/components/avatar/src/avatar.ts
```diff
@@ -140,6 +147,13 @@ export class Avatar extends LitElement {
     this.borderWidth = this.#config?.badgeGapWidth * 2;
     this.setAttribute('shape', this.#config.shape);
 
+    this.#resizeObserver = new ResizeObserver(() => {
```
jpzwarte, line 150: This way a new observer will be created every time `connectedCallback` is called. That can be multiple times. It's probably better to create the observer directly when you declare the variable.

packages/components/avatar/src/avatar.ts
```diff
@@ -140,6 +147,13 @@ export class Avatar extends LitElement {
...
+      this.#checkOverflow();
+      return;
+    });
+
+    this.#resizeObserver.observe(this);
```
jpzwarte, line 155: I don't see a `.disconnect()` anywhere? Shouldn't that be done inside the `disconnectedCallback()`?

jpzwarte (approved):

packages/components/avatar/src/avatar.ts
anna-lach: And also `@sl-design-system/tooltip` dependency in the `package.json` needs to be added?

anna-lach (approved):

anna-lach:
Can you please add changeset as well? :)

packages/components/avatar/src/avatar.ts
Diaan: I copied this from the grid, there it is also done on the connected callback, maybe it needs to change there as well?

packages/components/avatar/src/avatar.ts
jpzwarte: Excellent point :)

PR #884, opened 2024-01-16
Feature/629 development of card

Diaan:


packages/components/card/register.ts
```diff
@@ -0,0 +1,9 @@
+import { Card } from './src/card.js';
+
+customElements.define('sl-card', Card);
+
```
anna-lach, line 4: Can you please add changeset? :)

packages/components/card/src/card.scss
```diff
@@ -0,0 +1,198 @@
...
+  --_content-gap: var(--sl-space-card-gap-content-block) var(--sl-space-card-gap-content-inline);
+  --_content-padding-block: 0 var(--sl-space-card-content-block);
+  --_content-padding-inline: var(--sl-space-card-content-inline);
+  --_font-body: var(--sl-text-card-body);
+  --_font-sub-title: var(--sl-text-card-subtitle);
+  --_font-title: var(--sl-text-card-title);
+  --_media-padding-full: 0;
+  --_media-padding: var(-sl-space-card-media-margins);
+
+  --_media-column: var(--_media-width);
```
anna-lach, line 21: Maybe we should have `:host` separated here as well like in other components so `host` with variables and separated one with exact usage? :)

packages/components/card/src/card.ts
```diff
@@ -0,0 +1,88 @@
...
+  @property({ reflect: true }) height: CardHeightOptions = 'fixed';
+  @property({ reflect: true }) orientation: CardOrientation = 'horizontal';
+  @property({ reflect: true }) icon?: string;
+  @property({ reflect: true, attribute: 'media-position' }) mediaPosition: CardMediaPosition = 'start';
+
+  override connectedCallback(): void {
+    super.connectedCallback();
+    this.#setOrientation();
+
+    this.#resizeObserver?.observe(this);
```
anna-lach, line 54: `disconnect()` is missing in the `disconnectedCallback`

website/src/categories/components/card/card.md
```diff
@@ -0,0 +1,16 @@
...
+layout: "components/components.njk"
+tags: component
+packageName: card
+eleventyNavigation:
+  parent: Components
+  key: Card
+  order: 5
+  status: ready
```
anna-lach, line 15: Do we want to keep it as `ready` even if all parts are not ready yet?

packages/components/card/src/card.ts
```diff
@@ -0,0 +1,88 @@
...
+import { LitElement, html, nothing } from 'lit';
+import { property, queryAssignedElements } from 'lit/decorators.js';
+import styles from './card.scss.js';
+
+export type CardHeightOptions = 'fixed' | 'flex';
+export type CardOrientation = 'horizontal' | 'vertical';
+export type CardMediaPosition = 'start' | 'end';
+
+/**
+ * Let the user know you are processing their data or that the (part of the) page is loading.
```
anna-lach, line 11: Description of spinner? :)

DanielleRameau:
I'm not sure about the 'eye' icons, can we replace them with the dots?

DanielleRameau:
Can we show the badges in primary styling instead of neutral? Just to have more colors :D

DanielleRameau:
I like Denmark though, but I think we need to ask chatgtp to create some learning material content for the future. No need to change it now.

anna-lach (approved):

packages/components/card/src/card.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  /** Observe the grid width. */
+  #resizeObserver?: ResizeObserver = new ResizeObserver(() => {
+    this.#setOrientation();
+  });
+
+  /** @private The slotted media. */
+  @queryAssignedElements({ slot: 'media' }) media?: HTMLElement[];
+
+  @property({ type: Boolean, reflect: true }) padding: boolean = false;
```
jpzwarte, line 43: Don't we want to document the various properties?

packages/components/card/src/card.ts
```diff
@@ -0,0 +1,102 @@
...
+  @property({ reflect: true }) height: CardHeightOptions = 'fixed';
+  @property({ reflect: true }) orientation: CardOrientation = 'horizontal';
+  @property({ reflect: true }) icon?: string;
```
jpzwarte, line 46: Do we want this? Or should we have an icon slot instead? The latter is more flexible (but also slightly more work).

packages/components/card/src/card.ts
```diff
@@ -0,0 +1,102 @@
...
+    super.updated(changes);
+
+    if (changes.has('orientation')) {
+      this.#setOrientation();
+    }
+  }
+
+  override render(): TemplateResult {
+    return html`
+      <div class="container">
```
jpzwarte, line 73: Why have a container div when you also have `:host`?

packages/components/card/src/card.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  #setOrientation(): void {
+    const breakpoint = parseInt(window.getComputedStyle(this).getPropertyValue('--sl-card-horizontal-breakpoint')) || 0;
+    const hasMedia = this.media ? this.media.length > 0 : false;
+    this.classList.remove('horizontal', 'has-media');
+    if (!hasMedia) return;
+
+    if (this.orientation === 'horizontal' && (this.getBoundingClientRect().width > breakpoint || breakpoint === 0)) {
+      this.classList.add('horizontal');
```
jpzwarte, line 97: Please prefix any classes with `sl-`

jpzwarte (approved):

packages/components/card/src/card.ts
Diaan: I think i added it when i wanted to use container queries, but that didn't end up working.

jpzwarte: You may have run into a bug/regression in Safari? That was fixed recently. It definitely should work just fine on `:host`.

packages/components/card/src/card.ts
Diaan: What kind of flexibility would you like?

jpzwarte: I other places we use slots. So perhaps do the same here.

packages/components/card/src/card.ts
Diaan: no, container queries only work when the container has an explicit size, the :host couldn't have that, so my solution was to use a container. But with all the different combinations of overflow and flexible height i couldn't get it to work with container queries. Our requirements were the issue, not the support of container queries.

PR #887, opened 2024-01-17
fix closing of dialog when reduced motion is on

Diaan:


packages/components/dialog/src/dialog.scss
```diff
@@ -22,6 +22,9 @@ $md-breakpoint: 600px;
   --_subtitle-padding-bottom: var(--sl-space-dialog-subtitle-padding-bottom);
   --_title-font: var(--sl-text-dialog-heading);
   --_body-font: var(--sl-text-body-lg-normal);
+
+  --_closing-duration: 0ms;
+  --_opening-duration: 0ms;
 }
```
anna-lach, line 29: Can you please also add changeset with patch? :)

anna-lach (approved):

PR #888, opened 2024-01-17
Fix/required optional label mark

jpzwarte:


packages/components/form/src/label.ts
```diff
@@ -138,37 +137,8 @@ export class Label extends LitElement {
...
-      this.required = required && showRequired;
-    } else {
-      this.optional = false;
-      this.required = required;
-    }
+  #update(): void {
+    this.disabled = this.formControl?.disabled ?? false;
+    this.required = this.formControl?.required ?? false;
   }
 }
```
anna-lach, line 144: It looks like those label stories for optional/required needs to be updated? :)

image

anna-lach (approved):

packages/components/form/src/label.ts
jpzwarte: Fixed!

PR #894, opened 2024-01-18
Add support for async validation to form controls

jpzwarte:


packages/components/checkbox/src/checkbox.stories.ts
```diff
@@ -224,7 +224,34 @@ export const CustomValidity: Story = {
...
+          <sl-checkbox value="3">Three</sl-checkbox>
+        </sl-checkbox-group>
+      `;
+    }
+  }
+};
+
+export const CustomAsyncValidity: Story = {
+  args: {
+    hint: 'This story has an async validator. You need to the middle option to make the field valid. It will wait 2 seconds before validating.',
```
anna-lach, line 243:
```suggestion
    hint: 'This story has an async validator. You need to select the middle option to make the field valid. It will wait 2 seconds before validating.',
```

packages/components/form/custom-elements.json
```diff
@@ -453,7 +462,7 @@
                 {
                   "name": "message",
                   "type": {
-                    "text": "string"
+                    "text": "string | Promise<string>"
```
anna-lach, line 465: It looks like in the documentation website it's not showing `Promise<string>` properly, it only shows `Promise`, but maybe it's sth that we still need to check maybe within another issue?
image

packages/components/form/src/form-control-mixin.ts
```diff
@@ -108,9 +114,7 @@ export function FormControlMixin<T extends Constructor<ReactiveElement>>(constru
     /** The name of the form control. */
     @property({ reflect: true }) name?: string;
 
-    /** Whether to show the validity state.
-     * @type {'valid' | 'invalid' | undefined}
```
anna-lach, line 112: The type is still not properly shown, when it comes from external file, so maybe we should keep it for now? Diana was aware about that issue with cem plugin. 
image

anna-lach (approved):

packages/components/form/custom-elements.json
jpzwarte: I believe this may be related to the PR @Diaan created yesterday on `custom-elements.json` in the website not updating.

packages/components/form/src/form-control-mixin.ts
jpzwarte: See other comment: probably related to build issue with website.

PR #895, opened 2024-01-18
Vertical aligned checkbox for selection column

Pauline-Kingma:
Vertical aligned checkbox for selection column

### DIFF
--- a/packages/components/grid/src/grid.scss
+++ b/packages/components/grid/src/grid.scss
@@ -108,3 +108,7 @@ tr:not(:first-of-type) td {
 .selection-count {
   font-weight: normal;
 }
+
+sl-checkbox {
+  align-self: center;
+}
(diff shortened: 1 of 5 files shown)


packages/components/grid/src/grid.scss
```diff
@@ -108,3 +108,7 @@ tr:not(:first-of-type) td {
 .selection-count {
   font-weight: normal;
 }
+
+sl-checkbox {
```
jpzwarte, line 112: Could you make the selector a little more specific? Perhaps add a `select-toggle` class and use that here. Otherwise things may break in the future when other checkboxes are added to the grid.

jpzwarte (changes requested):

packages/components/grid/src/grid.scss
Pauline-Kingma: Yes I will

jpzwarte (approved):

PR #910, opened 2024-01-23
Rework avatar to fix change-in-update warnings

### LINKED ISSUE #859: Bug in Avatar causing `change-in-update` Lit warning
When running unit tests:

```
packages/components/avatar/src/avatar.spec.ts:

 🚧 Browser logs:
      Element sl-avatar scheduled an update (generally because a property was set) after an update completed, causing a new update to be scheduled. This is inefficient and should be avoided unless the next update can only be scheduled as a side effect of the previous update. See https://lit.dev/msg/change-in-update for more information.
```

jpzwarte:
Fixes #859

### DIFF
--- a/packages/components/avatar/src/avatar.ts
+++ b/packages/components/avatar/src/avatar.ts
@@ -15,10 +15,58 @@ import { LitElement, html, nothing, svg } from 'lit';
 import { property, state } from 'lit/decorators.js';
 import styles from './avatar.scss.js';
 
-let nextUniqueId = 0;
+const BORDER_WIDTH = 4; //has to be double the desired "gap"; the stroke is centered on the path, so only half of is it outside the badge rect.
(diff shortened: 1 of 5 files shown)


Diaan (approved):

packages/components/avatar/src/avatar.ts
```diff
@@ -194,12 +179,23 @@ export class Avatar extends LitElement {
   }
 
   #renderPicture(): TemplateResult {
-    const { x, y, containerSize, radius, size } = this.image!;
+    const { x, y, containerSize, radius, size } = this.image!,
+      parts: string[] = [];
```
Diaan, line 183: `parts` isn't a very clear name (i assume it was called this in a function/context that made the label, so it could be inferred from the context that it were label-parts, but that context is gone here.)

Issue #527, opened 2023-07-17
[V1] Tabs


DanielleRameau:
## User Story
As a Sanoma Learning Design System user, I want a versatile and user-friendly Tab Component that can effectively handle various use cases across our educational products. This Tab Component should be adaptable to convey:

## Definition
Tabs are used to organize related content. They allow the user to navigate between groups of information that appear within the same context.

## Key Points
- Clear Indication of Active Tab.
- Ensure that your tabs adapt well to different screen sizes.
- Maintain a standardised placement for tabs.
- Flexibility for Content Variation.
- Make sure that tab navigation is keyboard-friendly.

## Acceptance Criteria

**Elements**
| Name  | Optional |
| ------------- | ------------- |
| Label | No |
| Subtitle | Yes |
| Icon | Yes |
| Badge | Yes |

**States**
The status defines how the design system visually displays the components inside of the message. 
- Default
- Hover
- Active/Pressed
- Disabled

**Sizes** 
- Medium (Default)

**Layout**
- Horizontal (Default)
- Vertical

**Alignment**
- Left

**Width (tabs)**
- Hug Content 
- Fill Container

**Behaviour** 
- Sticky Tabs Header
- Overflowing tabs: show more tab + dropdown, (show all tabs in dropdown) + slide
Example: https://ant.design/components/tabs

**A11y**
- Navigation through content with the tab key.
- Move through the tabs with the arrow keys.
- Show up "Contextual Menu" (···) when the tabs overflow the width of the container.

## Figma Links
- [Figma Tabs Component Research](https://www.figma.com/file/PGb66fVN325tu0IngRka66/Component-Research?type=design&node-id=270-1340&mode=design&t=eQ4xHOVJ9nGgAcYR-4)

```[tasklist]
### User stories
- [ ] #139
- [ ] https://github.com/sl-design-system/components/issues/728
- [ ] https://github.com/sl-design-system/components/issues/729
- [ ] https://github.com/sl-design-system/components/issues/730
- [ ] https://github.com/sl-design-system/components/issues/731
- [ ] https://github.com/sl-design-system/components/issues/931
```

anna-lach:
@arecuenco-dsgn it looks like this point from **a11y** part is not applicable:
`Shows up arrow tabs when the tabs are overflow the width of the container.`. We've chosen the version with dropdown showing all tabs.

arecuenco-dsgn:
@anna-lach, I update it; what do you think? 
[...]

PR #921, opened 2024-02-01
Feature/730 tabs component

anna-lach:
Tabs component


packages/components/tabs/src/tab-group.scss
```diff
@@ -1,110 +1,259 @@
...
+    width: 100%;
+  }
+}
+
+.indicator,
+.indicator-listbox {
+  background-color: var(--_tab-group-indicator-color);
   position: absolute;
-  z-index: 1;
+  z-index: 2;
```
jpzwarte, line 189: Perhaps we should add comments every time we have a `z-index` > 1?

packages/components/tabs/src/tab-group.scss
```diff
@@ -1,110 +1,259 @@
...
 }
 
 @media (prefers-reduced-motion: no-preference) {
   .indicator {
-    transition: transform 0.5s ease-in-out;
+    transition: transform 500ms ease-in-out;
```
jpzwarte, line 194: Should the animation duration not be a token?

packages/components/tabs/src/tab-group.scss
```diff
@@ -1,110 +1,259 @@
...
+  display: none;
+  flex-direction: column;
+  gap: var(--_listbox-gap);
+  margin: 0;
+  overflow-x: hidden; //because of firefox issue: https://bugzilla.mozilla.org/show_bug.cgi?id=764076
+  overflow-y: auto;
+  padding: var(--_listbox-padding);
+  transition: opacity 0.2s cubic-bezier(0.25, 0, 0.3, 1);
+
+  sl-tab {
```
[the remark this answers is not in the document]
jpzwarte: `:host-context` is Chrome only :(

packages/components/tabs/src/tab-group.spec.ts
```diff
@@ -11,10 +11,18 @@ describe('sl-tab-group', () => {
...
       el = await fixture(html`<sl-tab-group></sl-tab-group>`);
     });
-  
+
     it('should not break', () => {
       expect(el).shadowDom.to.equalSnapshot();
     });
+
+    it('should have horizontal layout by default', () => {
+      expect(el.getAttribute('vertical')).to.equal('false');
```
jpzwarte, line 20: `expect(el).not.to.have.attribute('vertical');` (as a result of the earlier `@property` comment)

packages/components/tabs/src/tab-group.spec.ts
```diff
@@ -11,10 +11,18 @@ describe('sl-tab-group', () => {
...
+    });
+
+    it('should have left alignment by default', () => {
+      expect(el.getAttribute('alignment')).to.equal('left');
```
jpzwarte, line 24: `expect(el).to.have.attribute('alignment', 'left');`

Also, perhaps we should use logical values? `start` instead of `left`?

packages/components/tabs/src/tab-group.spec.ts
```diff
@@ -153,33 +167,128 @@ describe('sl-tab-group', () => {
...
+
+    const showListbox = async () => {
+      await tabGroup.updateComplete;
+      return new Promise(resolve => setTimeout(resolve, 700));
+    }
+
+    describe('with more button in a small container',() => {
+      beforeEach(async () => {
+        element = await fixture(html`
+          <div style="width: 70px">
```
jpzwarte, line 215: Why wrap this in a div? Setting a width on the tab-group should work as well?

packages/components/tabs/src/tab-group.stories.ts
```diff
@@ -107,12 +133,108 @@ export const LongTitles: StoryObj = {
   </sl-tab-group>`
 };
 
+export const Vertical: StoryObj = {
+  render: ({ alignment }) => html` <style>
+      sl-tab-group::part(container) {
+        position: sticky;
+        inset-block-start: 16px;
+        align-self: flex-start;
```
jpzwarte, line 141: I think `flex-start` is the old syntax. Just `start` is the new one.

packages/components/tabs/src/tab-group.ts
```diff
@@ -43,48 +67,156 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
...
-        <span class="indicator" role="presentation"></span>
-        <slot name="tabs" @slotchange=${() => this.#rovingTabindexController.clearElementCache()}></slot>
+      <div class="container" part="container" @keydown=${this.#handleKeydown}>
+        <div class="wrapper" part="wrapper">
+          <div @click=${this.#handleTabChange} role="tablist" part="tab-list">
+            <span class="indicator" role="presentation"></span>
+            <slot name="tabs" @slotchange=${() => this.#rovingTabindexController.clearElementCache()}></slot>
+            <div
+              id="tabs-popover"
+              ${anchor({ element: this.#moreButton, position: this.vertical ? 'bottom-start' : 'bottom-end' })}
```
jpzwarte, line 122: Since you use the `anchor` directive here, do we still need to set `position: fixed` in the CSS if popover isn't supported?

packages/components/tabs/src/tab-group.ts
```diff
@@ -43,48 +67,156 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
...
+    this.#sizeObserver.disconnect();
+
+    super.disconnectedCallback();
+  }
+
   override firstUpdated(): void {
     this.#observer = new MutationObserver(this.#handleMutation);
     this.#observer?.observe(this, TabGroup.#observerOptions);
-    setTimeout(() => this.#updateSelectionIndicator(), 100);
+    setTimeout(() => this.#updateSelectionIndicator(), 700);
```
jpzwarte, line 171: Please add a comment why 700 specifically.

packages/components/tabs/src/tab-group.ts
```diff
@@ -43,48 +67,156 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
...
+  }
+
+  override updated(changes: PropertyValues<this>): void {
+    super.updated(changes);
+
+    if (changes.has('alignment') || changes.has('vertical')) {
+      requestAnimationFrame(() => {
```
jpzwarte, line 178: Shouldn't we do this *inside* the `#updateSelectionIndicator()` method?

packages/components/tabs/src/tab-group.ts
```diff
@@ -43,48 +67,156 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
...
+      if (!this.listbox || !this.selectedTab) {
+        return;
+      }
+      this.selectedTabInListbox = this.listbox.querySelector(`#${this.selectedTab.id}`) as Tab;
+      indicatorListbox.style.transform = `translateY(${this.selectedTabInListbox?.offsetTop}px)`;
+      indicatorListbox.style.height = `${this.selectedTabInListbox.offsetHeight}px`;
+    });
+
+    if (
+      ((event as ToggleEvent).newState === 'closed' && (event.target as HTMLElement).matches(':popover-open')) ||
```
jpzwarte, line 208: `matches(':popover-open')` only works for non-polyfilled browsers. You need to include the `\:popover-open` selector so it works in older ones as well (please check the polyfill to be sure).

packages/components/tabs/src/tab-group.ts
```diff
@@ -43,48 +67,156 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
...
+      (event as ToggleEvent).newState === (event as ToggleEvent).oldState
+    ) {
+      event.stopPropagation();
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      (event.target as HTMLElement)?.hidePopover();
```
jpzwarte, line 214: CleanShot 2024-02-07 at 13 17 55@2x

`hidePopover` should be on `HTMLElement`?

Also, it's better to do `event: Event & { target: HTMLElement }` in the method definition.

packages/components/tabs/src/tab-group.ts
```diff
@@ -134,14 +266,17 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
   }
 
   #handleTabChange(event: Event): void {
```
jpzwarte, line 268: Please cast `event.target` here once, instead of below twice :)

packages/components/tabs/src/tab-group.ts
```diff
@@ -43,48 +67,156 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
...
+      ((event as ToggleEvent).newState === 'closed' && (event.target as HTMLElement).matches(':popover-open')) ||
+      (event as ToggleEvent).newState === (event as ToggleEvent).oldState
+    ) {
+      event.stopPropagation();
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      (event.target as HTMLElement)?.hidePopover();
+    }
+
+    this.#moreButton?.setAttribute('aria-expanded', `${isPopoverOpen(this.listbox)}`);
```
jpzwarte, line 217: you probably want to do `.toString()` instead of wrapping the method call in a template string?

packages/components/tabs/src/tab-group.ts
```diff
@@ -193,48 +330,81 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
...
+        indicatorListbox = this.shadowRoot?.querySelector('.indicator-listbox') as HTMLElement,
+        wrapper = this.shadowRoot?.querySelector('.container') as HTMLElement,
+        tabsWrapper = this.shadowRoot?.querySelector('.wrapper') as HTMLElement,
+        tablist = this.shadowRoot?.querySelector('[role="tablist"]') as HTMLElement,
+        tabs = this.querySelectorAll('sl-tab');
+
+      let totalTabsWidth = 0;
+      let totalTabsHeight = 0;
+      tabs.forEach(tab => {
+        totalTabsWidth = totalTabsWidth + tab.offsetWidth;
```
jpzwarte, line 366: `+=` is shorter.

packages/components/tabs/src/tab-panel.spec.ts
```diff
@@ -15,7 +15,6 @@ describe('sl-tab-panel', () => {
   });
 
   it('should have the correct attributes', () => {
-    expect(el).attribute('tabindex').to.equal('0');
     expect(el).attribute('role').to.equal('tabpanel');
```
jpzwarte, line 18: `expect(el).to.have.attribute('role', 'tabpanel');`

packages/components/tabs/src/tab.spec.ts
```diff
@@ -20,7 +20,6 @@ describe('sl-tab', () => {
   });
 
   it('should have the correct attributes', () => {
-    expect(el).attribute('slot').to.equal('tabs');
     expect(el).attribute('role').to.equal('tab');
```
jpzwarte, line 23: Use `.to.have.attribute(....)` syntax plz

jpzwarte (changes requested):

packages/components/tabs/src/tab-group.scss
```diff
@@ -1,110 +1,259 @@
...
+  border-bottom: var(--_border-bottom);
   display: flex;
   position: relative;
 }
 
+::slotted(sl-tab-panel) {
+  padding: var(--_panel-horizontal-padding);
+}
+
+sl-button {
```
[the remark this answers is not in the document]
anna-lach: The button styling should be the same as tab styling, that's why I used those here. :)

packages/components/tabs/src/tab-group.scss
```diff
@@ -1,110 +1,259 @@
...
-  gap: 8px;
+  height: inherit;
+  width: inherit;
+}
+
+.container {
+  display: grid;
+  grid-template-columns: 1fr auto;
+  inset-block-start: calc(-1 * var(--_wrapper-spacing));
+  position: sticky;
```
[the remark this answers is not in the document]
anna-lach: We have such acceptance criteria https://github.com/sl-design-system/components/issues/527 that the header tabs should be sticky
image

But maybe by default it should not be sticky and we should leave it up to the developer? And just add a story with sticky tabs (styled this way only as an example)?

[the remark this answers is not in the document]
jpzwarte: I think we should have an example for developers how to make tabs sticky. But I don't think they should be sticky *by default*.

packages/components/tabs/src/tab-group.spec.ts
anna-lach: I used the same alignment as in Figma

image

If we want to change this one I think it also needs to be changed in Figma? @RoaldBoerema what do you think about it? :)

packages/components/tabs/src/tab-group.ts
anna-lach: Without  `position: fixed;` it's not working properly in Firefox:

image

packages/components/tabs/src/tab-group.spec.ts
anna-lach: Changed (in Figma as well) :)

packages/components/tabs/src/tab-group.scss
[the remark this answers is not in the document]
Diaan: But still, with a class you can achieve it as well, and you separate the tab and tab-group styling better.

packages/components/tabs/src/tab-group.scss
```diff
@@ -176,26 +164,22 @@ sl-button {
...
   }
 }
 
-.indicator,
-.indicator-listbox {
-  background-color: var(--_tab-group-indicator-color);
+.indicator {
+  background-color: var(--_indicator-color);
   position: absolute;
+  /** z-index 2 to make it visible, when the tab is selected, focused and hovered */
```
Diaan, line 190:
```suggestion
  /** z-index is set to 2 to make sure the indicator is above the tab, when that is selected, focused or hovered */
```

jpzwarte (approved):

Diaan (approved):

PR #937, opened 2024-02-08
Various dialog fixes

### LINKED ISSUE #934: Dialog body text is larger than buttons
CleanShot 2024-02-07 at 10 46 58@2x

The body text has a font-size of 16px, while the buttons are 14px. I would expect the body text to have the same font-size as the buttons?

jpzwarte:
Fixes #934 and #935


packages/components/dialog/src/dialog.scss
```diff
@@ -21,7 +21,7 @@ $md-breakpoint: 600px;
   --_subtitle-font: var(--sl-text-dialog-subheading);
   --_subtitle-padding-bottom: var(--sl-space-dialog-subtitle-padding-bottom);
   --_title-font: var(--sl-text-dialog-heading);
-  --_body-font: var(--sl-text-body-lg-normal);
+  --_body-font: var(--sl-text-body-md-normal);
```
anna-lach, line 24: This one was made like in Figma before, maybe it should be changed in Figma as well to have it consistent?

image

packages/components/dialog/src/dialog.scss
```diff
@@ -126,18 +109,18 @@ slot[name='footer'] {
...
-    animation: backdrop-hide var(--_closing-duration) ease-out;
+    animation: backdrop-hide var(--_closing-duration) both ease-out;
   }
 }
 
 dialog[open]:not([closing]) {
-  animation: slide-in-up var(--_opening-duration) ease-out;
+  animation: slide-in-up var(--_opening-duration) both ease-out;
 
   &::backdrop {
```
anna-lach, line 122: it looks like the backdrop is not animating when closing the dialog with `Escape` key?

packages/components/dialog/src/dialog.scss
```diff
@@ -126,18 +109,18 @@ slot[name='footer'] {
 }
 
 dialog[closing] {
-  animation: slide-down var(--_closing-duration) ease-out;
+  animation: slide-down var(--_closing-duration) both ease-out;
 
   &::backdrop {
```
anna-lach, line 114: It looks like the backdrop is not animated in Firefox?

packages/components/dialog/src/dialog.scss
jpzwarte: @DanielleRameau See Anna comment above; can you update Figma?

packages/components/dialog/src/dialog.scss
jpzwarte: Good catch, fixed!

packages/components/dialog/src/dialog.scss
jpzwarte: As discussed in Slack, I see no way to fix this in FF. I can't even see the `::backdrop` pseudo element in the FF devtools.

packages/components/dialog/src/dialog.stories.ts
```diff
@@ -1,129 +1,121 @@
...
-  render: () => html`
-    <sl-button fill="outline" @click=${onClick}>Show Dialog</sl-button>
-    <sl-dialog disable-cancel close-button align="space-between">
-      <span slot="title">Disable cancel</span>
-      <p>You cannot close me by pressing the Escape key, or clicking the backdrop.</p>
-      <sl-button slot="actions" fill="solid" variant="default" sl-dialog-close autofocus>Action 2</sl-button>
-      <sl-button slot="actions" fill="solid" variant="primary" sl-dialog-close>Action</sl-button>
-    </sl-dialog>
-  `
+export const HeaderButtons: Story = {
```
anna-lach, line 93: Can you please also show here that the close button can have the same styling as the default, but a different icon inside? As far as I remember that was the case, that it can have a different icon in the closing button :) and other ghost icon buttons next to this one.

image

packages/components/dialog/src/dialog.stories.ts
```diff
@@ -1,129 +1,121 @@
...
+    closeButton: false,
+    disableCancel: true,
+    maxWidth: '300px',
+    title: 'Disable cancel and close button'
+  }
+};
 
-        sl-button-bar[reverse] > sl-button:first-child {
-          margin-left: inherit;
+export const FooterButtons: Story = {
```
anna-lach, line 72: Can you please show an option for the `reverse` to footer buttons like it was before?

image

packages/components/dialog/src/dialog.stories.ts
```diff
@@ -1,129 +1,121 @@
...
-  render: ({ title, subtitle, bodyContent, closeButton }) => {
+  render: ({ body, closeButton, disableCancel, footerButtons, headerButtons, title, subtitle, maxWidth }) => {
+    const onClick = (event: Event & { target: HTMLElement }): void => {
+      (event.target.nextElementSibling as Dialog).showModal();
+    };
+
     return html`
+      <style>
+        sl-dialog {
+          ${maxWidth ? `--sl-dialog-max-inline-size: ${maxWidth}` : ''};
```
anna-lach, line 35: Can you please add sth like  `@cssproperty --sl-dialog-max-inline-size: ...` to dialog.ts file? Just to document it and to make it visible on the website in the future as well?

jpzwarte: Fixed

packages/components/dialog/src/dialog.stories.ts
jpzwarte: Added

packages/components/dialog/src/dialog.stories.ts
jpzwarte: Fixed

PR #960, opened 2024-02-14
Docs/602 tooltip documentation

DanielleRameau:


DanielleRameau:
code en a11ty tab need to be filled and usage tab needs a live demo :)

packages/components/tooltip/src/tooltip.ts
```diff
@@ -62,7 +65,9 @@ export class Tooltip extends LitElement {
     }
   };
 
-  /** Tooltip position. */
+  /** Tooltip position.
+   * @type {'top' | 'right' | 'bottom' | 'left' | 'top-start' | 'top-end' | 'right-start' | 'right-end' | 'bottom-start' | 'bottom-end' | 'left-start' | 'left-end'}
```
jpzwarte, line 69: I thought we no longer needed to do this? cem will automatically extract this info?

website/src/categories/components/tooltip/code.md
```diff
@@ -1,35 +1,96 @@
...
+
+</div>
+
+The complete usage example might appear as follows:
+
+
+<div class="ds-code">
+
+  ```html
+    <sl-button ${tooltip('This tooltip is from a directive')} fill="outline">I have a tooltip</sl-button>
```
jpzwarte, line 89: The `fill` doesn't add anything. Probably best to remove it.

website/src/categories/components/tooltip/tooltip.md
```diff
@@ -1,15 +1,15 @@
 ---
 title: Tooltip
-description: Tooltip description
+description: A tooltip is a small, contextual modal overlay that provides additional information when users interact with an element in a user interface.
 componentType: overlay
-shortDescription: We are still working on the tooltip component
+shortDescription: A tooltip is a small, contextual modal overlay that provides additional information.
```
jpzwarte, line 5: It's not modal :) It would be modal if it had a backdrop.

website/src/categories/components/tooltip/usage.md
```diff
@@ -5,27 +5,82 @@ eleventyNavigation:
...
+
+<sl-button aria-describedby="tooltip" fill="solid" variant="primary">Button with a tooltip</sl-button>
+<sl-tooltip id="tooltip">This is a tooltip with some additional information</sl-tooltip>
+
+</div>
+
+<div class="ds-code">
+
+  ```html
+<sl-button aria-describedby="tooltip" fill="solid" variant="primary">Button with a tooltip</sl-button>
```
jpzwarte, line 19: Again, probably best to only focus on the tooltip. So remove `fill` and `variant`.

website/src/ts/scripts/slds-components.ts
```diff
@@ -1,5 +1,6 @@
 import '@oddbird/popover-polyfill';
 import '@webcomponents/scoped-custom-element-registry/scoped-custom-element-registry.min.js';
+import '@sl-design-system/shared';
```
jpzwarte, line 3: Does this import make sense? Why do we need it?

website/src/ts/scripts/slds-components.ts
```diff
@@ -19,6 +20,7 @@ import '@sl-design-system/tabs/register.js';
 import '@sl-design-system/text-field/register.js';
 import '@sl-design-system/textarea/register.js';
 import '@sl-design-system/tooltip/register.js';
+import '@sl-design-system/tooltip';
```
jpzwarte, line 23: Same question here. This does nothing?

jpzwarte (approved):

website/src/ts/scripts/slds-components.ts
anna-lach: No, I forgot to remove it :)

packages/components/tooltip/src/tooltip.ts
anna-lach: Sometimes it's working, sometimes not, that's why I added it here, because it was not working in this case

PR #965, opened 2024-02-16
Add message dialog utility

jpzwarte:


packages/components/message-dialog/src/message-dialog.scss
```diff
@@ -0,0 +1,7 @@
+sl-dialog {
+  --sl-dialog-max-inline-size: 20rem;
+}
+
+p {
+  margin-block: 0;
```
anna-lach, line 6: I think we need to have for the content a variable to set the font size like `--sl-dialog-body-font-size` or sth to make it possible to override by the developer.

packages/components/message-dialog/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "register.js"
+  ],
+  "scripts": {
+    "test": "echo \"Error: run tests from monorepo root.\" && exit 1"
+  },
+  "dependencies": {
+    "@lit/localize": "^0.12.1",
+    "@open-wc/scoped-elements": "^3.0.5",
+    "@sl-design-system/button": "0.0.18",
+    "@sl-design-system/button-bar": "0.0.4",
```
anna-lach, line 44: Do we need the button bar dependency here?

website/src/categories/components/message-dialog/code.md
```diff
@@ -0,0 +1,119 @@
...
+- `message` (string): The message of the dialog.
+- `buttons` (array): An array of button objects. Each button object can have the following properties:
+  - `text` (string): The text of the button.
+  - `value` (string): The value of the button. This is what will be returned when the button is clicked.
+  - `variant` (string): The variant of the button. This can be "primary", "success", "info", "warning", or "danger".
+- `disableCancel` (boolean): Determines whether the user can cancel the message dialog using the Escape key or by clicking on the backdrop.
+
+The `MessageDialog.show()` API returns a promise that resolves with the value of the button that was clicked. This allows you to perform different actions based on the button that was clicked.
+
+</section>
```
anna-lach, line 119: It looks like the API table is missing (like we have in other places to keep it consistent):

`{% include "../component-table.njk" %}`

jpzwarte: Yeah, I had that intially, but I removed it. This is not your regular web component API. This is kind of a Javascript-only API.

packages/components/message-dialog/src/message-dialog.scss
jpzwarte: In this specific case, I just want to make sure the paragraph doesn't have any margins. If I remove this rule, then the chances that outside styling "breaks" the message dialog is big. Also the more I think about it, the more I think the dialog should just use the `<body>` font.

anna-lach: Ok, so let's add `font-size: 1.4rem` to the `body` in `base.sccs` file:

image

PR #983, opened 2024-02-23
feature/844 inline message title ↔️ description

Diaan:


packages/components/inline-message/src/inline-message.ts
```diff
@@ -81,8 +81,8 @@ export class InlineMessage extends ScopedElementsMixin(LitElement) {
                 <sl-icon name=${this.iconName} size="md"></sl-icon>
               </slot>`}
           <div class="content-details">
-            <slot></slot>
-            <slot name="description" part="description"></slot>
+            <slot name="title"></slot>
+            <slot part="description"></slot>
```
jpzwarte, line 85: Having a slot be a part can be confusing, unless the is something other than the default `display: contents`.

jpzwarte (approved):

PR #994, opened 2024-03-01
950 - Tabs Basic Documentation

DanielleRameau:


website/src/_data/tokens/core.json
```diff
@@ -494,6 +494,10 @@
         "value": "{size.4xl}",
         "type": "sizing"
       }
+    },
+    "breacrumb": {
```
jpzwarte, line 498: typo; but that's from upstream I think (also noticed it in another PR)

DanielleRameau:
@anna-lach When I press on 'tabs' in the navigation the page opens at the point of the live example and not the hero

PR #995, opened 2024-03-04
Improve eslint setup

jpzwarte:
- Use the new flat config eslint setup (https://eslint.org/docs/latest/use/configure/configuration-files-new)
- Add a new `@sl-design-system/eslint-config` package so others can reuse out eslint configuration
- Add specific plugins for a11y, web components, testing, storybook etc.
- Improve performance of listing components by running eslint once for all components instead of separately for each component


packages/components/message-dialog/src/message-dialog.spec.ts
```diff
@@ -1,10 +1,10 @@
...
 import { type Dialog } from '@sl-design-system/dialog';
 import { sendKeys } from '@web/test-runner-commands';
-import { MessageDialog } from './message-dialog.js';
+import { html } from 'lit';
 import { spy } from 'sinon';
 import '../register.js';
-import { html } from 'lit';
+import { MessageDialog } from './message-dialog.js';
 
 describe('sl-message-dialog', () => {
```
anna-lach, line 9: It looks like few tests are not working properly

image

jpzwarte: Fixed

anna-lach (approved):

PR #1026, opened 2024-03-06
Add new package for stylelint config

jpzwarte:


packages/components/text-field/src/text-field.ts
```diff
@@ -15,6 +15,7 @@ let nextUniqueId = 0;
  * Single line text field component.
  *
  * @csspart wrapper - The input's wrapper
+ * @cssprop --sl-textarea-rows - The number of rows initially visible in the textarea
```
anna-lach, line 18: This one should be added to the `textarea.ts` file, right? :)

jpzwarte: Right :)

anna-lach (approved):

PR #1030, opened 2024-03-07
626 - Spinner Basic Documentation

arecuenco-dsgn:
**Status:**
- The usage page is complete and needs review.
- The A11y page has some key points, and a WAI-ARIA table needs to be completed by a developer.


DanielleRameau (commented):
some minor change requests

packages/components/spinner/src/spinner.ts
```diff
@@ -34,7 +34,7 @@ export class Spinner extends LitElement {
     super.connectedCallback();
 
     if (!this.hasAttribute('role')) {
-      this.setAttribute('role', 'alert');
+      this.setAttribute('role', 'status');
```
jpzwarte, line 37: Hmm, should we set a role and nothing else? Or should this be invisible to ARIA? (`role=presentation` or `aria-hidden="true"`)

jpzwarte (approved):

anna-lach (approved):

PR #1036, opened 2024-03-11
Feature/907 accordion component

anna-lach:


packages/components/accordion/src/accordion-item.ts
```diff
@@ -0,0 +1,119 @@
...
+   */
+  #accordionItemId = `sl-accordion-item-${nextUniqueId++}`;
+
+  /**
+   * Unique ID for each accordion item component content.
+   */
+  #accordionItemContentId = `sl-accordion-item-content-${nextUniqueId++}`;
+
+  /** A text shown in the header - as a title of the accordion item. */
+  @property() summary!: string;
```
jpzwarte, line 32: Why the exclamation mark here? In other places we just do `summary?: string`

packages/components/accordion/src/accordion-item.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  /** Whether the element is disabled. */
+  @property({ type: Boolean, reflect: true }) disabled?: boolean;
+
+  /** Emits when the accordion item has been toggled. */
+  @event({ name: 'sl-toggle' }) toggleEvent!: EventEmitter<string>;
+
+  /** Whether the details element is opened. */
+  @property({ reflect: true, type: Boolean }) open?: boolean;
```
jpzwarte, line 41: Nitpicking, but I try to keep the variables alphabetically sorted. So `toggleEvent` after `open` instead of before

packages/components/accordion/src/accordion-item.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+@localized()
+export class AccordionItem extends LitElement {
+  /** @private */
+  static override styles: CSSResultGroup = styles;
+
+  /**
+   * Unique ID for each accordion item component.
+   */
+  #accordionItemId = `sl-accordion-item-${nextUniqueId++}`;
```
jpzwarte, line 24: why the `accordionItem` prefix?

packages/components/accordion/src/accordion-item.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Unique ID for each accordion item component content.
+   */
+  #accordionItemContentId = `sl-accordion-item-content-${nextUniqueId++}`;
```
jpzwarte, line 29: Same

packages/components/accordion/src/accordion-item.ts
```diff
@@ -0,0 +1,119 @@
...
+  @property({ type: Boolean, reflect: true }) disabled?: boolean;
+
+  /** Emits when the accordion item has been toggled. */
+  @event({ name: 'sl-toggle' }) toggleEvent!: EventEmitter<string>;
+
+  /** Whether the details element is opened. */
+  @property({ reflect: true, type: Boolean }) open?: boolean;
+
+  /** @private */
+  @query('.wrapper') wrapper!: HTMLDivElement;
```
jpzwarte, line 44: Since we only use this variable in the `#onClick` method, it might be better to just do `const wrapper = this.renderRoot.querySelector('.wrapper')` at the top of the `#onClick` method?

packages/components/accordion/src/accordion-item.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  override render(): TemplateResult {
+    return html`
+      <details @toggle=${this.#onToggle} ?open=${this.open}>
+        <summary
+          id=${this.#accordionItemId}
```
jpzwarte, line 50: Do we actually need unique ids *inside the shadow DOM*? iirc in other places, I don't do that. In `dialog.ts` for example: `<slot name="title" id="title"></slot>`

packages/components/accordion/src/accordion.ts
```diff
@@ -0,0 +1,63 @@
...
+      this.items.forEach(item => {
+        if (item.renderRoot.querySelector('details')?.hasAttribute('open')) {
+          item.renderRoot.querySelector('summary')?.click();
+        }
+      });
+    }
+  }
+
+  override render(): TemplateResult {
+    return html` <slot @click=${this.#onClick}></slot> `;
```
jpzwarte, line 46: Unnecessary whitespace

packages/components/accordion/src/accordion.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  #onClick(event: Event): void {
+    if (!this.single || event.defaultPrevented) {
+      // No toggling when no `single` or the user prevents it.
+      return;
+    }
+
+    this.items.forEach(item => {
+      if (item !== event.target) {
```
jpzwarte, line 56: This could probably be a `.filter(item => item !== event.target)` instead of an if statement?

jpzwarte (approved):

packages/components/accordion/src/accordion.ts
```diff
@@ -0,0 +1,63 @@
...
+ *   <sl-accordionn>...</sl-accordion>
+ * ```
+ *
+ * @slot default - The place for multiple <sl-accordion-item>
+ */
+export class Accordion extends ScopedElementsMixin(LitElement) {
+  /** @private */
+  static get scopedElements(): ScopedElementsMap {
+    return {
+      'sl-accordion-item': AccordionItem
```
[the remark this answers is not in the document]
anna-lach: right :)

packages/components/accordion/src/accordion-item.ts
anna-lach: Because summary should be always added, so I thought that we can assume that it's always there.

jpzwarte: We can't really. We can't force the user to set the `summary` property. In some places you can assume that something is set, but in this case it's dangerous.

packages/components/accordion/src/accordion-item.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      await el.updateComplete;
+      await new Promise(resolve => setTimeout(resolve));
+
+      expect(details).to.have.attribute('open');
+      expect(summary).to.have.attribute('aria-expanded', 'true');
+    });
+
+    it('should open on Space', async () => {
+      summary.focus();
+      await sendKeys({ press: 'Enter' });
```
Diaan, line 65:
```suggestion
      await sendKeys({ press: 'Space' });
```

packages/components/accordion/src/accordion-item.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+      await el.updateComplete;
+      await new Promise(resolve => setTimeout(resolve));
+
+      expect(details).not.to.have.attribute('open');
+      expect(summary).not.to.have.attribute('aria-expanded', 'true');
+    });
+
+    it('should ignore Space', async () => {
+      summary.focus();
+      await sendKeys({ press: 'Enter' });
```
Diaan, line 143:
```suggestion
      await sendKeys({ press: 'Space' });
```

packages/components/accordion/src/accordion-item.ts
anna-lach: removed :)

packages/components/accordion/src/accordion.ts
```diff
@@ -0,0 +1,63 @@
...
+import { type CSSResultGroup, LitElement, type PropertyValues, type TemplateResult, html } from 'lit';
+import { property, queryAssignedElements } from 'lit/decorators.js';
+import { AccordionItem } from './accordion-item.js';
+import styles from './accordion.scss.js';
+
+/**
+ * An accordion component that can contain accordion-items
+ *
+ * ```html
+ *   <sl-accordionn>...</sl-accordion>
```
Diaan, line 10:
```suggestion
 *   <sl-accordion>...</sl-accordion>
```

website/package.json
```diff
@@ -158,7 +158,7 @@
     "@11ty/eleventy-img": "^3.1.8",
     "@11ty/eleventy-navigation": "^0.3.5",
     "@11ty/eleventy-plugin-syntaxhighlight": "^5.0.0",
-    "@custom-elements-manifest/analyzer": "^0.9.3",
+    "@custom-elements-manifest/analyzer": "^0.9.2",
```
Diaan, line 161: is this deliberate or a merging mistake?

anna-lach: Good one! Merging mistake.

.changeset/six-mails-learn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sl-design-system/accordion": minor
```
jpzwarte, line 2: Please change this to `patch`

PR #1045, opened 2024-03-13
Accordion documentation

DanielleRameau:


packages/components/accordion/src/accordion-item.scss
```diff
@@ -143,6 +143,7 @@ details {
     --_animation-duration: 0.001ms;
 
     animation-duration: var(--_animation-duration);
+    background-color: var(--_background);
```
jpzwarte, line 146: There is no reason why the token should only contain a color, right? Also, it saves 6 characters :) So `background: var(--_background);`

jpzwarte (approved):

PR #1056, opened 2024-03-15
fixed checkbox checked value

Pauline-Kingma:
When a grid includes both a selection column and a filter column, a potential issue may arise. If you select a filter and then proceed to check a checkbox in the selection column, subsequently deselecting it, the checked status of the filtered item might not be visible. However, the filtering functionality remains intact.

Fixed it by changing ?checked to .checked since it should have the same effect, as both the property and attribute control whether the checkbox is checked or not.

### DIFF
--- a/.changeset/mighty-dryers-exist.md
+++ b/.changeset/mighty-dryers-exist.md
@@ -2,4 +2,7 @@
 "@sl-design-system/grid": patch
 ---
 
-fixed checkbox checked value
+Fixed checkbox checked value;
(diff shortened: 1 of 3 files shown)


.changeset/mighty-dryers-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sl-design-system/grid": patch
+---
+
+fixed checkbox checked value
```
jpzwarte, line 5: Please the line with a capital letter (this text will end up in the `CHANGELOG.md` file after release). Please elaborate a bit more what is fixed :)

jpzwarte (approved):

.changeset/mighty-dryers-exist.md
Pauline-Kingma: Added more info

PR #1058, opened 2024-03-15
Tab group improvements

### LINKED ISSUE #1039: Height of tabs
```[tasklist]
### Tasks
- [ ] add 2px of spacing on top and buttom of tabs (height of 40px in total)
```

### LINKED ISSUE #1044: [tabs] Add ability to have line extend further than the tabs in case of sticky tabs
CleanShot 2024-03-13 at 09 41 06@2x

Top tabs are `<dna-tab-bar>`, button tabs is `<sl-tab-group>` with some custom CSS applied:

```css
sl-tab-group {
  // This makes the component occupy the full width in the page layout grid
  grid-column: full-width;

  // This is to make the line run all the way to the page edges
  &::part(container) {
    border-bottom: var(--_border-bottom);
  }

  &::part(wrapper) {
    inline-size: 100%;
    padding-block: 0;
    margin-inline: auto;
    // Integration with `<dna-page-layout>` so it matches the responsive behavior of that component
    max-inline-size: var(--dna-page-layout-max-width);
    padding-inline: var(--dna-page-layout-min-spacing);
  }

  // Remove the original line
  &::part(tab-list) {
    border-bottom: 0;
  }
}
```

### LINKED ISSUE #1046: [tabs] Support links in `<sl-tab>`
So it should be possible to slot an `<a href>` tag in an `<sl-tab>` when you want the contents of the tab to be routed in an application. However atm, there are a couple of problems with that:

1. The default link styling is visible
2. The link tag does not have the size of the sl-tab, making it harder to click the link

CleanShot 2024-03-13 at 10 39 17@2x

```html
<sl-tab>
  <a href="/foo">Registrations</a>
</sl-tab>
```

### LINKED ISSUE #1047: [tabs] Title should have a regular font-width if there is no subtitle

### LINKED ISSUE #1049: [V2 Tabs]
# Title of feature
## This feature is part of the milestone [link to feature]

| <div align="left">**What is the problem we are solving and why?**  |
| ------------- |
|As a user, I want to ___, so I can ___| 
|As a developer, I want ___, so I can ___|
|As a designer, I want to ___, so I can ___| 

```[tasklist]
### User stories
- [ ] link to user story
- [ ] link to user story
- [ ] link to user story
```

jpzwarte:
- Animate the width of the active tab indicator between tabs
- Do not add `aria-controls` if there are no tab panels
- Fixes inability to scroll menu contents
- Increase the block padding in the tabs
- Use the `<sl-menu-button>` instead of a custom button
- Add public CSS custom properties that can be used to set the min/max inline-size of the menu
- Add public CSS custom properties to sl-tab-group for setting the min/max inline-size of the overflow menu
- Use a negative `outline-offset` to work better when scrolling
- Fix popover shared utility to not overwrite any existing min/max size properties on an element
- Add linear gradient fade left & right when you can scroll in that direction
- Replace `alignment` property with `alignTabs`/`align-tabs` property that matches the `justify-content` values
- Add `href` property to `<sl-tab>` for use with a router
- Fixes #1039
- Fixes #1044
- Fixes #1046 
- Fixes #1047


packages/components/tabs/src/tab-group.stories.ts
```diff
@@ -1,276 +1,339 @@
...
+//           <sl-tab-panel>Contents tab 8</sl-tab-panel>
+
+//           <sl-tab>This is the nineth tab</sl-tab>
+//           <sl-tab-panel>Contents tab 9</sl-tab-panel>
+
+//           <sl-tab>This is the tenth tab</sl-tab>
+//           <sl-tab-panel>Contents tab 10</sl-tab-panel>
+//         </sl-tab-group>
+//       </div>`
+// };
```
Diaan, line 339: can this be removed?

packages/components/tabs/src/tab-group.stories.ts
```diff
@@ -1,276 +1,339 @@
...
+    <sl-tab-group .alignTabs=${alignTabs}>
+      <sl-tab>First tab</sl-tab>
+      <sl-tab>Second tab</sl-tab>
+      <sl-tab disabled>Disabled</sl-tab>
+      <sl-tab>Last tab that is longer than the rest</sl-tab>
+
+      <sl-tab-panel>Contents tab 1</sl-tab-panel>
+      <sl-tab-panel>Contents tab 2</sl-tab-panel>
+      <sl-tab-panel>Contents tab 3</sl-tab-panel>
+      <sl-tab-panel>Contents tab 4</sl-tab-panel>
```
Diaan, line 243: The order was tab - panel, tab - panel, tab - panel initially to show that that is a possibility to order the elements in that way. When placing tabs and panels with f.e. Angular it would mean you can have one `for-loop`/`map` to place all your elements, instead of one for the tabs and one for the panels.

packages/components/tabs/src/tab-group.ts
```diff
@@ -32,380 +30,351 @@ const OBSERVER_OPTIONS: MutationObserverInit = {
...
+
+          ${this.showMenu
+            ? html`
+                <sl-menu-button aria-label=${msg('Show all')} fill="ghost">
+                  <sl-icon name="ellipsis" slot="button"></sl-icon>
+                  ${this.menuItems?.map(
+                    menuItem => html`
+                      <sl-menu-item @click=${() => this.#onMenuItemClick(menuItem.tab)} ?disabled=${menuItem.disabled}>
+                        <div class="title">${menuItem.title}</div>
+                        ${menuItem.subtitle ? html`<div class="subtitle">${menuItem.subtitle}</div>` : nothing}
```
anna-lach, line 199: According to previous design and Roald's opinion, can you please remove subtitles from dropdown? So we will always need to have titles only in the menu.

packages/components/tabs/src/tab-group.stories.ts
```diff
@@ -1,276 +1,339 @@
...
+        <sl-tab-panel>
+          <sl-button @click=${onPrevious}>Focus previous tab</sl-button>
+        </sl-tab-panel>
+      `;
+    }
+  }
+};
+
+export const Sticky: Story = {
+  render: ({ alignTabs }) => html`
```
anna-lach, line 223: Can you please add an argument of vertical here as well to make it possible to check how sticky can look like in the vertical version?

image

packages/components/tabs/src/tab-group.stories.ts
```diff
@@ -1,276 +1,339 @@
...
-          <sl-tab>This is the nineth tab</sl-tab>
-          <sl-tab-panel>Contents tab 9</sl-tab-panel>
-
-          <sl-tab>This is the tenth tab</sl-tab>
-          <sl-tab-panel>Contents tab 10</sl-tab-panel>
-        </sl-tab-group>
-      </div>`
-};
-
-export const VerticalInSmallContainer: StoryObj = {
```
anna-lach, line 189: So we will not have overflow vertical version of tabs? Like here in vertical version: https://ant.design/components/tabs#tabs-demo-slide

packages/components/tabs/src/tab-group.ts
```diff
@@ -32,380 +30,351 @@ const OBSERVER_OPTIONS: MutationObserverInit = {
...
-              part="listbox"
-              popover
-              role="listbox"
-            >
-              ${this.#allTabs}
+      <div part="container">
+        <div part="wrapper">
+          <div class="fade-container">
+            <div class="fade fade-left"></div>
+            <div class="fade fade-right"></div>
```
Diaan, line 182: can't this be done with pure css? `::before` and `::after`?

packages/components/tabs/src/tab-group.stories.ts
RoaldBoerema: I think we _do_ need a vertical overflow

Diaan (approved):

packages/components/tabs/src/tab-group.ts
jpzwarte: Unfortunately not. The reason for this is also why both `.fade-left` and `.fade-right` are *before* the scroller element: if you hover over a tab, that tab changes `z-index` so it's above the fade elements. If you use `::after`, then you can't get above the fade (same as placing `.fade-right` after the scroller).

packages/components/tabs/src/tab-group.spec.ts
```diff
@@ -1,299 +1,295 @@
...
-            <sl-tab>Tab 3</sl-tab>
-            <sl-tab-panel>Panel 1</sl-tab-panel>
-            <sl-tab-panel>Panel 2</sl-tab-panel>
-            <sl-tab-panel>Panel 3</sl-tab-panel>
-          </sl-tab-group>
-        `);
-      });
+    it('should align tabs to start', () => {
+      expect(el).not.to.have.attribute('align-tabs');
+      expect(el).to.have.style('--_tabs-alignment', 'start');
```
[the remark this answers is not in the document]
jpzwarte: This checks the *default* alignment. And since the default alignment is done in CSS and not through the attribute reflection, I check it here.

packages/components/tabs/src/tab-group.stories.ts
jpzwarte: Fixed!

anna-lach (approved):

PR #1063, opened 2024-03-18
First attempt at how "inverted" state could work for breadcrumbs

jpzwarte:
- Adds a `inverted` property
- Add specific CSS custom properties for inverted colors


jpzwarte:
Closing as this is more of a PoC. And we will discuss this further #1067

PR #1073, opened 2024-03-21
Fix/shared missing events exports

anna-lach:
Missing events exports


packages/components/shared/index.ts
```diff
@@ -13,6 +13,11 @@ export * from './src/decorators/base.js';
 export * from './src/decorators/event.js';
 export * from './src/decorators/observe.js';
 export * from './src/directives/anchor.js';
+export * from './src/events/blur.js';
```
jpzwarte, line 16: Please change this to just export `./src/events.js`; that exports the rest.

jpzwarte (approved):

PR #1170, opened 2024-04-09
Feature/770 design button bar

Diaan:


packages/components/button-bar/src/button-bar.stories.ts
```diff
@@ -20,31 +20,73 @@ export default {
       options: ['start', 'center', 'end', 'space-between']
     }
   },
+  parameters: {
+    viewport: {
+      defaultViewport: 'default'
```
jpzwarte, line 25: Do you need to set this, since it's the default anyway?

jpzwarte (approved):
Perhaps add an "Alignment" story that has several button-bars stacked vertically that have different `align` values?

packages/components/button-bar/src/button-bar.stories.ts
```diff
@@ -20,31 +20,73 @@ export default {
...
+    }
+  },
   render: ({ align, buttons, reverse }) => html`
+    <style>
```
anna-lach, line 29: Could you please also make some changes in the dialog stories? :)

packages/components/button-bar/src/button-bar.stories.ts
Diaan: yes, when you switch back from mobile to default you want this, otherwise is stays on the mobile viewport

packages/components/dialog/src/dialog.scss
```diff
@@ -6,8 +6,8 @@ $md-breakpoint: 600px;
   --_border-radius: var(--sl-border-radius-dialog-mobile);
   --_border: var(--sl-border-width-dialog-default) solid var(--sl-color-dialog-border);
   --_color: var(--sl-color-dialog-foreground);
-  --_footer-align: var(--sl-dialog-footer-justify-content);
-  --_footer-direction: column-reverse;
+  // --_footer-align: var(--sl-dialog-footer-justify-content);
```
anna-lach, line 9: Can we remove those commented lines? :)

anna-lach (approved):

PR #1207, opened 2024-04-18
Fix/972 popover voice over

Diaan:


.changeset/itchy-tables-wave.md
```diff
@@ -0,0 +1,6 @@
+---
+"@sl-design-system/popover": patch
+"@sl-design-system/shared": patch
+---
+
+Fixed accesibility support for popover by using correct aria-labels
```
jpzwarte, line 6: Typo: "accessibility"

packages/components/shared/src/controllers/anchor.ts
```diff
@@ -111,10 +111,10 @@ export class AnchorController implements ReactiveController {
 
     if (anchorElement && !this.#host.hasAttribute('aria-details')) {
       anchorElement.id ||= `sl-anchor-${nextUniqueId++}`;
+      this.#host.id ||= `sl-popover-${nextUniqueId++}`;
```
jpzwarte, line 114: It doesn't always have to be a `<sl-popover>` right? Perhaps change to `sl-anchor-host-${nextUniqueId++}`?

packages/components/shared/src/controllers/anchor.ts
```diff
@@ -111,10 +111,10 @@ export class AnchorController implements ReactiveController {
...
 
-      this.#host.setAttribute('aria-details', anchorElement.id);
+      anchorElement?.setAttribute('aria-details', this.#host.id);
     }
-
```
jpzwarte, line 117: Please leave an empty line between the if block and the next statement

jpzwarte (approved):

packages/components/shared/src/controllers/anchor.ts
```diff
@@ -111,10 +111,10 @@ export class AnchorController implements ReactiveController {
...
     anchorElement?.setAttribute('aria-expanded', expanded.toString());
```
anna-lach, line 119: I've just recalled that there should also be a focus trap inside the popover when it's still open?

packages/components/shared/src/controllers/anchor.ts
Diaan: But this is the replacement for the popovertarget functionality according to the comment above the function. And it isn't referring to the sl-popover element, but it's a prefixed id. I think it's more confusing to name it something else.

packages/components/shared/src/controllers/anchor.ts
Diaan: I don't think they should have by default. They aren't modal either.
Popover should be a "lighter" interaction than a dialog. If you have a lot of stuff the user needs to navigate through; use a dialog, that has focus trapping (hopefully 😉)

https://hidde.blog/dialog-modal-popover-differences/

packages/components/popover/src/popover.ts
```diff
@@ -47,6 +47,13 @@ export class Popover extends LitElement {
    */
   @property() position?: PopoverPosition = 'bottom';
 
+  /**
+   * When the contents of a popover is plain text the opener refers to the popover with 'aria-describedby'
+   * so the contents is read by screenreaders. When the contents of your popover is too long to be read
+   * inline this should be set to true so the user can navigate to the popover content themselves.
+   */
+  @property({ type: Boolean, attribute: 'no-describedby' }) noDescribedby?: boolean;
```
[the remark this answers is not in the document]
Diaan: they don't need to do anything? They just need to set it to true, so the describedby isn't added and the contents isn't read automatically.
It is explained in further detail on the a11y tab, i'll refer to that

anna-lach (approved):

PR #1208, opened 2024-04-18
Feature/1008 replace horizontal tabs with tabs of the sl design system

anna-lach:
Website:
 - Added web component for vertical navigation (tabs) 
 - Using `sl-tab-group` component instead of old tabs
 - Each tab as a separated URL
 - Navigation changes (sticky header on all pages)
 - Vertical navigation added for getting started for designers and developers
 
 SLDS:
- tabs fixes: changing tab using keyboard with link inside, problems with scrolling and sticky element, only tab taking into account in the mutationObserver (not all `selected` elements), added missing background for the `container`


packages/components/tabs/src/tab-group.ts
```diff
@@ -83,11 +83,11 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
    */
   #mutationObserver = new MutationObserver(entries => {
     entries.forEach(entry => {
-      if (entry.attributeName === 'selected' && entry.oldValue === null) {
+      if (entry.attributeName === 'selected' && entry.oldValue === null && entry.target instanceof Tab) {
```
jpzwarte, line 86: Why is this needed?

packages/components/tabs/src/tab-group.ts
```diff
@@ -351,6 +351,10 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
     if (selectedTab !== this.selectedTab) {
       this.tabs?.forEach(tab => tab.toggleAttribute('selected', tab === selectedTab));
 
+      if (selectedTab.renderRoot.querySelector('a[href]')) {
+        (selectedTab.renderRoot.querySelector('a[href]') as HTMLAnchorElement).click();
```
jpzwarte, line 355: Should this not be done on the `<sl-tab>` component itself?

packages/components/tabs/src/tab-group.ts
```diff
@@ -414,7 +418,7 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
       this.menuItems = undefined;
     }
 
-    this.selectedTab?.scrollIntoView();
+    // this.selectedTab?.scrollIntoView(); // TODO: causes unnecessary scrolling down on tab change when there is a sticky element
```
jpzwarte, line 421: Is this issue reproducible somewhere? I'd rather remove the code entirely if that fixes the issue than leave it in here.

.changeset/rotten-pans-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sl-design-system/tabs": patch
+---
+
+Possibility of changing tab with URL by keyboard, added missing background for container part, fixed scrolling, fixed changing selected tab only.
```
jpzwarte, line 5: Please make a list of things fixed/added.

jpzwarte (approved):

packages/components/tabs/src/tab-group.ts
anna-lach: There was an issue, when on the website `sl-select` was added (in the sl-tab-panel) which also has `selected` attribute, we need to take only tab into account here.

packages/components/tabs/src/tab-group.ts
anna-lach: When you'll uncomment it and try to run the website or even story in storybook with sticky tabs, you can see that it's scrolling down or up only to the sticky element, not totally to the top. Should I remove it then?

packages/components/tabs/src/tab-group.ts
jpzwarte: So the tab itself probably doesn't know a user used the keyboard to select it. You can check if a link is present using `selectebTab.href`. 

But perhaps you could just add a `keydown` event listener to `<sl-tab>`? And listen for space/enter etc.?

anna-lach: We are adding possibility to use href there, so on click href just changes the url by default, but it's not working when we're using keyboard navigation there, in my opinion it should be possible.

jpzwarte: I agree, but I think that should be done in tab, not in the tab-group component :)

DanielleRameau:
@anna-lach is it possible to fill out the tabs on full width on mobile?

anna-lach:
> @anna-lach is it possible to fill out the tabs on full width on mobile?

yes, I can change it :)

packages/components/dialog/src/dialog.scss
```diff
@@ -20,9 +18,7 @@ $md-breakpoint: 600px;
   --_padding-inline: var(--sl-space-dialog-mobile-body-inline);
   --_subtitle-font: var(--sl-text-dialog-subheading);
   --_title-font: var(--sl-text-dialog-heading);
-
   --sl-button-bar-vertical: var(--sl-ON);
-  // --sl-button-bar-direction-column: column-reverse;
```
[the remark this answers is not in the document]
anna-lach: no problem :)

website/src/categories/getting-started/developers.md
```diff
@@ -254,3 +254,7 @@ We support the 2 latest versions of Angular. You can find these version in [the
 Server-side rendered web components is a hard problem and the web standards related to this continue to evolve. At the moment we do not support this, but we may look at this again in the future.
 
 </section>
+
+</div>
+<ds-vertical-tabs></ds-vertical-tabs>
+</section>
```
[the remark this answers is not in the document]
anna-lach: Right now it uses base.njk file, should I add separated njk file for getting-started pages?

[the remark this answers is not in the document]
anna-lach: I've added separated njk file.

anna-lach:
> > @anna-lach is it possible to fill out the tabs on full width on mobile?
> 
> @DanielleRameau yes, I can change it :)

For now we are leaving the version without `stretch` tabs on mobile.

packages/components/tabs/src/tab.ts
```diff
@@ -76,6 +82,14 @@ export class Tab extends LitElement {
     }
   }
 
+  #onKeydown(event: KeyboardEvent): void {
+    if (event.key === 'Enter' || event.key === ' ') {
+      if (this.href) {
+        (this.renderRoot.querySelector('a[href]') as HTMLAnchorElement).click();
```
jpzwarte, line 88: Can you add a unit test for this plz?

jpzwarte (approved):

packages/components/tabs/src/tab.ts
anna-lach: Tests added

PR #1245, opened 2024-05-15
Add tags to each story and show that in the sidebar

### LINKED ISSUE #1247: the status of the components (important to show status of components that are experimental/drafts)

jpzwarte:
Fixes #1247


packages/angular/package.json
```diff
@@ -90,7 +90,7 @@
     "karma-jasmine-html-reporter": "~2.1.0",
     "ng-packagr": "^17.3.0",
     "rxjs": "~7.8.1",
-    "storybook": "^8.0.10",
+    "storybook": "^8.1.1",
```
anna-lach, line 93: Are we going to add tags to Angular wrappers? Or it's rather a separated category, so I'm not sure how we should treat those elements? :)

image

packages/components/grid/src/stories/basics.stories.ts
```diff
@@ -10,7 +10,7 @@ import { type GridColumnDataRenderer } from '../column.js';
 type Story = StoryObj;
 
 export default {
-  title: 'In progress/Grid/Basics'
+  title: 'Layout/Grid/Basics'
```
anna-lach, line 13: Do we want to set `draft` status for grid as well? :)

anna-lach (approved):

packages/angular/package.json
jpzwarte: I think the angular wrappers should reflect the web components they're based on.

packages/components/grid/src/stories/basics.stories.ts
jpzwarte: Grid is a separate folder atm. I'd have to look into that.

PR #1248, opened 2024-05-16
Rename sl-textarea to sl-text-area

jpzwarte:


.changeset/small-apples-brake.md
```diff
@@ -0,0 +1,6 @@
+---
+"@sl-design-system/text-area": patch
```
anna-lach, line 2: Or maybe `minor`? Because it's bigger change than just a fix? I mean it's going to be a breaking change :)

anna-lach (approved):

PR #1251, opened 2024-05-16
Add search field component

### LINKED ISSUE #1146: [Search Field] Development

jpzwarte:
Fixes #1146


packages/components/search-field/src/search-field.ts
```diff
@@ -0,0 +1,74 @@
...
+  /** @internal Emits when the user presses enter. */
+  @event({ name: 'sl-search' }) searchEvent!: EventEmitter<SlSearchEvent>;
+
+  override renderPrefix(): TemplateResult {
+    return html`<sl-icon name="search"></sl-icon>`;
+  }
+
+  override renderSuffix(): TemplateResult | typeof nothing {
+    return this.value && !this.disabled
+      ? html`<button @click=${this.#onClick}><sl-icon name="xmark"></sl-icon></button>`
```
anna-lach, line 45: Not `sl-button`, just a `button`?

packages/components/search-field/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@sl-design-system/search-field",
+  "version": "0.0.0",
+  "description": "Text field component for the SL Design System",
```
Diaan, line 4:
```suggestion
  "description": "Search field component for the SL Design System",
```

packages/components/search-field/src/search-field.scss
```diff
@@ -0,0 +1,15 @@
...
+
+button {
+  align-items: center;
+  appearance: none;
+  background: transparent;
+  border: 0;
+  cursor: pointer;
+  display: inline-flex;
+  padding: var(--_button-padding);
+}
```
Diaan, line 15: This button is able to get the focus when using keyboard navigation, but it doesn't use the default focus ring (styling), causing it to look like this in Max for example:
image

Diaan: There is of course also a case to be made for it not being focusable with the keyboard. (although I'm not sure if the 'esc' functionality is commonly known)

Diaan (changes requested):

Diaan:
In the acceptance criteria there is mention of a(n optional) button, i don't see that in the code (examples).

packages/components/search-field/src/search-field.ts
jpzwarte: I've changed it to a small `<sl-button>` for now.

packages/components/search-field/src/search-field.scss
jpzwarte: I've changed it to use a small `<sl-button>`. And the button is keyboard focusable.

anna-lach (approved):

PR #1258, opened 2024-05-21
Various form improvements

jpzwarte:


packages/components/form/src/form.ts
```diff
@@ -96,36 +126,55 @@ export class Form<T extends Record<string, unknown> = Record<string, unknown>> e
     return this.controls.map(c => c.reportValidity()).every(Boolean);
   }
 
+  reset(): void {
+    console.log('reset');
```
anna-lach, line 130: Do we need that console.log?

examples/lit/src/nested-form/nested-form.ts
```diff
@@ -0,0 +1,64 @@
...
+          <sl-button @click=${this.#onClick}>Submit</sl-button>
+        </sl-button-bar>
+      </sl-form>
+      <pre>${JSON.stringify(this.#form.value ?? {}, null, 2)}</pre>
+    `;
+  }
+
+  #onClick() {
+    if (this.#form.reportValidity()) {
+      console.log(this.#form.value);
```
anna-lach, line 61: Maybe showing the value in an alert would be better?

anna-lach (approved):

PR #1263, opened 2024-05-22
turbosnap test

Diaan:


.storybook/manager.ts
```diff
@@ -1,5 +1,6 @@
 import { addons, type API, type LeafEntry, type HashEntry } from '@storybook/manager-api';
 import { styled } from '@storybook/theming';
+import slTheme from './slTheme'
```
jpzwarte, line 3: Could you use dasherize naming here? So `import theme from './sl-theme.js';`?

packages/components/avatar/src/avatar.scss
```diff
@@ -67,6 +63,10 @@ $statuses: danger, success, warning, accent, neutral, primary;
   }
 }
 
+:host([image-only]) [part='wrapper'] {
```
jpzwarte, line 66: This is a bugfix for avatar? (not seeing a changeset for it?)

jpzwarte (approved):

PR #1324, opened 2024-06-10
Fix/1216 select hide on scroll

Diaan:


packages/components/shared/src/directives/anchor.ts
```diff
@@ -23,6 +23,15 @@ export class AnchorDirective extends Directive {
   #cleanup?: () => void;
   #config?: AnchorDirectiveConfig;
   #host?: HTMLElement;
+  visible: boolean = true;
```
anna-lach, line 26: Do we use this property?

packages/components/shared/src/directives/anchor.ts
```diff
@@ -23,6 +23,15 @@ export class AnchorDirective extends Directive {
...
+
+  observer = new IntersectionObserver(
+    (entries, _) => {
+      const outOfView = entries.some(e => e.intersectionRatio < 0.5);
+      updatePopoverVisibility(this.#host, outOfView);
+    },
+    { threshold: 0.5 }
```
anna-lach, line 33: Maybe we can add possibility for developers to set `rootMargin` there? Or whole `options` object for the intersection observer? Then they would be able to fit it to their needs (like sticky header size)?  And then we can also show it in the story example with sticky header that it could hide below the sticky header.

anna-lach (approved):

packages/components/shared/src/directives/anchor.ts
Diaan: Oohhh, that's smart I'll look into that.

jpzwarte (approved):

packages/components/select/src/select.ts
```diff
@@ -114,6 +114,10 @@ export class Select<T = unknown> extends FormControlMixin(ScopedElementsMixin(Li
   /** The size of the select. */
   @property({ reflect: true }) size: SelectSize = 'md';
 
+  /** The number of pixels from the top of the viewport the select should be hidden on scroll.
+   * Use this when there is a sticky header you don't want dropdowns to fall on top of. */
+  @property({ attribute: 'hide-margin-top' }) hideMarginTop: number = 0;
```
anna-lach, line 119: or maybe `stickyMarginTop`? Hard to say how to name it properly :)

anna-lach (approved):

PR #1325, opened 2024-06-11
Reuse the badge component in avatar

jpzwarte:


packages/components/avatar/src/avatar.ts
```diff
@@ -92,407 +26,133 @@ export class Avatar extends ScopedElementsMixin(LitElement) {
...
-   * @type {'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl'}
-   */
+  /** The size of the avatar. */
   @property({ reflect: true }) size: AvatarSize = 'md';
 
-  /**
-   * Optional user status to show.
-   * @type {'danger' | 'success' | 'warning' | 'accent' | 'neutral' | 'primary'}
-   */
-  @property({ reflect: true }) status?: UserStatus;
```
anna-lach, line 163: Can you please change examples on the website as well? There are few with `status` left :)

image

packages/components/avatar/src/avatar.scss
```diff
@@ -1,74 +1,96 @@
...
+
+  slot:not([name]) {
+    display: none;
   }
 }
 
-:host(:focus-visible) {
-  outline: var(--_outline);
-  outline-offset: var(--sl-border-width-focusring-offset);
+:host([size='lg']) {
```
anna-lach, line 44: What about using `@each` for sizes like in the buttons?

anna-lach (approved):

packages/components/avatar/src/avatar.ts
```diff
@@ -92,407 +43,133 @@ export class Avatar extends ScopedElementsMixin(LitElement) {
...
+                <slot name="fallback">
+                  <span part="initials">${this.initials}</span>
+                </slot>
+              `}
+        </div>
+      </div>
       ${this.imageOnly
         ? nothing
         : html`
             <sl-tooltip id="avatar-tooltip">${this.displayName}</sl-tooltip>
```
Diaan, line 131: This doesn't work anymore

Diaan (changes requested):

Diaan:
Not sure this is how it is supposed to look? 
image
Wrong icon and wrong icon size

packages/components/avatar/src/avatar.ts
```diff
@@ -1,85 +1,36 @@
...
+ * @csspart avatar - The container for positioning the badge.
+ * @csspart initials - The initials to display when no picture is set.
  * @csspart name - The display name, either a <span> or <a> if `href` is set.
+ * @csspart picture - The element containing the image, initials or fallback content.
  * @csspart wrapper - The wrapper element around the image and name.
  *
- * @cssprop --sl-avatar-max-inline-size - Max inline-size of the container in vertical mode. If not set it will behave like a regular `display: block` element.
+ * @slot badge - The badge to display on the avatar.
+ * @slot default - The subheading of the avatar.
+ * @slot fallback - The fallback content to display when no picture is set.
```
Diaan, line 33: Why would you want this as a slot? You want this to be the same over your entire application (with a possible overwrite, but not sure why you would want that.) so I would suggest an `sl-icon` which we choose.

Diaan: For the sl-icon we can use the same setup we have for the breadcrumbs; there we use an icon named `breadcrumb-separator`, that points to the `slash-forward` icon by default but for magister it's `chevron-right`

packages/components/avatar/src/avatar.scss
```diff
@@ -1,74 +1,96 @@
...
+  --_heading-font: var(--sl-text-avatar-header-3xl);
+  --_initials-font: var(--sl-text-avatar-initials-3xl);
+  --_picture-size: var(--sl-size-avatar-3xl);
+  --_subheading-font: var(--sl-text-avatar-subheader-3xl);
 }
 
-:host([image-only]) [part='wrapper'] {
-  grid-template: 'picture' auto;
+:host([image-only]) {
+  border-radius: 50%;
```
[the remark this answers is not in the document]
jpzwarte: Yep, discussed this with Roald: square isn't being used anywhere atm.

packages/components/avatar/src/avatar.spec.ts
```diff
@@ -1,394 +1,143 @@
...
-      expect(name).to.match('span');
-      expect(subheader).not.to.exist;
-    });
+    expect(wrapper?.childElementCount).to.equal(1);
+    expect(wrapper?.children[0]).to.match('[part="avatar"]');
+    expect(name).not.to.exist;
+  });
+
+  it('should not render an image', () => {
+    expect(el.renderRoot.querySelector('img')).not.to.exist;
```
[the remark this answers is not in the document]
jpzwarte: Not really. That test explicitly tests whether the fallback content is rendered. I could test it there, but then it feels like i'm testing 2 things in 1 test.

packages/components/avatar/src/avatar.spec.ts
```diff
@@ -1,394 +1,143 @@
...
+
+  it('should wrap the contents in a link tag if href is set', async () => {
+    el.href = 'https://www.example.com';
+    await el.updateComplete;
+
+    const wrapper = el.renderRoot.querySelector('[part="wrapper"]');
+
+    expect(wrapper).to.have.attribute('href', 'https://www.example.com');
+    expect(wrapper).to.match('a');
   });
```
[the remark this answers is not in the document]
jpzwarte: That's more a visual test i think. That's why we have Chromatic, right? ;)

packages/components/avatar/src/avatar.ts
jpzwarte: Fixed

PR #1360, opened 2024-06-24
Feature/1339 poc format date component

anna-lach:


packages/components/format-date/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "url": "https://github.com/sl-design-system/components/issues"
+  },
+  "type": "module",
+  "main": "./index.js",
+  "module": "./index.js",
+  "types": "./index.d.ts",
+  "customElements": "custom-elements.json",
+  "exports": {
+    ".": "./index.js",
+    "./format-date-time.js": "./src/format-date-time.js",
```
jpzwarte, line 25: `import { formatDateTime } from '@sl-design-system/format-date/format-date-time.js';` is kind of repeating :) In format number, i have `import { format } from '@sl-design-system/format-number/format.js';` Perhaps do the same here?

packages/components/format-date/src/format-date.stories.ts
```diff
@@ -0,0 +1,160 @@
+import { type Meta, type StoryObj } from '@storybook/web-components';
+import { type TemplateResult, html, nothing } from 'lit';
+import '../register.js';
+import { type FormatDate } from './format-date.js';
+import { formatDateTime } from './format-date-time';
+
+interface Props extends FormatDate {
```
jpzwarte, line 7: `type Props = Pick<FormatDate, 'date' | 'locale' | etc> & { fallback: TemplateResult };`

At the moment, `Props` has *all* properties of LitElement etc.

packages/components/format-date/src/format-date.stories.ts
```diff
@@ -0,0 +1,160 @@
...
+      .timeZone=${timeZone}
+      .hour12=${hour12}>
+      ${fallback}
+    </sl-format-date>
+  `
+} satisfies Meta<Props>;
+
+export const Basic: Story = {
+  args: {
+    date: new Date,
```
jpzwarte, line 108: Should always use parens? `date: new Date(),`?

packages/components/format-date/src/format-date.stories.ts
```diff
@@ -0,0 +1,160 @@
...
+};
+
+export const InvalidDate: Story = {
+  args: {
+    fallback: 'This date is not valid and I cannot render it.',
+  }
+};
+
+export const Function: StoryObj = {
+  render: ({ date, locale, weekday, era, year, month, day, dayPeriod, hour, minute, second, timeZoneName, timeZone, hour12 }) => {
```
jpzwarte, line 120: Have you tried `render: props => {` here? and then `...props` into the options?

packages/components/format-date/src/format-date.stories.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+};
+
+export const InvalidDate: Story = {
+  args: {
+    fallback: 'This date is not valid and I cannot render it.',
+  }
+};
+
+export const Function: StoryObj = {
```
jpzwarte, line 119: Not sure we should have this here. We're already implicitly testing it via `<sl-format-date>`. Perhaps only have unit tests for the function?

packages/components/format-date/src/format-date.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+}
+
+/**
+ * A format date component for formatting date and time.
+ *
+ * @slot default - A place for the fallback when there is no valid date/time applied.
+ */
+export class FormatDate extends LocaleMixin(LitElement) {
+  #date?: Date = new Date();
```
jpzwarte, line 18: Won't this cause the fallback never to be rendered? Unless you first set the `date` property to null/undefined/something that is not a date?

jpzwarte: My initial idea with `<dna-format-date>` is that if the developer does not set a date, or it is undefined, then it by default shows the slot content.

packages/components/format-date/src/format-date.ts
```diff
@@ -0,0 +1,100 @@
...
+      this.#date = new Date(value);
+    } else {
+      this.#date = undefined;
+    }
+
+    this.requestUpdate('date', oldValue);
+  }
+
+  /** The date/time to format. If not set, the current date and time will be used. */
+  @property()
```
jpzwarte, line 79: I think that if you move the decorator to the setter, then you don't need the `this.requestUpdate('date', ...)`

jpzwarte: The docs probably changed at some point. It now shows the decorator on the setter: https://lit.dev/docs/components/properties/#accessors-custom

packages/components/format-date/src/format-date.ts
```diff
@@ -0,0 +1,100 @@
+import { LocaleMixin } from '@sl-design-system/shared';
```
jpzwarte, line 1: This should be `from '@sl-design-system/shared/mixins.js';` right?

jpzwarte (approved):

packages/components/format-date/src/format-date.stories.ts
```diff
@@ -0,0 +1,160 @@
...
+  fallback: string | TemplateResult;
+}
+type Story = StoryObj<Props>;
+
+export default {
+  title: 'Utilities/Format date',
+  tags: ['draft'],
+  args: {
+    fallback: 'invalid date',
+    locale: 'en-US',
```
jpzwarte, line 17: If you check the number-field branch, i have an argType for locale as well, with all the locales in use within SL :)

jpzwarte: I also know that some people don't really like US english, since we're a European company :) I think `'en'` is a better default. Or leave it blank and have it format automatically based on the browser's language.

packages/components/format-date/src/format-date.stories.ts
anna-lach: I wanted to show explicitly that people can also use this function only, but then they need to remember about checking if the date is really a date etc., should I remove it?

packages/components/format-date/src/format-date.stories.ts
```diff
@@ -0,0 +1,160 @@
...
+      control: 'inline-radio',
+      options: ['numeric', '2-digit']
+    },
+    minute: {
+      control: 'inline-radio',
+      options: ['numeric', '2-digit']
+    },
+    second: {
+      control: 'inline-radio',
+      options: ['numeric', '2-digit']
```
Diaan, line 56: maybe also add the option `none` here, so you can "reset" these options.

packages/components/format-date/src/format-date.stories.ts
```diff
@@ -0,0 +1,130 @@
...
+export default {
+  title: 'Utilities/Format date',
+  tags: ['draft'],
+  args: {
+    fallback: 'invalid date'
+  },
+  argTypes: {
+    locale: {
+      control: 'inline-radio',
+      options: ['de', 'en', 'es', 'fi', 'it', 'nl', 'no', 'pl', 'sv']
```
Diaan, line 34: it says here `en` but that uses the US notation, not the UK one. Maybe we should give that one as an example?

packages/components/format-date/src/format-date.stories.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export const Basic: Story = {
+  args: {
+    fallback: 'This is not a valid date'
+  }
+};
+
+export const InvalidDate: Story = {
+  args: {
+    fallback: 'This date is not valid and I cannot render it.'
```
Diaan, line 128:
```suggestion
    fallback: 'This date is not valid and it cannot be rendered.'
```

packages/components/format-date/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './src/format-date.js';
+export * from './src/format.js';
```
jpzwarte, line 2: Please export this separately, so that if somebody only wants to use the `format()` function, the web component isn't also loaded.

packages/components/format-date/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "url": "https://github.com/sl-design-system/components/issues"
+  },
+  "type": "module",
+  "main": "./index.js",
+  "module": "./index.js",
+  "types": "./index.d.ts",
+  "customElements": "custom-elements.json",
+  "exports": {
+    ".": "./index.js",
+    "./format.js": "./src/format.js",
```
jpzwarte, line 25: This is sufficient, the export can be removed from `index.ts`

packages/components/format-date/src/format-date.stories.ts
```diff
@@ -0,0 +1,146 @@
+import { type Meta, type StoryObj } from '@storybook/web-components';
+import { html } from 'lit';
+import '../register.js';
+import { type FormatDate } from './format-date.js';
+
+type Props = Pick<
+  FormatDate,
+  | 'locale'
```
jpzwarte, line 8: Please sort alphabetically

packages/components/format-date/src/format-date.stories.ts
```diff
@@ -0,0 +1,146 @@
...
+  `
+} satisfies Meta<Props>;
+
+export const Basic: Story = {
+  args: {
+    fallback: 'This is not a valid date'
+  }
+};
+
+export const InvalidDate: Story = {
```
jpzwarte, line 142: Better to call this story "Fallback"

packages/components/format-date/src/format-date.ts
```diff
@@ -0,0 +1,154 @@
...
+      this.#date = undefined;
+    }
+  }
+
+  get date(): Date | undefined {
+    return this.#date;
+  }
+
+  override render(): TemplateResult {
+    return html` ${!this.date ? html`<slot></slot>` : this.#formatDateTime(this.date)} `;
```
jpzwarte, line 122: Unnecessary whitespace

packages/components/format-date/src/format-date.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  #formatDateTime(date: Date): string {
+    const localeString = this.locale ? this.locale : 'en',
+      {
+        dateStyle,
```
jpzwarte, line 128: Sort alphabetically plz

jpzwarte (approved):

Diaan (approved):

PR #1365, opened 2024-06-27
Feature/1172 toggle button UI design

Diaan:


packages/components/toggle-button/src/toggle-button.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    it('should have a button role', () => {
+      expect(el.role).to.equal('button');
+    });
+
+    it('should have a tabindex', () => {
+      expect(el).to.have.attribute('tabindex', '0');
+    });
+
+    it('should allow for a custom tabindex', async () => {
```
jpzwarte, line 30: This should be removed, right?

packages/components/toggle-button/src/toggle-button.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+    it('should toggle the state on click', () => {
+      const clickEvent = new Event('click');
+      const onToggle = spy();
+      expect(el).not.to.have.attribute('pressed');
+
+      el.addEventListener('sl-toggle', onToggle);
+
+      el.dispatchEvent(clickEvent);
+
+      expect(onToggle).to.have.been.calledOnce;
```
jpzwarte, line 91: Perhaps also test whether the event has the correct boolean detail value? From my current branch:

```
    it('should emit an sl-toggle event when button is clicked', async () => {
      const button = el.renderRoot.querySelector('button'),
        onToggle = spy();

      el.addEventListener('sl-toggle', (event: SlToggleEvent<boolean>) => {
        onToggle(event.detail);
      });

      button?.click();
      await el.updateComplete;

      expect(onToggle).to.have.been.calledOnce;
      expect(onToggle.lastCall.args[0]).to.be.true;

      button?.click();
      await el.updateComplete;

      expect(onToggle).to.have.been.calledTwice;
      expect(onToggle.lastCall.args[0]).to.be.false;
    });
```

I did need to make a small change in the `SlToggleEvent` type definition for this though.

packages/components/toggle-button/src/toggle-button.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+    beforeEach(async () => {
+      el = await fixture(html`
+        <sl-toggle-button pressed>
+          <sl-icon name="far-gear"></sl-icon>
+          <sl-icon name="fas-gear" slot="pressed"></sl-icon>
+        </sl-toggle-button>
+      `);
+    });
+
+    it('should not be pressed by default', () => {
```
jpzwarte, line 148: Description doesn't match the test

packages/components/toggle-button/src/toggle-button.stories.ts
```diff
@@ -0,0 +1,180 @@
...
+    `;
+  }
+} satisfies Meta<Props>;
+
+export const Basic: Story = {};
+
+export const Errors: Story = {
+  render: () => {
+    return html`
+      <p></p>
```
jpzwarte, line 72: Typo with the closing tag

packages/components/toggle-button/src/toggle-button.ts
```diff
@@ -0,0 +1,128 @@
...
+  @queryAssignedElements({ slot: 'pressed' }) pressedIcon?: Icon[];
+
+  /** @internal Emits when the button item has been toggled. */
+  @event({ name: 'sl-toggle' }) toggleEvent!: EventEmitter<SlToggleEvent<boolean>>;
+
+  /** Whether the toggle-button is disabled; when set no interaction is possible. */
+  @property({ type: Boolean, reflect: true }) disabled?: boolean;
+
+  /** The variant of the toggle-button. */
+  @property({ reflect: true }) fill: ToggleButtonFill = 'ghost';
```
jpzwarte, line 53: I still think this should be the default (`undefined`). What happens if somebody does `<sl-toggle-button fill="Outline">` in HTML?

packages/components/toggle-button/src/toggle-button.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  /** The state of the toggle-button. */
+  @property({ type: Boolean, reflect: true }) pressed = false;
```
jpzwarte, line 56: I always do `pressed?: boolean;` in these cases. The main different is that if `undefined`, then the lifecycle methods won't be called for this property. If you initialize to `false`, they will be called.

packages/components/toggle-button/src/toggle-button.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  /** The size of the toggle-button component. */
+  @property({ reflect: true }) size: ToggleButtonSize = 'md';
```
jpzwarte, line 59: Same argument here as with `fill`.

packages/components/toggle-button/src/toggle-button.ts
jpzwarte: I know your code matches how we do things in Button, but this kinda falls under "continued insight" :)

packages/components/toggle-button/src/toggle-button.ts
```diff
@@ -0,0 +1,128 @@
...
+  static override styles: CSSResultGroup = styles;
+
+  // eslint-disable-next-line no-unused-private-class-members
+  #events = new EventsController(this, {
+    click: this.#onClick,
+    keydown: this.#onKeydown
+  });
+
+  /** @internal The slotted default icon. */
+  @queryAssignedElements() defaultIcon?: Icon[];
```
jpzwarte, line 41: This will query the DOM every time you access this variable. Why not set this variable from the slotchange event handler?

packages/components/toggle-button/src/toggle-button.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  /** @internal The slotted pressed state icon. */
+  @queryAssignedElements({ slot: 'pressed' }) pressedIcon?: Icon[];
```
jpzwarte, line 44: Same

packages/components/toggle-button/src/toggle-button.ts
```diff
@@ -0,0 +1,128 @@
...
+      console.error('There needs to be an sl-icon in the "pressed" slot for the component to work');
+      this.setAttribute('error', 'true');
+    }
+
+    if (changes.has('size')) {
+      this.#setIconProperties();
+    }
+
+    if (this.pressedIcon?.[0]?.name === this.defaultIcon?.[0]?.name) {
+      console.error(
```
jpzwarte, line 84: I'm not sure we do this anywhere yet, but perhaps only do this when NOT in production? I'm not sure we should do this in *this* PR though. And it requires build changes as well. From Lit: 
CleanShot 2024-07-10 at 14 47 21@2x

website/src/categories/components/toggle-button/code.md
```diff
@@ -0,0 +1,43 @@
...
+---
+
+<section class="no-heading">
+<div class="ds-example">
+  <div class="ds-example__code-wrapper">
+    <sl-toggle-button>
+      <sl-icon name="far-gear"></sl-icon>
+      <sl-icon name="fas-gear" slot="pressed"></sl-icon>
+    </sl-toggle-button>
+      <sl-toggle-button fill="outline" pressed>
```
jpzwarte, line 17: Indentation doesn't match

jpzwarte (approved):

.changeset/shy-feet-perform.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sl-design-system/toggle-button': patch
+---
+
+Added toggle button component
```
anna-lach, line 5: Can you please also add changeset for tokens? :)

packages/components/toggle-button/src/toggle-button.scss
```diff
@@ -0,0 +1,188 @@
...
+  --_focus-outline: var(--sl-color-focusring-default) solid var(--sl-border-width-focusring-default);
+  --_transition-duration: var(--sl-animation-button-duration);
+  --_transition-easing: var(--sl-animation-button-easing);
+  --_border-width: var(--sl-border-width-button-ghost);
+  --_background: var(--sl-color-button-default-ghost-idle-background);
+  --_border-color: var(--sl-color-button-default-ghost-idle-border);
+  --_color: var(--sl-color-button-default-ghost-idle-foreground);
+  --_padding-block: var(--sl-space-button-ghost-icon-only-md);
+  --_padding-inline: var(--sl-space-button-ghost-icon-only-md);
+  --_border-radius: var(--sl-border-radius-button-md);
```
anna-lach, line 17: Do we want to order the variables alphabetically?

anna-lach (approved):

packages/components/toggle-button/src/toggle-button.ts
Diaan: In Icon we have something similar: 
image

packages/components/toggle-button/src/toggle-button.ts
Diaan: Because if nothing is set in the slot the slotchange handler isn't called. So I needed some other trigger to do the check (and a way to then check if there are indeed no icons in the `pressed` slot)

PR #1369, opened 2024-06-28
Fix behavior of checkbox (group) to match the native behavior

jpzwarte:


packages/components/checkbox/src/checkbox-group.ts
```diff
@@ -37,7 +37,7 @@ export class CheckboxGroup<T = unknown> extends FormControlMixin(LitElement) {
 
   /** Observe changes to the checkboxes. */
   #observer = new MutationObserver(() => {
```
anna-lach, line 39: Can you please update the example on the website? Right now it's not rendering
image

image

Diaan (approved):

packages/components/checkbox/src/checkbox-group.ts
jpzwarte: Fixed

anna-lach (approved):

PR #1374, opened 2024-07-01
Minor dialog fixes

jpzwarte:
- Reset the `margin` on elements in the (sub)title slots
- Use `<h1>` and `<h2>` in the storybook dialog examples
- Reduce the padding in dialogs from 30px to 24px for desktop and from 24px to 16px on mobile

### DIFF
--- a/packages/tokens/src/core.json
+++ b/packages/tokens/src/core.json
@@ -1058,32 +1058,34 @@
           "paddingTop": {
             "value": "{space.xl}",
             "type": "spacing",
-            "description": "space.button.solid.block.md"
+            "description": "space.dialog.desktop.container.paddingTop"
           },
           "inline": {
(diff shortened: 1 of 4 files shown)


packages/tokens/src/core.json
```diff
@@ -1058,32 +1058,34 @@
           "paddingTop": {
             "value": "{space.xl}",
             "type": "spacing",
-            "description": "space.button.solid.block.md"
+            "description": "space.dialog.desktop.container.paddingTop"
```
Diaan, line 1061: should this not be "blockStart"?

Diaan (approved):

packages/tokens/src/core.json
jpzwarte: Will delay this rename until contextual tokens.

PR #1377, opened 2024-07-01
Feature/1358 progress bar component

anna-lach:


packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -0,0 +1,98 @@
+import { localized, msg } from '@lit/localize';
+import { type ScopedElementsMap, ScopedElementsMixin } from '@open-wc/scoped-elements/lit-element.js';
```
jpzwarte, line 2: This should be added to `package.json` as a dependency

packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -0,0 +1,98 @@
...
+@localized()
+export class ProgressBar extends ScopedElementsMixin(LitElement) {
+  /** @internal */
+  static get scopedElements(): ScopedElementsMap {
+    return {
+      'sl-icon': Icon
+    };
+  }
+
+  /** @private */
```
jpzwarte, line 39: Please use `@internal` here as well

packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -0,0 +1,98 @@
...
+  static override styles: CSSResultGroup = styles;
+
+  /** Progress value (from 0...100). */
+  @property({ type: Number }) value = 0;
```
jpzwarte, line 43: Please sort all the variables alphabetically

packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /** Whether the progress bar has the indeterminate state. */
+  @property({ type: Boolean, reflect: true }) indeterminate = false;
+
+  /** Label describing the value of the progress bar. */
+  @property() label?: string;
+
+  /** The state of the progress bar. */
+  @property({ reflect: true }) state: ProgressState = 'active';
```
jpzwarte, line 52: If the default state is `'active'`, why not make that `undefined` instead? Then you could simply have it as default styling within CSS as well.

packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -0,0 +1,98 @@
...
+        return 'octagon-exclamation-solid';
+      case 'error':
+        return 'triangle-exclamation-solid';
+      default:
+        return 'circle-check-solid';
+    }
+  }
+
+  override render(): TemplateResult {
+    return html` ${this.label
```
jpzwarte, line 69: Please add a newline after the opening quote

packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -0,0 +1,98 @@
...
+        ? html`<span id="label" class="label">
```
jpzwarte, line 70: Same here

packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -0,0 +1,98 @@
...
+            ${this.label}
+            ${this.state !== 'active' ? html` <sl-icon .name=${this.iconName} size="md"></sl-icon>` : nothing}
```
jpzwarte, line 72: superfluous whitespace after the opening `

packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -0,0 +1,98 @@
...
+      <div
+        aria-labelledby=${ifDefined(this.label ? 'label' : undefined)}
+        aria-describedby="helper"
+        class="container"
+        role="progressbar"
+        aria-valuemin="0"
+        aria-valuenow=${ifDefined(!this.indeterminate ? `${this.value}` : undefined)}
+        aria-valuemax="100"
+      >
+        <div class="progress" style="inline-size: ${this.value}%"></div>
```
jpzwarte, line 84: You probably want to use `scale` instead of `inline-size`: `scale` can be animated on the GPU, `inline-size` is animated on the CPU.

packages/components/progress-bar/src/progress-bar.stories.ts
```diff
@@ -0,0 +1,241 @@
...
+            progressBar.value = Math.round(currentProgress + step);
+            return currentProgress + step;
+          } else {
+            clearInterval(interval);
+            return size;
+          }
+        }, 300);
+      }, 2500);
+    });
+    return html` <style>
```
jpzwarte, line 164: newline after ` please

packages/components/progress-bar/src/progress-bar.stories.ts
```diff
@@ -0,0 +1,241 @@
...
+    label: 'Generating report',
+    slot: () => html`<span>This may take a few minutes</span>`
+  }
+};
+
+export const Download: StoryObj = {
+  args: {
+    ...Basic.args
+  },
+  argTypes: {
```
jpzwarte, line 110: Move the `argTypes` to the `default` export. Atm this is copied in the `All` story

packages/components/progress-bar/src/progress-bar.stories.ts
```diff
@@ -0,0 +1,241 @@
...
+    }
+  },
+  parameters: {
+    viewport: {
+      defaultViewport: 'default'
+    }
+  },
+  render: ({ indeterminate, state, value, label, slot }) => {
+    return html`
+      <sl-progress-bar .indeterminate=${indeterminate} .value=${value} .label=${label} .state=${state}>
```
jpzwarte, line 45: Use `?indeterminate`, since it's a boolean

packages/components/progress-bar/src/progress-bar.scss
```diff
@@ -0,0 +1,109 @@
...
+  gap: var(--_label-gap);
+  justify-content: space-between;
+
+  sl-icon {
+    --sl-icon-fill-default: var(--_icon-fill);
+  }
+}
+
+.helper {
+  display: flex;
```
jpzwarte, line 58: The `.helper` is a `<span>`, but you're styling it as a block element. Better to make it a `<div>` then.

packages/components/progress-bar/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "test": "echo \"Error: run tests from monorepo root.\" && exit 1"
+  },
+  "dependencies": {
+    "@sl-design-system/icon": "^1.0.0"
+  },
+  "devDependencies": {
+    "lit": "^3.1.4"
+  },
+  "peerDependencies": {
+    "lit": "^3.1.2"
```
jpzwarte, line 47: Please sync the lit version

packages/components/progress-bar/src/progress-bar.scss
```diff
@@ -0,0 +1,109 @@
...
+  }
+}
+
+:host([state='error']) {
+  ::slotted(*) {
+    color: var(--_helper-color-error);
+  }
+}
+
+.label {
```
jpzwarte, line 46: Label is styles as a block element: use `<div>` instead of `<span>`.

packages/components/progress-bar/src/progress-bar.scss
```diff
@@ -0,0 +1,109 @@
...
+
+@each $state in $states {
+  :host([state='#{$state}']) {
+    --_progress-background: var(--sl-color-progressbar-#{$state}-track);
+    --_icon-fill: var(--sl-color-progressbar-#{$state}-track);
+  }
+}
+
+:host([indeterminate][state='active']) {
+  .progress {
```
jpzwarte, line 34: Move to previous line (no need for a nested block if it's the only rule)

packages/components/progress-bar/src/progress-bar.scss
```diff
@@ -0,0 +1,109 @@
...
+  --_border-radius: var(--sl-border-radius-full);
+  --_label-font: var(--sl-text-input-field-label-label-md);
+  --_helper-font: var(--sl-text-input-field-label-hint-md);
+  --_helper-color: var(--sl-color-input-helper-text-default);
+  --_helper-color-error: var(--sl-color-progressbar-error-track);
+  --_height: var(--sl-size-progressbar);
+  --_gap: var(--sl-space-progressbar-vertical-gap);
+  --_label-gap: var(--sl-space-progressbar-horizontal-gap);
+  --_indeterminate-duration: 20s;
+  --_bar-background: var(--sl-color-progressbar-background);
```
jpzwarte, line 13: Please sort alpabetically

packages/components/progress-bar/src/progress-bar.scss
```diff
@@ -0,0 +1,109 @@
+$states: active, success, warning, error;
+
+:host {
+  --_border-radius: var(--sl-border-radius-full);
+  --_label-font: var(--sl-text-input-field-label-label-md);
+  --_helper-font: var(--sl-text-input-field-label-hint-md);
+  --_helper-color: var(--sl-color-input-helper-text-default);
+  --_helper-color-error: var(--sl-color-progressbar-error-track);
+  --_height: var(--sl-size-progressbar);
```
jpzwarte, line 9: Since you already have `--_bar-background`, rename this to `--_bar-thickness` perhaps?

packages/components/progress-bar/src/progress-bar.scss
```diff
@@ -0,0 +1,109 @@
...
+
+:host([indeterminate][state='active']) {
+  .progress {
+    animation: var(--_indeterminate-duration) ease 500ms infinite normal none running indeterminate;
+    position: relative;
+  }
+}
+
+:host([state='error']) {
+  ::slotted(*) {
```
jpzwarte, line 41: Can also be on the same line as the `:host` one

jpzwarte (approved):

jpzwarte:
Please also add changesets :)

packages/components/progress-bar/src/progress-bar.ts
anna-lach: I prefer to leave the `active` here if it's ok. It's used for example here:
image


and also for the aria-live with translations etc.
image

jpzwarte: For your first question, you could simply make those the default.
For the second: i'm not really sure what `${msg('state')} ${msg(this.state)}` is supposed to do?

Also, what happens if somebody does `state=""` or `.state=${undefined}`?

packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -0,0 +1,98 @@
...
+import { ifDefined } from 'lit/directives/if-defined.js';
+import styles from './progress-bar.scss.js';
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'sl-progress-bar': ProgressBar;
+  }
+}
+
+export type ProgressState = 'active' | 'success' | 'warning' | 'error';
```
Diaan, line 15: Is this a state or a variant? 
I feel a state is something that changes based on a user action within the component. This is an input, a choice by the designer/developer.

packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Progress bar component that can be used to communicate process status.
+ * Hidden aria-live element will help making the progress bar more accessible and should cause fewer bugs in products than leaving it up for the developers.
```
Diaan, line 19: this is not very nice towards the developers of the products ;)
Maybe it's nicer to say they don't need to worry about it than to say they'll probably mess it up :P

Diaan (approved):

packages/components/progress-bar/src/progress-bar.ts
anna-lach: I was thinking about `status` there instead of `state`, but due to discussion with Danielle and to make it the same as in Figma I've changed it to `state`. @DanielleRameau what do you think about renaming it to `variant`?

We are also using state to inform a user that sth has changed (in aria-live) so maybe it would be better to leave `state` here?

packages/components/progress-bar/src/progress-bar.ts
anna-lach: Right 😆

packages/components/progress-bar/src/progress-bar.ts
jpzwarte: I like `variant`, but just because it would then match with other components that have `variant` and `success`, `warning` values. Not sure if that is reason enough. But i also agree with @Diaan .

packages/components/progress-bar/src/progress-bar.ts
anna-lach: For the second - This is used to inform the user via screen reader about state changes, the default is `active` so there will be no such situation with `undefined` or `""`, but ok I can change it to make it that way with removing `active`. :)

packages/components/progress-bar/src/progress-bar.ts
anna-lach: I've changed it to `variant` :)

packages/components/progress-bar/src/progress-bar.ts
anna-lach: Changed :)

.changeset/yellow-cars-shout.md
```diff
@@ -0,0 +1,16 @@
...
+'@sl-design-system/bingel-dc': minor
+'@sl-design-system/clickedu': minor
+'@sl-design-system/magister': minor
+'@sl-design-system/kampus': minor
+'@sl-design-system/neon': minor
+'@sl-design-system/teas': minor
+'@sl-design-system/max': minor
+---
+
+Added progress bar tokens
```
jpzwarte, line 16: Please make this `patch`

jpzwarte (approved):

PR #1389, opened 2024-07-03
Fix styling issue where certain characters were cutoff due to not enough line-height

### LINKED ISSUE #1371: Certain characters are cutoff in avatar (like y and g)
CleanShot 2024-07-03 at 14 35 57@2x

DanielleRameau:
Fixes #1371


.changeset/sweet-carrots-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sl-design-system/avatar': patch
+---
+
+Due to #1371, the subtitle is no longer shown for the `md` size
```
anna-lach, line 5: Are we sure that this will not cause a breaking change for people, who are already using subtitle in `md` size?

anna-lach (approved):

.changeset/sweet-carrots-protect.md
jpzwarte: I will highlight this change in Slack to be sure, but i don't consider this breaking enough for a major.

PR #1392, opened 2024-07-04
Fix inline size behavior of dialog

jpzwarte:
- The max inline size now takes into account the margins on either side
CleanShot 2024-07-04 at 14 13 17@2x

- The min inline size on mobile is `100dvw`
CleanShot 2024-07-04 at 14 13 45@2x


packages/components/dialog/src/dialog.scss
```diff
@@ -38,6 +41,7 @@ dialog {
     100vh - var(--_margin-top-container) - var(--_margin-bottom-container) - 2 * var(--_padding-block)
   );
   max-inline-size: var(--_max-inline-size);
+  min-inline-size: 100dvw;
```
anna-lach, line 44: I think that we wanted to have a different inline-size before? :) 
https://www.figma.com/design/1oIVcD9VMbOtLXYp48ehsA/Sanoma-Learning-Light-mode?node-id=5303-20817&t=5rOdHrMaGkLLoMTe-4

@arecuenco-dsgn do we want to change it this way?

But maybe this is sth already discussed :)

jpzwarte: There is a bug in any case here on mobile:
CleanShot 2024-07-04 at 15 58 18@2x

The `100dvw` change fixes that.

anna-lach (approved):

packages/components/message-dialog/src/message-dialog.scss
```diff
@@ -1,7 +1,49 @@
...
+        )
+    );
+
+    margin: auto;
+    min-inline-size: min(var(--sl-size-dialog-message-mobile-min), var(--_max-inline-size));
+  }
+}
+
+sl-dialog::part(titles) {
+  @media (max-width: $md-breakpoint) {
```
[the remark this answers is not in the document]
jpzwarte: Done!

Diaan (approved):

PR #1410, opened 2024-07-12
New button-group component

jpzwarte:


packages/components/button-group/src/button-group.ts
```diff
@@ -0,0 +1,58 @@
...
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'sl-button-group': ButtonGroup;
+  }
+}
+
+type ButtonLike = HTMLElement & {
+  disabled: boolean;
+  fill?: ButtonFill;
```
anna-lach, line 14: Will it be applicable to a toggle button as well?

anna-lach (approved):

packages/components/button-group/src/button-group.ts
jpzwarte: Yes, toggle-button has that property/value as well. If if somebody slots something that doesn't have that, it's basically just a NOOP.

PR #1412, opened 2024-07-12
Fix/1327 cleaning up unused themes

Diaan:


anna-lach (approved):

jpzwarte (approved):
Perhaps safe to set "private": true in the package.json files of myvanin & utbildning? So they don’t accidentally get published?

PR #1416, opened 2024-07-15
New tool-bar component

jpzwarte:


Diaan:
The ... button appears just a tiny bit too late when you resize: 
image
Maybe related, in the overflow story it looks a bit weird as well: 
image

You added paddings last minute, do you take these into account in the calculations as well?

Diaan:
Also, in the align-end story the last button disappears but the ... button doesn't show: 
image

packages/components/tool-bar/src/tool-bar.ts
```diff
@@ -0,0 +1,229 @@
...
+  override render(): TemplateResult {
+    return html`
+      <div part="wrapper">
+        <slot @slotchange=${this.#onSlotChange}></slot>
+      </div>
+
+      ${this.menuItems.length
+        ? html`
+            <sl-menu-button>
+              <sl-icon name="far-ellipsis-vertical" slot="button"></sl-icon>
```
anna-lach, line 135: `aria-label` is missing here? :)

packages/components/tool-bar/src/tool-bar.ts
```diff
@@ -0,0 +1,229 @@
...
+          ${item.buttons.map(button => this.renderMenuItem(button))}
+        </sl-menu-item-group>
+        <hr />
+      `;
+    } else if (item.type === 'divider') {
+      return html`<hr />`;
+    } else {
+      return html`
+        <sl-menu-item @click=${() => item.click?.()} ?disabled=${item.disabled} ?selectable=${item.selectable}>
+          ${item.icon ? html`<sl-icon .name=${item.icon}></sl-icon>` : nothing} ${item.label}
```
anna-lach, line 156: `aria-label` is missing here? :)

packages/components/tool-bar/src/tool-bar.ts
jpzwarte: The aria-label should probably on the actual `<sl-button>`, but afaics, there is no way to set that atm. Something to fix.

packages/components/tool-bar/src/tool-bar.ts
jpzwarte: How do you mean? If the menu item has text, it doesn't need `aria-label`, right?

packages/components/tool-bar/src/tool-bar.ts
```diff
@@ -0,0 +1,229 @@
...
+export interface ToolBarItemGroup extends ToolBarItemBase {
+  type: 'group';
+  buttons: ToolBarItemButton[];
+  label?: string | null;
+  selects?: 'single' | 'multiple';
+}
+
+export type ToolBarItem = ToolBarItemButton | ToolBarItemDivider | ToolBarItemGroup;
+
+Icon.register(faEllipsisVertical);
```
Diaan, line 47: didn't we agree that all the icons we use in our own components will be added in the default icon set, so they are always available?
(This is done by adding them in the core.json)

packages/components/tool-bar/src/tool-bar.ts
Diaan: Is it necessary, since there is a label behind the icon?

Diaan (approved):

packages/components/tool-bar/src/tool-bar.ts
jpzwarte: I specifically did not add it yet, since it's still a draft version and there is no UI design yet.

packages/components/tool-bar/src/tool-bar.ts
jpzwarte: I've added a `FIXME` comment so we remember to do this

packages/components/tool-bar/src/tool-bar.ts
anna-lach: right, I didn't notice the label 😆

PR #1428, opened 2024-07-30
fix for select popover being too small

Diaan:


packages/components/shared/src/popover.ts
```diff
@@ -51,24 +51,30 @@ export const positionPopover = (
...
-        apply: ({ availableWidth, availableHeight }) => {
+        apply: ({ availableWidth, availableHeight, elements }) => {
           // Make sure that the overlay is contained by the visible page.
-          const maxBlockSize =
-              Math.max(MIN_OVERLAY_HEIGHT, Math.floor(availableHeight)) - (options.viewportMargin ?? 0),
-            maxInlineSize = options.maxWidth ?? Math.floor(availableWidth);
-
           const style = getComputedStyle(element),
-            currentMaxBlockSize = parseInt(style.maxBlockSize) ?? 0,
+            max = style.getPropertyValue('--sl-popover-max-block-size'),
```
anna-lach, line 59: Can you please add a changeset? :)

packages/components/shared/src/popover.ts
```diff
@@ -51,24 +51,30 @@ export const positionPopover = (
...
+            currentMaxBlockSize = !isNaN(parseInt(max)) ? parseInt(max) : 0,
+            min = style.getPropertyValue('--sl-popover-min-block-size'),
```
anna-lach, line 61: It looks like the tooltip is currently too high

image

anna-lach (approved):

packages/components/shared/src/popover.ts
```diff
@@ -1,5 +1,5 @@
 import { arrow, flip, offset, shift, size } from '@floating-ui/core';
-import { autoUpdate, computePosition } from '@floating-ui/dom';
+import { Elements, autoUpdate, computePosition } from '@floating-ui/dom';
```
jpzwarte, line 2: `Elements` should be a `type` import?

jpzwarte (approved):

PR #1448, opened 2024-08-14
Do not show static properties in the docs

### LINKED ISSUE #1440: Remove inaccessible API props
### Provide a general summary of the documentation request here

Please remove inaccessible props. Here is an example - static variable: 
image

### 🧢 Your Product/Team

ETSL

jpzwarte:
Fixed #1440

### DIFF
--- a/website/src/categories/components/component-table.njk
+++ b/website/src/categories/components/component-table.njk
@@ -35,9 +35,11 @@
                   </thead>
                   <tbody>
                   {%- for member in declaration.members -%}
-                    {%- if member.kind == 'field' and not member.static -%}
+                    {%- if member.kind == 'field' -%}
                       <tr>
-                      <td><code>{{member.name | escape}}</code></td>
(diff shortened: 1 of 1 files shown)


website/src/categories/components/component-table.njk
```diff
@@ -35,7 +35,7 @@
                   </thead>
                   <tbody>
                   {%- for member in declaration.members -%}
-                    {%- if member.kind == 'field' -%}
+                    {%- if member.kind == 'field' and not member.static -%}
```
anna-lach, line 38: I'm not sure it's always right? For example in the breadcrumbs we want to show that part:
image

Maybe we need to take a look at components and when the static should not be really shown we need to add `@ignore` there? :)

Diaan: those properties are still visible on the website. Right now they are displayed twice: 
image

PR #1454, opened 2024-08-15
New toggle group component

### LINKED ISSUE #1162: [Toggle Group] Development
```[tasklist]
### Tasks
- [ ] dont highlight the label only isolation, only in cobination with the toggle group
- [ ] update naming from button group to toggle group
- [ ] multi-select (option) and single select (default) property
```

image

jpzwarte:
Fixes #1162

This includes the changes from the #1161 branch


packages/components/toggle-button/src/toggle-button.stories.ts
```diff
@@ -1,66 +1,86 @@
...
+      options: ['md', 'lg']
     }
   },
-  render: ({ fill, size, pressed, label, disabled }) => {
-    const onToggle = (event: Event & { target: ToggleButton }): void => {
-      console.log(event.target.pressed);
-    };
-
+  render: ({ disabled, fill, icons, label, pressed, size }) => {
     return html`
```
anna-lach, line 41: Do we want to show an example of `toggle button` with icon and text next to it?

packages/components/toggle-group/src/toggle-group.scss
```diff
@@ -0,0 +1,59 @@
...
+  --_border-color: var(--sl-color-button-default-outline-idle-border);
+  --_border-radius: var(--sl-border-radius-button-md);
+  --_border-width: var(--sl-border-width-button-outline);
+  --_paddingInline: 1.25rem 1.5rem;
+  --_paddingInline-textOnly-pressed: calc(1.5rem - var(--sl-space-button-gap-md)) 1.5rem;
+  --_paddingInline-textOnly: 2rem;
+  --_separator-color: var(--sl-color-button-default-outline-idle-border);
+  --_separator-width: var(--sl-border-width-button-outline);
+
+  border: 1px solid var(--_border-color);
```
anna-lach, line 11: For the border width there needs to be a token, right?

packages/components/toggle-group/src/toggle-group.scss
```diff
@@ -0,0 +1,59 @@
+:host {
+  --_border-color: var(--sl-color-button-default-outline-idle-border);
+  --_border-radius: var(--sl-border-radius-button-md);
+  --_border-width: var(--sl-border-width-button-outline);
+  --_paddingInline: 1.25rem 1.5rem;
```
anna-lach, line 5: We are not using tokens for the padding?

packages/components/toggle-button/src/toggle-button.ts
```diff
@@ -11,23 +13,31 @@ declare global {
...
+ *    <sl-icon name="far-gear" slot="default"></sl-icon>
+ *    <sl-icon name="fas-gear" slot="pressed"></sl-icon>
+ *  </sl-toggle-button>
  * ```
  *
  * @slot default - The icon shown in the default state of the button
  * @slot pressed - The icon shown in the pressed state of the button
  */
-export class ToggleButton extends LitElement {
+export class ToggleButton extends ScopedElementsMixin(LitElement) {
```
anna-lach, line 32: It looks like `toggle-button` and `toggle-group` is missing in the `tsconfig.all.json` file?

packages/components/toggle-group/src/toggle-group.ts
```diff
@@ -0,0 +1,80 @@
...
+   * handle the toggling of buttons yourself.
+   */
+  @property({ type: Boolean }) multiple?: boolean;
+
+  /** Determines the size of all buttons in the group. */
+  @property({ reflect: true }) size?: ToggleButtonSize;
+
+  override updated(changes: PropertyValues<this>): void {
+    super.updated(changes);
+
```
anna-lach, line 49: Maybe `role="group"` needs to be applied?

Diaan:
Why are all these separate stories in storybook? Chromatic takes a snapshot for every story. Adding so many stories for such a simple component feels quite wasteful in terms of our snapshots.
image

Diaan:
There are changes in chromatic: https://www.chromatic.com/test?appId=646dd42469efabe23410bb59&id=66bdf5b46100d7b0d365002b

packages/components/toggle-button/src/toggle-button.stories.ts
jpzwarte: We do not. The users are meant to use that functionality outside of the toggle-group. This was discussed with Danielle.

packages/components/toggle-group/src/toggle-group.ts
jpzwarte: I looked at https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role, but that doesn't really match with the toggle group?

packages/components/toggle-group/src/toggle-group.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (changes.has('disabled') || changes.has('size')) {
+      this.#updateButtonProperties();
+    }
+  }
+
```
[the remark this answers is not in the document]
jpzwarte: Done

packages/components/toggle-group/src/toggle-group.ts
jpzwarte: `"region"` is apparently the right one: https://lea.verou.me/blog/2022/07/button-group/ && https://twitter.com/LeonieWatson/status/1545788775644667904

packages/components/toggle-button/src/toggle-button.stories.ts
```diff
@@ -85,93 +105,98 @@ export const All: Story = {
...
+        #root-inner {
+          align-items: center;
+          align-self: start;
+          display: inline-grid;
+          gap: 1rem 2rem;
+          grid-template-columns: auto 1fr 1fr 1fr 1fr;
+          justify-items: center;
         }
+      </style>
+      <span>Ghost</span>
```
Diaan, line 117: why not use a table for this?

packages/components/toggle-button/src/toggle-button.ts
```diff
@@ -11,23 +13,31 @@ declare global {
   }
 }
 
-export type ToggleButtonSize = 'sm' | 'md' | 'lg';
 export type ToggleButtonFill = 'ghost' | 'outline';
+export type ToggleButtonSize = 'md' | 'lg';
 
 /**
- * Let the user toggle between two states.
+ * Let's the user toggle between two states.
```
Diaan, line 20:
```suggestion
 * Lets the user toggle between two states.
```

packages/components/toggle-button/src/toggle-button.ts
```diff
@@ -37,92 +47,157 @@ export class ToggleButton extends LitElement {
...
 
-  /** Whether the toggle-button is disabled; when set no interaction is possible. */
+  /** Whether the button is disabled; when set no interaction is possible. */
   @property({ type: Boolean, reflect: true }) disabled?: boolean;
 
   /** The variant of the toggle-button. */
   @property({ reflect: true }) fill?: ToggleButtonFill;
 
-  /** The state of the toggle-button. */
+  /** @internal True when the user has slotted text in the  */
```
Diaan, line 62:
```suggestion
  /** @internal True when the user has slotted text in the button  */
```

packages/components/toggle-group/src/toggle-group.ts
```diff
@@ -0,0 +1,102 @@
...
+import styles from './toggle-group.scss.js';
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'sl-toggle-group': ToggleGroup;
+  }
+}
+
+/**
+ * A component for visually grouping toggle buttons together.
```
Diaan, line 15: maybe also add something about the single/multiple functionality? It's more than just visual

packages/components/toggle-group/src/toggle-group.scss
jpzwarte: Fixed

packages/components/toggle-button/src/toggle-button.stories.ts
jpzwarte: Grid is less markup than `<table>`. Also allows for more flexibility (see the "All" story of toggle-group)

jpzwarte: But in this case, perhaps the difference is less clear

Diaan (approved):

PR #1455, opened 2024-08-15
Improve static property documentation

jpzwarte:
- Rename jsdoc `@private` to `@internal`
- Mark `offset`/`viewportMargin` etc. static properties as `@internal`


Diaan:
Can we make this a bit clearer? 
image
maybe a separate tag or some sort of styling?

anna-lach (approved):

jpzwarte:
> Can we make this a bit clearer? image maybe a separate tag or some sort of styling?

I already discussed this in Slack. This was the outcome :)

PR #1457, opened 2024-08-19
Various tag improvements

jpzwarte:
`<sl-tag>`:
- Simplify the HTML structure
- Slot the text instead of using a property
- Use a `ResizeObserver` for determining whether the text overflows
- Rework the CSS so we don't need JS for the hover styling
- Use a lazy tooltip
- Remove `readonly` property for now due to uncertainty wrt focusability (https://adrianroselli.com/2022/11/brief-note-on-aria-readonly-support-html.html)

`<sl-tag-list>`:
- Remove `MutationObserver` and watch the tags using the `ResizeObserver` instead
- Simplify styling by setting a class on the `stack`

`Tooltip.lazy`: returns a cleanup function that can be used to clean up, but also to prevent multiple lazy tooltips from being initialized.


packages/components/tag/src/tag.stories.ts
```diff
@@ -1,292 +1,121 @@
...
-        ></sl-tag>
-        ${amount.map(
-          (_, index) => html`
-            <sl-tag .label=${label + ' ' + (index + 1)} ?removable=${removable} ?readonly=${readonly}></sl-tag>
-          `
-        )}
-      </sl-tag-list>
-    `;
+    label: 'This is a very long label which overflows',
+    maxWidth: '200px'
```
anna-lach, line 62: When I change the maxWidth in the story, even when there is ellipsis visible, the tooltip is no longer showing on hover/focus

image

packages/components/tag/src/tag.scss
```diff
@@ -1,263 +1,134 @@
...
-
-  .remove-button {
-    background: var(--sl-color-tag-subtle-hover-background);
-
-    --_button-border: var(--sl-color-tag-subtle-hover-border);
-  }
+slot:hover,
+slot:hover + button,
+button:hover {
+  --_background: var(--sl-color-tag-subtle-hover-background);
```
anna-lach, line 28: When the bold tag is focused and I hover it (the tag, not the button in it), the background of the button changes, but it should not when I hover over the tag, not the button. Right now it looks like this:

image

packages/components/tag/src/tag-list.scss
```diff
@@ -1,89 +1,76 @@
...
-    }
+  &:has(sl-tag[emphasis='bold']:focus) {
+    --_border-color: var(--sl-color-tag-subtle-active-border);
   }
 }
 
+.double::before,
+.triple::before,
+.triple::after {
+  border-color: var(--_border-color);
```
anna-lach, line 56: It looks like it's not alligned with the tag list?
image
It sticks out from the left side?

packages/components/tag/src/tag-list.scss
```diff
@@ -1,89 +1,76 @@
...
+  border-inline-start: 2px solid var(--_border-color);
```
anna-lach, line 57: Why we are not using tokens here? :)

packages/components/tag/src/tag-list.scss
```diff
@@ -1,89 +1,76 @@
...
+  border-radius: 4px;
```
anna-lach, line 58: Can you please use tokens here? :)

packages/components/tag/src/tag-list.scss
```diff
@@ -1,89 +1,76 @@
...
+  content: '';
+  display: flex;
+  inline-size: 4px;
```
anna-lach, line 61: Can you please use tokens here? :)

packages/components/tag/src/tag-list.scss
anna-lach: We could have problems with shape in other themes when we will leave the exact value here, not a token?

packages/components/tag/src/tag-list.ts
```diff
@@ -47,175 +41,161 @@ export class TagList extends ScopedElementsMixin(LitElement) {
...
-      this.tags?.forEach(tag => (tag.emphasis = this.emphasis));
+    if (changes.has('stacked') && !this.stacked) {
+      this.tags.forEach(tag => (tag.style.display = ''));
+    }
+
+    if (changes.has('stacked') || changes.has('stackSize') || changes.has('tags')) {
+      if (this.stacked) {
+        const total = this.tags.length;
+
+        this.setAttribute('aria-label', `${msg(str`Showing ${total - this.stackSize} out of ${total} tags`)}`);
```
anna-lach, line 118: So we are leaving `tags` for now, yes?

packages/components/tag/src/tag.scss
```diff
@@ -1,263 +1,134 @@
...
-  border-inline-end: none;
-  border-inline-start: 1px solid var(--_button-border);
-  border-start-end-radius: var(--_border-radius);
-  cursor: pointer;
-  display: flex;
-  inline-size: var(--sl-size-tag-md);
+  aspect-ratio: 1;
+  background: var(--_background);
+  border: 0;
+  border-inline-start: 1px solid var(--_border-color);
```
anna-lach, line 121: Can you please use a token for the border width?

packages/components/tag/src/tag.stories.ts
jpzwarte: I cannot reproduce that at this time.

anna-lach: I think when I changed the `maxWidth` few times there in options it happened.

packages/components/tag/src/tag-list.scss
jpzwarte: Fixed

packages/components/tag/src/tag.ts
```diff
@@ -58,128 +65,105 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
+  }
 
-    this.#checkOverflow();
+  override render(): TemplateResult {
+    return html`
+      <div class="wrapper" .tabIndex=${this.disabled ? -1 : 0}>
+        <slot @slotchange=${this.#onSlotChange}></slot>
+        ${this.removable
+          ? html`
+              <button @click=${this.#onRemove} aria-label=${msg(`Remove ${this.label}`)} ?disabled=${this.disabled}>
```
anna-lach, line 119: It looks like the `this.label` is not being rendered inside the `aria-label`?

image

packages/components/tag/src/tag.ts
```diff
@@ -58,128 +65,105 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
+                <sl-icon name="xmark" size=${ifDefined(this.size)}></sl-icon>
```
anna-lach, line 120: I've just noticed that for `xmark` icon in Figma there is a `xs` size of icon used for `md` and `lg` size of tag,
 so 12x12 px:

image

anna-lach (approved):

Diaan (approved):

PR #1458, opened 2024-08-19
updated TEAS focus ring

DanielleRameau:
Increased the focus ring thickness in TEAS from 1px to 2px.


jpzwarte (approved):
Please add a changeset for this

PR #1469, opened 2024-08-22
updated color default ghost button in TEAS

DanielleRameau:


.changeset/loud-meals-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sl-design-system/tokens": patch
```
anna-lach, line 2: The fix is connected with a `teas` package so here instead of `@sl-design-system/tokens`
we should have `@sl-design-system/teas` :)

anna-lach (approved):

PR #1483, opened 2024-08-27
Fix/1366 slow fa icon load

Diaan:


Diaan:
When reviewing please also check https://github.com/sl-design-system/angular-demo/pull/7

jpzwarte (approved):

PR #1488, opened 2024-08-29
Fix/1480 padding border radius issues

Diaan:


Diaan:
Made a PR so i could check chromatic, but Chromatic doesn't take new snapshots... 🤔

packages/components/toggle-button/src/toggle-button.scss
```diff
@@ -1,7 +1,7 @@
 :host {
```
anna-lach, line 1: Can you please add changeset? :)

DanielleRameau (changes requested):
- [x] The buttons in the large toggle group shift/jump when one of them is selected
- [x] Tooltip margin and border-radius is different in toggle group compared with the regular tooltip
- [ ] The end button in text pressed variant has a odd white line on the right side (itslearning, magister, bingel DC)
- [x] Remove border ‘disabled pressed’, token should be the same as the background color (all themes except neon light and 

dark)

Diaan:
> The end button in text pressed variant has a odd white line on the right side (itslearning, magister, bingel DC)
Is nearly impossible to solve; this is a Safari issue that occurs when 4 properties are combined:

- `sl-toggle-group` needs to be places in a specific place in a grid
- overflow: hidden on the `sl-toggle-group`
- position:relative on the `sl-toggle-button`
- `border-radius` set to something other than 0.

Because of the first in this list the chance that this will occur in a CFA is pretty small, so we decided to not put more effort into this.

https://codepen.io/dianabroeders/pen/KKjrgao

anna-lach (approved):

DanielleRameau (approved):

PR #1510, opened 2024-09-11
Improve naming of contextual CSS custom properties

jpzwarte:
As discussed, for contextual tokens we want to use the following naming convention: `--sl-[a-z][a-zA-Z]*(-[a-z][a-zA-Z]*)*`

This PR changes the `build-themes.js` to apply this transformation to the following files:
- primitives.json
- system.json
- *-new.json

All the other JSON files still use the old everyting-is-kebab-case convention.

### DIFF
--- a/package.json
+++ b/package.json
@@ -467,7 +467,7 @@
     "storybook": "^8.2.9",
     "stylelint": "^16.9.0",
     "typescript": "^5.4.5",
-    "vite": "^5.4.3",
+    "vite": "^5.4.4",
     "wireit": "^0.14.9"
   }
 }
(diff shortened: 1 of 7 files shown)


.changeset/clever-vans-help.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sl-design-system/menu': patch
+'@sl-design-system/sanoma-learning': patch
+---
+
+Do not transform token parts to kebab case, only add '-' in between
```
anna-lach, line 6: Maybe you can also add some information that it's added due to contextual tokens usage? :)

anna-lach (approved):
Maybe @Diaan can also take w look? :)

Diaan (approved):

PR #1524, opened 2024-09-17
Feature/1163 toggle group make available in all themes

DanielleRameau:
- added toggle group in all themes
- fixed color issues in toggle button and toggle group in all themes


jpzwarte (commented):
Can you a changeset plz?

jpzwarte (approved):

PR #1541, opened 2024-09-23
Add aria-label to icon only menu button stories

### LINKED ISSUE #1511: [menu-button] Missing `aria-label` attributes on icon only stories

See title

### 🤔 Expected Behavior?

-

### 😯 Current Behavior

-

### 💁 Possible Solution

- Add a `label` property to `<sl-menu-button>` and apply that as `aria-label` on the nested button
- Log a warning when the button doesn't have a label or aria-label property

### 🖥️ Steps to Reproduce

-

Chrome

### If other, please specify.

_No response_

macOS

### 👤 Your name

Jeroen Zwartepoorte

### 🧢 Your product/team

SLDS

### 🎨 Your Theme(s)

Sanoma Learning

jpzwarte:
Fixes #1511

### DIFF
--- a/packages/components/menu/src/menu-button.ts
+++ b/packages/components/menu/src/menu-button.ts
@@ -51,6 +51,9 @@ export class MenuButton extends ScopedElementsMixin(LitElement) {
   /** The fill of the button. */
   @property() fill: ButtonFill = 'outline';
 
+  /** If the menu button is icon only, use this to set the `aria-label` on the button. */
+  @property() label?: string;
+
   /** @internal The menu. */
   @query('sl-menu') menu!: Menu;
 
@@ -80,6 +83,7 @@ export class MenuButton extends ScopedElementsMixin(LitElement) {
         @click=${this.#onClick}
         @keydown=${this.#onKeydown}
         ?disabled=${this.disabled}
+        aria-label=${ifDefined(this.label)}
         fill=${ifDefined(this.fill)}
         part="button"
         size=${ifDefined(this.size)}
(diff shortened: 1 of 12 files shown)


packages/components/menu/src/menu-button.ts
```diff
@@ -51,6 +51,9 @@ export class MenuButton extends ScopedElementsMixin(LitElement) {
   /** The fill of the button. */
   @property() fill: ButtonFill = 'outline';
 
+  /** If the menu button is icon only, use this to set the `aria-label` on the button. */
+  @property() label?: string;
```
anna-lach, line 55: Are we sure that this is really necessary? Adding `aria-label` on the `menu-button` is enough for the Voice Over, like in this example:
https://proud-water-0d02ee203-1541.westeurope.2.azurestaticapps.net/categories/components/menu/usage/

Maybe this PR should only cover adding `aria-label` in the stories  of menu button - icon only?

anna-lach: If we really want to keep it this way, we also need to change the documentation about `aria-label` https://proud-water-0d02ee203-1541.westeurope.2.azurestaticapps.net/categories/components/menu/accessibility/ in the menu button.

Diaan: Agreed; if we force the use of `aria-label` we sort of remove the option of using `aria-labeled-by` when that is more appropriate.
If a dev doesn't think about using `aria-label` they probably also won't think to use `label`

jpzwarte: Changed the PR to now only add an `aria-label` to the icon only stories

PR #1550, opened 2024-09-25
Feature/1507 grid apply tokens

Diaan:


packages/components/grid/src/grid.scss
```diff
@@ -208,6 +229,20 @@ td[part~='text-field'] {
   font-weight: 600;
 }
 
+:host([clickable-row]) {
+  [part~='row'] {
+    cursor: pointer;
+
+    &:hover {
```
jpzwarte, line 236: So when i click a row, there is no `:active` styling (unlike in a button for example). Shouldn't we add that here as well?

packages/components/shared/src/controllers/selection.ts
```diff
@@ -20,6 +20,9 @@ export class SelectionController<T = any> {
    */
   #selection = new Set<T>();
 
+  //** The currently highlighted row */
```
jpzwarte, line 23: Typo in the comment. Also could you make the comment more generic? The idea behind the `SelectionController` is that it is not grid specific.

jpzwarte (approved):

Issue #1526, opened 2024-09-18
[Data grid] Add keyboard navigation


DanielleRameau:
(see: https://www.w3.org/WAI/ARIA/apg/patterns/grid/#keyboardinteraction-settingfocusandnavigatinginsidecells)

### Patterns and documentation:
**Pagination**
When using pagination it is a good idea to let the user know there is pagination and on which page they are before "entering" the table. If a teacher opens the page with a table of all students in a class, of which they know there are 30, it might be confusing to start reading the table and have the table say it's "row 1 of 11". If the`sl-paginator-status` is placed above the table a screen reader, but also a low vision user using a magnifier will know they are not getting the full table.
compare:
![Image](https://github.com/user-attachments/assets/f19f7ef9-c1e4-4745-9358-79bc6a5425af)
With 
![Image](https://github.com/user-attachments/assets/9ff8e8ce-7366-4340-8e79-910017433a10)
Because of shadow dom boundaries i couldn't get it to work with `aria-described by` and point to the `sl-paginator-status` 

**Size and tabstops**
Be clear there is a limit on the amount of rows and tab stops per row. 300 rows, with a text field and 3 buttons means 1200 tabstops, it is better make selectable rows and make the actions bulk actions or use pagination.

```[tasklist]
### Tasks
- [ ] styling of skip links (Roald)
```

Diaan:
Some interesting examples: https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/data-grids/

Diaan:
- [ ] Set tab-index of all elements that are only child of cell to -1 and focus that element when the cell is active

Diaan:
when navigating:
arrow keys: move around
enter -> go in edit mode

when in edit mode:
enter-> confirm and go to cell below, already in edit mode
esc -> exit edit mode and focus current cell
arrow keys -> like in "normal, single" input field/select

Diaan:
![Image](https://github.com/user-attachments/assets/34e742e1-bb12-424c-975a-2c4e2716f817)
https://inclusive-components.design/data-tables/

https://adrianroselli.com/2020/02/block-links-cards-clickable-regions-etc.html#Update02

Diaan:
research screenreader usage virtual list, both in role table and role grid

table
only data, no edit fields or selection column
clickable row
sorting
filtering

grid
interactive; edit fields, selection columns
clickable row
sorting
filtering
selection (multi select row)
all cells focusable by keyboard?

[...]

PR #1567, opened 2024-10-02
Add `ellipsizeText` property to grid and column for automatically ellipsizing long text

### LINKED ISSUE #1436: [Grid] Truncation of content in column

As just discussed with Jeroen, we would like to have the option to decide for individual columns in a table grid how long content should be dealt with. We want to be able to choose between truncation ('...') or multiline text.

### 🤔 Expected Behavior?

Truncated text on specific columns in a table grid. (For when the information in that column is not relevant enough to display in its entirety or if the information can be seen in a 'detailkaart') When the text is truncated, the full text can be viewed through a tooltip.

### 😯 Current Behavior

Currently, it is only possible to let the text overflow to the next line, therefore increasing the lineheight of that row. We mostly want all rows to have the same lineheight.

### 💁 Possible Solution

Maybe some kind of variant/variable of a column where the content of that column gets truncated ('...'), instead of changing to multiline text, when the content gets too long for the cell width?

### 🔦 Context

We deal mostly with large overviews of i.e. a lot of students. If the lineheight differs between rows, it becomes a lot harder to find the information you're looking for as a user.

### 💻 Examples

Screenshot 2024-08-08 at 14 51 55

### 🧢 Your Product/Team

Magister
```[tasklist]
### Tasks
- [x] check best UX practices for such situations (Alvaro)
- [ ] implement right solution
```

jpzwarte:
Fixes #1436


.changeset/good-eggs-pay.md
```diff
@@ -0,0 +1,10 @@
+---
+'@sl-design-system/grid': patch
+---
+
+Add `ellipsizeText` property to grid and column
+
+When set on either `<sl-grid>` or `<sl-grid-column>` (or any of their variants), the `ellipsizeText` property
+will render the table data using the `<sl-ellipsize-text>` component, which truncates text with an ellipsis when it
+overflows its container. This is useful for tables with long text that would otherwise cause row height to grow.
+The component also automatically adds a tooltip to the truncated text so that the full text can be viewed on hover.
```
anna-lach, line 10: The tooltip should be available on focus as well, right? Not only on hover.

packages/components/ellipsize-text/src/ellipsize-text.ts
```diff
@@ -0,0 +1,78 @@
...
+    super.disconnectedCallback();
+  }
+
+  override render(): TemplateResult {
+    return html`<slot></slot>`;
+  }
+
+  #onResize(): void {
+    if (this.offsetWidth < this.scrollWidth) {
+      this.#tooltip ||= Tooltip.lazy(
```
anna-lach, line 59: Maybe the tooltip should be also available on focus, not only on hover?

anna-lach: So the whole element should be focusable, when there is ellipsis?

jpzwarte: The tooltip also listens for `focusin`, so that is already covered :) I think the focus part also needs to wait until we have #1526 so we can test it properly.

Diaan (approved):

anna-lach (approved):

PR #1575, opened 2024-10-04
Data source improvements

### LINKED ISSUE #140: [Development] Add grid example with data from back-end
Add example/code for a `DataSource` where the data is fetched from a remote server.

This is one of the main things missing for `<sl-grid>` to be usable by teams.

Requirements:
- Should be grid agnostic (API should also work for other web components)
- Should support paginating for fetching data
- Should support sort/filtering on the server as well

jpzwarte:
- Moved `DataSource` and `ArrayDataSource` into its own `@sl-design-system/data-source` package
- Added `FetchDataSource` to the data source package
- Removed the `filteredItems` property of `DataSource` (just use the `items` property instead)
- Fixed `ArrayDataSource` to handle multiple filters for the same `path` using `OR` instead of `AND`
- Changed `GridColumn` to render an `<sl-skeleton>` for items that equal `FetchDataSourcePlaceholder`

Fixes #140


Diaan (approved):

packages/components/data-source/src/fetch-data-source.ts
```diff
@@ -0,0 +1,157 @@
+import { DataSource, type DataSourceSort } from './data-source.js';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export interface FetchDataSourceCallbackOptions<T = any> {
```
anna-lach, line 4: Maybe `unknown` instead of `any` would be better?

packages/components/data-source/src/array-data-source.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    it('should not have grouping', () => {
+      expect(ds.groupBy).to.be.undefined;
+    });
+
+    it('should not have sorting', () => {
+      expect(ds.sort).to.be.undefined;
+    });
+
+    it('should emit an sl-update event when calling update()', () => {
```
anna-lach, line 38:
```suggestion
    it('should emit an sl-data-source-update event when calling update()', () => {
```

.changeset/lazy-rockets-itch.md
```diff
@@ -0,0 +1,15 @@
...
+
+New `@sl-design-system/data-source` package
+
+This packages provides `DataSource`, `ArrayDataSource` and `FetchDataSource` classes
+for managing data sources in the design system. At the moment, it is only used by the
+grid components, but it can be used in future components as well.
+
+`DataSource` and `ArrayDataSource` were previously part of the `@sl-design-system/shared`
+package, but they have been moved to this new package to make them more reusable.
+
```
anna-lach, line 13: Maybe there should be also an information that instead of using `sl-update` event, from now should be used `sl-data-source-update`?

anna-lach (approved):

packages/components/data-source/src/fetch-data-source.ts
jpzwarte: If you use unknown, then you will get warnings/errors in code that uses it, without a type (speaking from experience). So while `unknown` is more correct, `any` is more practical.

packages/components/data-source/src/array-data-source.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+      ds.addFilter('id', 'profession', 'Gastroenterologist');
+      ds.update();
+
+      expect(ds.items).to.have.length(2);
+
+      ds.removeFilter('id');
+      ds.update();
+
+      expect(ds.items).to.deep.equal(people);
+    });
```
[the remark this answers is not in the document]
jpzwarte: That's not how path filters work. They only match the exact value. If you want to do a partial patch, you should have a function filter.

packages/components/data-source/src/data-source.ts
```diff
@@ -56,9 +65,6 @@ export abstract class DataSource<T = any> extends EventTarget {
   }
 
   /** The filtered & sorted array of items. */
```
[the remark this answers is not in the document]
jpzwarte: It is.

.changeset/lazy-rockets-itch.md
jpzwarte: I reverted the name change as discussed.

PR #1582, opened 2024-10-09
Feature/1577 contextual token upgrade

DanielleRameau:


anna-lach:
Will be solved in https://github.com/sl-design-system/components/pull/1607

PR #1590, opened 2024-10-10
Update whats-new.md

DanielleRameau:


.changeset/nasty-owls-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sl-design-system/website": patch
```
anna-lach, line 2: For the website this changeset file is not necessary, we only need changeset for changes in the packages. This file can be removed. :)

anna-lach (approved):
After removing the changeset file, will be ready for merge :)

PR #1599, opened 2024-10-16
Various combobox fixes

jpzwarte:
# combobox
- The label was associated with the `<sl-combobox>` element instead of the `<input>` element
- Set `aria-selected="false"` on the non-selected options
- Set `aria-multiselectable="true"` on the listbox when the `multiple` property is set
- Add an `aria-label` to the `<sl-tag-list>` with value `"Selected options"`

# tag-list
- Use `aria-labelledby` for the tooltip instead of `aria-describedby`
- Do not set an `aria-label` on the host element; `role="list"` provides enough information

# tag
- Add ability to use Delete or Backspace keys to remove the tag
- Use `aria-description` to describe how to remove the tag using the keyboard
- Make the delete button `aria-hidden="true"`
- Remove `tabindex` from the `.wrapper` element in the shadow DOM

# tooltip
- Make tooltip also work if linked via `aria-labelledby`

# RovingTabindexController
- Fix behavior so `<sl-tag-list>` works properly when switching between light & shadow DOM

The above changes improve the UX when using AT tooling to navigate the combobox/tag-list.


packages/components/tag/src/tag-list.ts
```diff
@@ -113,24 +113,14 @@ export class TagList extends ScopedElementsMixin(LitElement) {
     if (changes.has('stacked') && !this.stacked) {
       this.tags.forEach(tag => (tag.style.display = ''));
     }
-
-    if (changes.has('stacked') || changes.has('stackSize') || changes.has('tags')) {
-      if (this.stacked) {
-        const total = this.tags.length;
-
-        this.setAttribute('aria-label', `${msg(str`Showing ${total - this.stackSize} out of ${total} items`)}`);
```
anna-lach, line 121: Why are we removing this part? This part was added before to improve the accessibility and inform the user how many items are currently visible in the stacked version.

packages/components/tag/src/tag.ts
```diff
@@ -104,27 +104,35 @@ export class Tag extends ScopedElementsMixin(LitElement) {
     if (changes.has('disabled')) {
       this.setAttribute('tabindex', this.disabled ? '-1' : '0');
     }
+
+    if (changes.has('removable')) {
+      if (this.removable) {
+        this.setAttribute('aria-description', msg('Press the delete or backspace key to remove this tag'));
```
anna-lach, line 110: Isn't it necessary to add the information here like it was before in the `aria-label` which tag (`this.label`) is going to be exactly removed?

packages/components/tag/src/tag.ts
```diff
@@ -104,27 +104,35 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
+      } else {
+        this.removeAttribute('aria-description');
+      }
+    }
   }
 
   override render(): TemplateResult {
     return html`
-      <div class="wrapper" .tabIndex=${this.disabled ? -1 : 0}>
```
anna-lach, line 111: So currently we cannot focus on click the whole tag as we wanted before, is it ok?

anna-lach: And I think we wanted before to remove the tag only when the remove button was focused (with keyboard), not when the whole tag gets focus? I think the case was also that we could make selectable tags in the future?

packages/components/tooltip/src/tooltip.ts
```diff
@@ -108,7 +108,11 @@ export class Tooltip extends LitElement {
   #events = new EventsController(this);
 
   #matchesAnchor = (element: Element): boolean => {
-    return !!this.id && element.nodeType === Node.ELEMENT_NODE && this.id === element.getAttribute('aria-describedby');
+    return (
+      !!this.id &&
+      element.nodeType === Node.ELEMENT_NODE &&
+      (this.id === element.getAttribute('aria-describedby') || this.id === element.getAttribute('aria-labelledby'))
```
anna-lach, line 114: Can you please update the `accessibility` part in the documentation of changed components? :)

packages/components/tag/src/tag-list.ts
jpzwarte: After testing with VoiceOver, the `role="list"` on the `<sl-tag-list>` already causes it to announce how many items are in the list. And with the `aria-labelledby` instead of `aria-describedby` on the stacked tag, it now also announces which tags are in the stack. 
And finally, this aria-label blocks any aria-labels set from the outside of the tag-list, like `aria-label="Selected options"` from the combobox.

packages/components/tag/src/tag.ts
jpzwarte: VoiceOver first reads the label of the tag and then this description, so i don't think that is necessary :)

anna-lach: ok :)

packages/components/tag/src/tag.ts
jpzwarte: As discussed in yesterday's meeting: have the ability to use the keyboard to delete a tag improves the UX for people using the keyboard/screen readers.

Diaan (approved):

PR #1612, opened 2024-10-25
Improve combobox performance

jpzwarte:
- Improve performance by also supporting `<option>` and `<optgroup>`
- Fix chevron button to actually toggle the popover
- Scroll to the selected option in single select mode when opening the popover
- For `<sl-option>`, only use the text directly in the element for rendering the selected option
- Split stories into combobox, single & multiple files
- Call the `setup()` method from the SL theme for the unit tests, so the icons are there


Diaan:
https://github.com/user-attachments/assets/67c29121-3ec4-4708-be09-9fb08b227adc

When typing the correct option is selected but this is not visible. Only when i use the arrow key to go the next one the option i typed in appears in view.

PR #1616, opened 2024-10-31
Feature/1526 grid keyboard navigation

Diaan:


packages/components/grid/src/column.ts
```diff
@@ -122,6 +122,8 @@ export class GridColumn<T = any> extends LitElement {
   /** The position where the column should be sticky: at the start of the grid, or at the end. */
   @state() stickyPosition?: 'start' | 'end';
 
+  @state() selected?: boolean;
```
anna-lach, line 125: Can you please add short description?

packages/components/grid/src/stories/basics.stories.ts
```diff
@@ -99,6 +99,53 @@ export const EllipsizeTextSingleColumn: Story = {
...
+          grow="0"
+          width="64"
+          .scopedElements=${{
+            'sl-icon': Icon,
+            'sl-menu-button': MenuButton,
+            'sl-menu-item': MenuItem
+          }}
+        ></sl-grid-column>
+      </sl-grid>
+      <p>A paragraph that follows the table, with a <a href="http://home.com">link</a> in it.</p>
```
anna-lach, line 144: Maybe just sth like `href="#"`? :)

packages/components/shared/src/controllers/focus-group.ts
```diff
@@ -262,21 +288,27 @@ export class FocusGroupController<T extends HTMLElement> implements ReactiveCont
         break;
     }
     event.preventDefault();
+
     if (this.direction === 'grid' && this.currentIndex + diff < 0) {
-      this.currentIndex = 0;
+      // this.currentIndex = 0;
```
anna-lach, line 293: Do we want to keep those commented lines?

packages/components/shared/src/controllers/focus-group.ts
```diff
@@ -14,6 +16,7 @@ export type FocusGroupConfig<T> = {
 export class FocusGroupController<T extends HTMLElement> implements ReactiveController {
```
anna-lach, line 16: Can you please add changeset? :)

anna-lach (approved):

packages/components/grid/src/grid.scss
```diff
@@ -350,6 +361,47 @@ td[part~='text-field'] {
...
+}
+
+.skip-link-start:focus {
+  transform: translateY(calc(-1 * (1lh + 2 * var(--sl-space-new-md))));
+}
+
+.skip-link-end {
+  border-radius: 0 0 var(--sl-size-borderRadius-default) var(--sl-size-borderRadius-default);
+  inset-block-end: 0;
+  transform: translateY(0);
```
jpzwarte, line 398: etc.

packages/components/grid/src/grid.ts
```diff
@@ -440,6 +456,7 @@ export class Grid<T = any> extends ScopedElementsMixin(LitElement) {
         class=${classMap({ active })}
         part=${parts.join(' ')}
         index=${index}
+        aria-rowindex=${index}
```
jpzwarte, line 459: Should grid also include `aria-posinset` and `aria-setsize` next to `aria-rowindex`? I’ve added the first 2 to combobox. See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-posinset

jpzwarte (approved):

packages/components/grid/src/grid.ts
jpzwarte: Never mind the above, but do plz add https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-rowcount

jpzwarte: Which is already in the DOM :)

packages/components/grid/src/grid.ts
```diff
@@ -335,6 +343,14 @@ export class Grid<T = any> extends ScopedElementsMixin(LitElement) {
             `
           : nothing}
       </table>
+
+      <a
+        id="table-end"
+        href="#table-start"
+        class="skip-link-end"
+        @click=${(e: Event & { target: HTMLSlotElement }) => this.#onSkipTo(e, 'start')}
```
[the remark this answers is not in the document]
Diaan: no, hash links are messed with when using it in angular. I couldn't get it to work...

packages/components/grid/src/grid.ts
Diaan: when the role is table or grid instead of list you need to use `aria-rowcount` and `aria-rowindex`

packages/components/grid/src/column.ts
Diaan: i didn't need that state apparently :D

packages/components/grid/src/grid.ts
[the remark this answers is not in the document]
jpzwarte: So please add a comment in the event handler so we remember the reason :)

PR #1624, opened 2024-11-06
Various fixes

jpzwarte:
- Fix `ShadowRoot.createElement` type to match `document.createElement`
- Fix `getValueByPath` and `setValueByPath` to properly infer types
- Remove path utils from form package and use the shared functions


packages/components/shared/src/path.spec.ts
```diff
@@ -16,12 +16,122 @@ describe('path utils', () => {
   describe('getStringByPath', () => {
     const person = { address: { street: 'Main Street' } };
 
-    it('should fail gracefully when no path is set', () => {
-      expect(getStringByPath(person, '')).to.equal('');
+    it('should return the string representation of person', () => {
```
Diaan, line 19:
```suggestion
    it('should return the string representation of the object', () => {
```

Diaan (approved):

PR #1626, opened 2024-11-07
Add virtual list feature to listbox

jpzwarte:


packages/components/listbox/src/option.ts
```diff
@@ -15,7 +15,8 @@ declare global {
  *
  * @slot default - The option's label.
  */
-export class Option extends ScopedElementsMixin(LitElement) {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export class Option<T = any> extends ScopedElementsMixin(LitElement) {
```
anna-lach, line 19: Can we please use `unknown` instead of `any` here?

packages/components/listbox/src/listbox.ts
```diff
@@ -1,23 +1,128 @@
...
 
-export class Listbox extends LitElement {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ListboxRenderer<T = any> = (option: T, index: number) => Element | TemplateResult;
+
+/**
+ * Container for a list of selectable options.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export class Listbox<T = any> extends ScopedElementsMixin(LitElement) {
```
anna-lach, line 30: Can we please use `unknown` instead of `any` here?

packages/components/listbox/src/listbox.ts
```diff
@@ -1,23 +1,128 @@
...
+    createElement<K extends keyof HTMLElementTagNameMap>(
+      tagName: K,
+      options?: ElementCreationOptions
+    ): HTMLElementTagNameMap[K];
+  }
 }
 
-export class Listbox extends LitElement {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ListboxRenderer<T = any> = (option: T, index: number) => Element | TemplateResult;
```
anna-lach, line 24: Can we please use `unknown` instead of `any` here?

packages/components/listbox/src/option.ts
jpzwarte: As discussed on Slack, leaving it `any` for now

packages/components/listbox/src/listbox.ts
jpzwarte: Same

packages/components/listbox/src/listbox.ts
jpzwarte: Same

Issue #1553, opened 2024-09-27
Checkbox has redundant aria role and property


Magsu:
### Provide a general summary of the issue here

Due to redundant aria-checked, when using a screen reader, checbox’s status (checked/not checked) is read twice. This might confuse screen reader users. The checkbox also contains a duplicate role, which is redundant and might confuse some screen reader software. The semantic HTML input element does already have the correct role and status.

VoiceOver (on mac and iOS) and TalkBack (on Android) do ignore the duplicate status, but NVDA, which is the most commonly used screen reader on PC does not.

### 🤔 Expected Behavior?

The checkbox status is read once.

### 😯 Current Behavior

The checkbox’s status is read twice when focusing on the element using a screen reader.

### 💁 Possible Solution

Remove redundant duplicate role and state from the element.
![image](https://github.com/user-attachments/assets/4ff73edc-fb94-4ba0-b141-e9c8ef2fccd7)

Move required to the correct elements. The required is not interpreted by all screen readers, as it is not part of the active elements.
![image](https://github.com/user-attachments/assets/0f2f3cea-9109-4f23-9aa5-f7d008569925)

The input element does semantically already have everything built in for a screen reader to interpret it correctly.

### 🖥️ Steps to Reproduce

1. Open a page with a checkbox element (e.g. [Checkbox Group Required](Open https://storybook.sanomalearning.design/?path=/story/form-checkbox-group--required)
2. Start a screen reader (NVDA for Windows, VoiceOver will not display this problem correctly due to its differing interpretation method)
3. Focus the checkbox element (for NVDA use tab)

### What browsers are you seeing the problem on? (only for bugs in code)

Chrome

### What operating system are you using?

Windows 10

### 👤 Your name

Valtteri Järvinen

### 🧢 Your product/team

External

### 🎨 Your Theme(s)

Sanoma Learning
```[tasklist]
### Tasks
- [ ] only add type=checkbox, and remove duplicate roles from checkbox, only add them in input
- [ ] make sure the checkbox group/form intergration is ok
```

PR #1628, opened 2024-11-13
Fix/1553 checkbox has redundant aria role and property

anna-lach:
Fixes:
https://github.com/sl-design-system/components/issues/1553


packages/components/checkbox/src/checkbox.spec.ts
```diff
@@ -29,15 +29,18 @@ describe('sl-checkbox', () => {
...
+      expect(input.checked).to.be.false;
     });
 
     it('should be checked when set', async () => {
       el.checked = true;
       await el.updateComplete;
 
       expect(el).to.have.attribute('checked');
-      expect(input).to.have.attribute('checked');
+      expect(input?.getAttribute('aria-checked')).to.equal('true');
```
jpzwarte, line 42: `expect(input).to.have.attribute('aria-checked', 'true');`

packages/components/checkbox/src/checkbox.spec.ts
```diff
@@ -29,15 +29,18 @@ describe('sl-checkbox', () => {
...
+      expect(input?.matches(':checked')).to.be.true;
```
jpzwarte, line 43: `expect(input).to.match(':checked')`

jpzwarte (approved):

packages/components/checkbox/src/checkbox.spec.ts
jpzwarte: Please change all the expectations etc. not just this one :)

PR #1637, opened 2024-11-18
Fix/1560 tabs arent correctly marked for screen reader

anna-lach:


.changeset/rude-llamas-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sl-design-system/tabs': patch
+---
+
+Fix NVDA accessibility issues, fix keyboard navigation.
```
jpzwarte, line 5: The way you wrote this, it looks like 2 different things? Could you be more explicit what kind of NVDA accessibility issues you fixed?

packages/components/menu/src/menu-button.ts
```diff
@@ -27,6 +27,11 @@ declare global {
  */
 @localized()
 export class MenuButton extends ScopedElementsMixin(LitElement) {
+  /** @internal */
+  static override get observedAttributes(): string[] {
+    return [...super.observedAttributes, 'aria-disabled', 'aria-label'];
```
jpzwarte, line 32: This is awesome, but i'm left wondering if we couldn't make a mixin that does this? I'm entirely not sure you can change static things from a mixin, but it would be wort the investigation? .... I think you can; see this RFC and the `FormAssociatedMixin` https://github.com/lit/rfcs/issues/43

packages/components/menu/src/menu-button.ts
```diff
@@ -63,6 +68,23 @@ export class MenuButton extends ScopedElementsMixin(LitElement) {
   /** The variant of the button. */
   @property() variant: ButtonVariant = 'default';
 
+  override attributeChangedCallback(name: string, oldValue: string | null, newValue: string | null) {
```
jpzwarte, line 71: Same; would be awesome to have the generic part of this in a mixin

packages/components/tabs/src/tab-group.ts
```diff
@@ -122,12 +128,21 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
...
+    focusInIndex: (elements: Tab[]) => {
+      const index = elements.findIndex(el => el.selected);
+      return index === -1 ? 0 : index;
+    },
+    elements: () => (this.#menu && isPopoverOpen(this.#menu) ? this.#menuItems : this.tabs) || [],
     isFocusableElement: (el: Tab) => !el.disabled
   });
 
+  /** Menu element, is shown when the tabs are overflowing. */
+  #menu: Menu | undefined;
```
jpzwarte, line 141: Please do `#menu?: Menu;`; it's effectively the same, but less code

packages/components/tabs/src/tab-group.ts
```diff
@@ -215,10 +242,9 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
               </div>
             </div>
           </div>
-
           ${this.showMenu
             ? html`
-                <sl-menu-button aria-label=${msg('Show all')} fill="ghost">
+                <sl-menu-button aria-label=${msg('Show all')} fill="ghost" @keydown=${this.#onKeydown}>
```
jpzwarte, line 247: Minor nitpick, but please more `@keydown` to the start. I always try to put event handlers first, then boolean props, then props, then attrs.

jpzwarte (approved):

packages/components/menu/src/menu-button.ts
anna-lach: Those observed attributes can differ per component, so I left setting them in the components, but I've created a mixin that makes the rest.

packages/components/shared/src/mixins/observe-attributes-mixin.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export function ObserveAttributesMixin<T extends Constructor<LitElement>>(
+  constructor: T
+): T & Constructor<ObserveAttributesMixinInterface> {
+  class ObserveAttributesImpl extends constructor {
+    static #observedAttributes: string[] = [];
+
+    #targetElement?: HTMLElement;
+
+    setTargetElement(target: HTMLElement): void {
```
Diaan, line 20: when using this in the context of the component that uses the mixin, is this name clear enough? Should "attributes" somehow come back in the name so it's clear it's part of the mixin?

Diaan (approved):

packages/components/checkbox/src/checkbox.ts
```diff
@@ -98,6 +98,8 @@ export class Checkbox<T = unknown> extends FormControlMixin(LitElement) {
   override connectedCallback(): void {
     super.connectedCallback();
 
+    this.setObservedAttributes(['aria-disabled', 'aria-label', 'aria-labelledby']);
```
jpzwarte, line 101: You are repeating yourself here. You could add it to the mixin function?

```ts
extends ObserveAttributesMixin(LitElement, ['aria-disabled', 'aria-label', 'aria-labelledby']) {`
```

anna-lach:
Mixin improvement will be added in the future: https://github.com/sl-design-system/components/issues/1652

PR #1642, opened 2024-11-25
Various combobox fixes

jpzwarte:


packages/components/combobox/src/combobox.ts
```diff
@@ -503,6 +517,30 @@ export class Combobox<T = any, U = T> extends FormControlMixin(ScopedElementsMix
       }
     } else if (!this.wrapper?.matches(':popover-open') && ['ArrowDown', 'ArrowUp'].includes(event.key)) {
       this.wrapper?.showPopover();
+    } else if (['ArrowLeft', 'ArrowRight'].includes(event.key) && this.input.selectionStart === 0) {
```
anna-lach, line 520: In some cases I can get two tags focused at once (with arrow keys):

image

Diaan:
I would remove the 'current' label from an option the moment the dropdown closes. Now you can close the dropdown with `esc` and then with enter you select an option you didn't even see was selected.

Diaan:
https://github.com/user-attachments/assets/7d6c1d7f-5781-49cc-aa87-5e3acf9fd44f

This happens when i'm using only my up-arrow on the keyboard; the order of the options is a bit weird. Is it because they have the same index maybe?

.changeset/shiny-ants-cover.md
```diff
@@ -0,0 +1,12 @@
+---
+'@sl-design-system/combobox': patch
```
anna-lach, line 2: Can you please also add changeset for other component's changes like grid, listbox, tag and so on? :)

anna-lach (approved):

Diaan:
The tags in the lg variant are too big: 
image

PR #1649, opened 2024-11-28
Feature/1597 aria live container

Diaan:


packages/components/announcer/src/announcer.ts
```diff
@@ -0,0 +1,79 @@
...
+/** Sends a notification to the live aria.
+ * Please be aware that sending messages too soon after each other can cause the screenreader
+ * to stop reading the earlier messages, or even skip them, when a new message is sent.
+ * If you want to send multiple messages in a row, consider using a single message with all the information.
+ *
+ * @param message - The message to send to the live aria.
+ * @param urgency - The urgency of the message. Default is 'polite'.
+ */
+export function sendToAnnouncer(message: string, urgency?: 'polite' | 'assertive'): void {
+  const liveEvent = new EventEmitter<SlAnnounceEvent>(document.body, 'sl-announce');
```
jpzwarte, line 32: No need to do it this way; that is meant more for use in web components. 

`document.body.dispatchEvent(new CustomEvent('sl-announce', { detail: { message, urgency }}));`

packages/components/announcer/src/announcer.ts
```diff
@@ -0,0 +1,79 @@
...
+  override render(): TemplateResult {
+    return html`
+      <div aria-live="polite" aria-atomic="true"></div>
+      <div aria-live="assertive" aria-atomic="false"></div>
+    `;
+  }
+
+  #onLiveEvent(event: SlAnnounceEvent) {
+    const container = this.renderRoot.querySelector(`[aria-live="${event.detail.urgency || 'polite'}"]`);
+    const messageNode = document.createElement('sl-announcement');
```
jpzwarte, line 71: Since `<sl-announcer>` creates the `<sl-announcement>`, it owns the element. I think it should then also be the one to remove it after 500ms, not announcement itself.

packages/components/announcer/src/announcement.ts
```diff
@@ -0,0 +1,24 @@
...
+  static override styles: CSSResultGroup = styles;
+
+  override render() {
+    return html`<slot></slot> `;
+  }
+
+  override connectedCallback(): void {
+    super.connectedCallback();
+    setTimeout(() => {
+      this.remove();
```
jpzwarte, line 21: Do this in `<sl-announcer>`, since that owns this element (see comment in announcer.ts)

packages/components/announcer/src/announcer.ts
```diff
@@ -0,0 +1,79 @@
...
+    messageNode.innerText = event.detail.message;
+
+    // make sure the message is not already in the container
+    if (container?.textContent?.indexOf(event.detail.message) === -1) {
```
jpzwarte, line 75: Perhaps do this check *before* we create a new `<sl-annoucement>` element?

packages/components/progress-bar/src/progress-bar.spec.ts
```diff
@@ -47,11 +47,11 @@ describe('sl-progress-bar', () => {
     expect(progressBar).to.have.attribute('aria-valuenow', '0');
   });
 
-  it('should have aria-live by default', () => {
-    const ariaLive = el.renderRoot.querySelector('#live') as HTMLElement;
-    expect(ariaLive).to.have.attribute('aria-live', 'polite');
-    expect(ariaLive).to.have.rendered.text('state active 0%');
-  });
+  // it('should have aria-live by default', () => {
```
jpzwarte, line 50: Can you either uncomment it, fix it or remove it? :)

packages/components/progress-bar/src/progress-bar.spec.ts
```diff
@@ -116,13 +116,13 @@ describe('sl-progress-bar', () => {
     expect(helperIcon).to.exist;
   });
 
-  it('should change the aria-live when the value and variant have changed', async () => {
-    el.value = 60;
-    el.variant = 'warning';
-    await el.updateComplete;
+  // it('should change the aria-live when the value and variant have changed', async () => {
```
jpzwarte, line 119: Same

packages/components/progress-bar/src/progress-bar.spec.ts
```diff
@@ -116,13 +116,13 @@ describe('sl-progress-bar', () => {
...
+  //   el.value = 60;
+  //   el.variant = 'warning';
+  //   await el.updateComplete;
 
-    const ariaLive = el.renderRoot.querySelector('#live') as HTMLElement;
-    expect(ariaLive).to.have.attribute('aria-live', 'polite');
-    expect(ariaLive).to.have.rendered.text('state warning 60%');
-  });
+  //   const ariaLive = el.renderRoot.querySelector('#live') as HTMLElement;
```
jpzwarte, line 124: Same

packages/components/progress-bar/src/progress-bar.stories.ts
```diff
@@ -1,3 +1,4 @@
+import '@sl-design-system/announcer/register.js';
```
jpzwarte, line 1: Not necessary if already in `preview.ts`?

packages/components/grid/src/stories/pagination.stories.ts
```diff
@@ -1,3 +1,4 @@
+import '@sl-design-system/announcer/register.js';
```
jpzwarte, line 1: Not necessary if already in `preview.ts`?

packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -52,6 +51,8 @@ export class ProgressBar extends ScopedElementsMixin(LitElement) {
   /** Progress value (from 0...100). */
   @property({ type: Number }) value = 0;
 
+  private shouldSendToAnnouncer = true;
```
jpzwarte, line 54: Please rename to `#shouldAnnounce`

jpzwarte (approved):

packages/components/paginator/src/paginator.ts
```diff
@@ -345,8 +346,6 @@ export class Paginator extends ScopedElementsMixin(LitElement) {
           <sl-icon name="caret-right-solid"></sl-icon>
         </sl-button>
       </nav>
-      <!-- We want this to be read every time the active page changes. -->
-      <div id="live" aria-live="polite" aria-atomic="true">${msg(str`Page ${this.page} of ${this.#pages}`)}</div>
```
anna-lach, line 349: When we remove it, it's not announced that when we change the page-size in `sl-paginator-size` that the active page is being changed into `1` (it's going back to the first page). 
Also it will not work with the grid (will not be announced), when we change the filtering or sorting (it also changes the page to page `1` again).

packages/components/progress-bar/src/progress-bar.spec.ts
anna-lach, line 50: Do we want to keep those commented lines?

packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -91,10 +109,8 @@ export class ProgressBar extends ScopedElementsMixin(LitElement) {
       </div>
       <div id="helper" class="helper">
         <slot></slot>
-        <span id="live" aria-live="polite" aria-busy=${ifDefined(this.indeterminate)}>
-          ${msg('state')} ${this.variant ? html`${this.#getLocalizedVariant()}` : html`${msg('active')}`}
-          <!-- We want '%' to be read every time the value changes. -->
-          <span aria-live="polite" aria-atomic="true">${this.value}%</span>
+        <span id="live" aria-busy=${ifDefined(this.indeterminate)}>
+          ${msg('state')}: ${this.variant ? html`${this.#getLocalizedVariant()}` : html`${msg('active')}`}
```
anna-lach, line 113: Right now I'm not getting an information anymore that the state has changed (variant)?
In this example it should be said that the state has changed into `success`? Right now I'm getting only information that it's 100%.
image

packages/components/progress-bar/src/progress-bar.ts
```diff
@@ -66,6 +67,23 @@ export class ProgressBar extends ScopedElementsMixin(LitElement) {
     }
   }
 
+  override updated(changes: PropertyValues<this>): void {
+    super.updated(changes);
+
+    if (changes.has('value')) {
```
anna-lach, line 73: `variant` change should be also announced, right?

packages/components/progress-bar/src/progress-bar.spec.ts
anna-lach, line 119: Do we want to keep those commented lines?

packages/components/announcer/src/announcer.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  interface HTMLElementTagNameMap {
+    'sl-announcer': Announcer;
+  }
+}
+
+export type SlAnnounceEvent = CustomEvent<{ message: string; urgency?: 'polite' | 'assertive' }>;
+
+/** Sends a notification to the live aria.
```
anna-lach, line 23: Do we have plans to add the component description on the documentation website as well? :)

anna-lach: Maybe we need some changes here as well?
https://proud-water-0d02ee203-1649.westeurope.2.azurestaticapps.net/categories/components/progress-bar/accessibility/

packages/components/announcer/src/announcer.ts
```diff
@@ -0,0 +1,79 @@
...
+ * Please be aware that sending messages too soon after each other can cause the screenreader
+ * to stop reading the earlier messages, or even skip them, when a new message is sent.
+ * If you want to send multiple messages in a row, consider using a single message with all the information.
+ *
+ * @param message - The message to send to the live aria.
+ * @param urgency - The urgency of the message. Default is 'polite'.
+ */
+export function sendToAnnouncer(message: string, urgency?: 'polite' | 'assertive'): void {
```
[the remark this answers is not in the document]
Diaan: but when you use the announce function you always need the webcomponent, right? What would be left out by the treeshaking then?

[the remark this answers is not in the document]
jpzwarte: The user of the `announce()` API may not be the one that adds the `<sl-announcer>` element to the DOM (think microfrontends/libraries)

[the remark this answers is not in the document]
jpzwarte: So if the bundler can't tree shake the web component, it'll get loaded, even though it's not necessary at that point.

packages/components/announcer/src/announcer.ts
Diaan: But then it doesn't even need to be an `<sl-announcement>`, then it can just be a span. Or maybe even better, a list item...

packages/components/progress-bar/src/progress-bar.ts
Diaan: But not with every value, that will get annoying pretty fast...

packages/components/progress-bar/src/progress-bar.ts
Diaan: If an error occured while the progressbar is filling up, shouldn't there be more information than just the state changing? An inline message for example, with more information on what when wrong? If we then also announce the change in state you have the information twice.
I've added the state of the element when it's 100%, but I don't really see an added value in having it announced when the state changes; that will not be enough information.

Diaan: I've updated the documentation for the progressbar to indicate what is does and does not do and what we can't possibly do in the component (because we don't know the reason behind things happening) and what the dev need to do in their cfa.

anna-lach (approved):

PR #1653, opened 2024-12-02
Improve horizontal scrolling in grid

jpzwarte:


anna-lach (approved):

packages/components/grid/src/grid.scss
```diff
@@ -143,7 +228,9 @@ tr {
 th,
 td {
   align-items: center;
-  background: var(--_vertical-border), var(--_cell-background);
+
+  // background: var(--_vertical-border), var(--_cell-background);
+  background: var(--_cell-background);
```
Diaan, line 233: comment

packages/components/grid/src/grid.ts
```diff
@@ -304,12 +306,23 @@ export class Grid<T = any> extends ScopedElementsMixin(LitElement) {
...
-        <tbody @visibilityChanged=${this.#onVisibilityChanged} part="tbody">
+        <tbody @visibilityChanged=${this.#onVisibilityChanged} id="tbody" part="tbody">
           ${virtualize({
             items: this.view.rows,
             renderItem: (item, index) => this.renderItem(item, index)
           })}
         </tbody>
+        ${this.scrollbar
+          ? html`
+              <tfoot>
```
Diaan, line 317: is the `tfoot` the right element for this? (although i don't know what element would fit better semantically 😅 )

packages/components/grid/src/grid.ts
```diff
@@ -457,7 +470,7 @@ export class Grid<T = any> extends ScopedElementsMixin(LitElement) {
       .filter(col => !col.hidden && col.autoWidth)
       .forEach(col => {
         const index = this.view.headerRows[this.view.headerRows.length - 1].indexOf(col),
-          cells = this.renderRoot.querySelectorAll<HTMLElement>(`:where(td, th):nth-child(${index + 1})`);
+          cells = this.renderRoot.querySelectorAll<HTMLElement>(`:where(tbody td, thead th):nth-child(${index + 1})`);
```
Diaan, line 473: what about th cells in the tbody, when you have a "heading" in the first column in the body for example.

packages/components/grid/src/grid.scss
```diff
@@ -42,55 +45,137 @@
...
+    }
+  }
+}
+
+:host([scrollable-start]) {
+  --_vertical-border-end: linear-gradient(var(--_border-color), var(--_border-color)) 100% 100% / var(--_border-width)
+    100% no-repeat;
+
+  table:not(:has(.sticky-start-last))::before {
+    background: linear-gradient(to right, rgb(79 79 79 / 10%), transparent);
```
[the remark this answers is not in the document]
jpzwarte: Not yet; Roald was going to do this.

packages/components/grid/src/grid.ts
jpzwarte: `<tfoot>` is the only element that is acceptable: the HTML parser is strict when it comes to allowed elements inside of `<table>`. So this is a necessary "hack".

packages/components/grid/src/grid.ts
jpzwarte: Is that correct HTML? I thought `<th>` elements are only allowed in the `<thead>` element?

jpzwarte: Changed

packages/components/grid/src/grid.scss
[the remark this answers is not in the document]
jpzwarte: But this is most likely happening on a different branch. We need to refactor the styling of the grid anyway to use the new tokens. So i want to keep it like this for now.

Diaan (approved):

PR #1659, opened 2024-12-05
Fix/1554 user focus can escape dialog when navigating with tab

anna-lach:
- Implements focus-trap in the `dialog`
- Creates `FocusTrapController`


Diaan (approved):

packages/components/dialog/src/dialog.ts
```diff
@@ -4,6 +4,7 @@ import { Button } from '@sl-design-system/button';
 import { ButtonBar } from '@sl-design-system/button-bar';
 import { Icon } from '@sl-design-system/icon';
 import { type EventEmitter, breakpoints, event } from '@sl-design-system/shared';
+import { FocusTrap, createFocusTrap } from 'focus-trap';
```
jpzwarte, line 7: We should only import the `FocusTrap` *type* right?

packages/components/dialog/src/dialog.ts
```diff
@@ -65,6 +66,8 @@ export class Dialog extends ScopedElementsMixin(LitElement) {
   /** @internal */
   static override styles: CSSResultGroup = [breakpoints, styles];
 
+  #focusTrap?: FocusTrap;
```
jpzwarte, line 69: How about making this a `FocusTrapController`? And hide the logic inside the controller?

jpzwarte (approved):

packages/components/dialog/src/dialog.ts
jpzwarte: That way we also don't use `focus-trap` directly *except* inside the controller?

packages/components/dialog/src/dialog.ts
```diff
@@ -94,6 +97,14 @@ export class Dialog extends ScopedElementsMixin(LitElement) {
     super.connectedCallback();
 
     this.inert = true;
+
+    this.#focusTrapController = new FocusTrapController(this, { disableCancel: this.disableCancel });
```
jpzwarte, line 101: Is this necessary in `connectedCallback` perse? Can't you initialize the controller on line 69?

jpzwarte: You probably also want to update `disableCancel` when that property changes? So add that code to `updated()` lifecycle callback?

packages/components/dialog/src/dialog.ts
```diff
@@ -94,6 +97,14 @@ export class Dialog extends ScopedElementsMixin(LitElement) {
...
+  }
+
+  override disconnectedCallback(): void {
+    super.disconnectedCallback();
+
+    this.#focusTrapController?.deactivate();
```
jpzwarte, line 107: Can't this happen automatically? In the `hostDisconnected` callback in the controller?

jpzwarte (approved):

Diaan (approved):

PR #1676, opened 2024-12-12
Fix/1651 sl-grid strange numeric sort behavior

anna-lach:
- Improving sorting to make it working with numbers properly as well,
- Fixing custom sorting to make it working with `path`
- Adding another example of custom sorting


jpzwarte:
Could you also perhaps add a test for the improved sorting?

PR #1679, opened 2024-12-16
Refactor button-bar styling to use contextual tokens

jpzwarte:


packages/components/button-bar/src/button-bar.scss
```diff
@@ -1,46 +1,36 @@
 :host {
-  // TODO: remove this "hack" when style queries are possible.
-  --sl-ON: initial; // property gets fallback value
```
anna-lach, line 3: Can you please update button-bar documentation? :)

packages/components/button-bar/src/button-bar.scss
```diff
@@ -1,46 +1,36 @@
...
-  --sl-OFF: var(--sl-button-bar-direction-row);
```
anna-lach, line 4: Can you please also make changes of the button bar usage in the dialog?
In the mobile version of the dialog we want to have buttons reversed.

image

anna-lach (approved):

PR #1682, opened 2024-12-19
Upgrade to Angular 18, and drop support for 17

jpzwarte:


packages/angular/.storybook/main.ts
```diff
@@ -1,20 +1,9 @@
-import type { StorybookConfig } from '@storybook/angular';
+import { type StorybookConfig } from '@storybook/angular';
 
 const config: StorybookConfig = {
```
anna-lach, line 3: The Angular components/wrappers are not being loaded in the storybook?

image

.changeset/perfect-hotels-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sl-design-system/angular': major
+---
+
+Upgrade to Angular 18 and drop support for 17
```
Diaan, line 5: is this checked with the cfa's?

Diaan (approved):

packages/angular/.storybook/main.ts
jpzwarte: Hmm, apparently this happens on `main` as well. It worked fine when i ran Angular storybook locally. I'll investigate.

.changeset/perfect-hotels-flow.md
jpzwarte: No, but it's best practice for Angular libraries to support n + n-1. So now that Angular 19 is out, 18 + 19, but not 17.

packages/angular/.storybook/main.ts
jpzwarte: Fixed it! Apparently you to explicit define a `constructor` and call `super()` when you extend the `CePassthrough` directive.

PR #1693, opened 2024-12-31
New tree component

jpzwarte:


packages/components/data-source/src/list-data-source.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  constructor() {
+    super();
+
+    this.items = [...people];
+    this.size = people.length;
+  }
+
+  override update(): void {
+    // empty
```
anna-lach, line 18: Do we want to leave it that way? :)

packages/components/tree/src/flat-tree-data-source.ts
```diff
@@ -0,0 +1,151 @@
...
+  loadChildren?(node: T): Promise<T[]>;
+  selects?: 'single' | 'multiple';
+}
+
+/**
+ * A tree model that represents a flat list of nodes.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export class FlatTreeDataSource<T = any> extends TreeDataSource<T> {
+  #mapping: FlatTreeDataSourceMapping<T>;
```
anna-lach, line 22: Can you please add spacing between variables and maybe add descriptions as well? :)

packages/components/tree/src/nested-tree-data-source.ts
```diff
@@ -0,0 +1,139 @@
...
+  loadChildren?(node: T): Promise<T[]>;
+  selects?: 'single' | 'multiple';
+}
+
+/**
+ * A tree model that represents a nested list of nodes.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export class NestedTreeDataSource<T = any> extends TreeDataSource<T> {
+  #mapping: NestedTreeDataSourceMapping<T>;
```
anna-lach, line 22: Can you please add spacing between variables and maybe add descriptions as well?

packages/components/tree/src/tree-data-source.ts
```diff
@@ -0,0 +1,381 @@
...
+  /** Enables single or multiple selection of tree nodes. */
+  selects?: 'single' | 'multiple';
+}
+
+/**
+ * Abstract class used to provide a common interface for tree data.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export abstract class TreeDataSource<T = any> extends DataSource<T, TreeDataSourceNode<T>> {
+  #loadChildren?: TreeDataSourceOptions<T>['loadChildren'];
```
anna-lach, line 75: Can you please add spacing between variables and maybe add descriptions as well?

packages/components/tree/src/tree.ts
```diff
@@ -0,0 +1,263 @@
...
+  /**
+   * The custom elements used for rendering this tree. If you are using a custom renderer
+   * to render the tree nodes, any custom elements you use in the renderer need to be specified
+   * via this property. Otherwise those custom elements will not initialize, since the tree
+   * uses a Scoped Custom Element Registry.
+   */
+  @property({ attribute: false }) scopedElements?: Record<string, typeof HTMLElement>;
+
+  /** @internal Emits when the user selects a tree node. */
+  @event({ name: 'sl-select' }) selectEvent!: EventEmitter<SlSelectEvent<TreeDataSourceNode<T>>>;
```
anna-lach, line 98: I'm just thinking, maybe we should rename SlSelectEvent from `sl-select` to sth else? Since `sl-select` is also used for select component?

packages/components/tree/src/tree.ts
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  override willUpdate(changes: PropertyValues<this>): void {
+    super.willUpdate(changes);
+
+    if (changes.has('dataSource')) {
+      if (this.dataSource?.selects === 'multiple') {
+        this.setAttribute('aria-multiselectable', 'true');
+      } else {
+        this.removeAttribute('aria-multiselectable');
```
anna-lach, line 128: Maybe it should be `this.setAttribute('aria-multiselectable', 'false');`? So `aria-multiselectable="false"` by default?

anna-lach (approved):

packages/components/data-source/src/list-data-source.spec.ts
jpzwarte: In the unit tests? Yes :)

packages/components/tree/src/tree.ts
jpzwarte: Good catch.

packages/components/tree/src/tree.ts
jpzwarte: If we change it, let's do it in a different PR :)

anna-lach: hmm maybe `sl-choose` or sth? Sounds a little bit weird, but still I think that maybe it would be better with a different name than `select` component. Maybe @Diaan do you have any other idea? :)

PR #1695, opened 2025-01-06
Fix skip table links when at viewport edge

Diaan:


packages/components/grid/src/grid.ts
```diff
@@ -311,11 +323,12 @@ export class Grid<T = any> extends ScopedElementsMixin(LitElement) {
         id="table-start"
         href="#table-end"
         class="skip-link-start"
+        @focus=${(e: Event & { target: HTMLSlotElement }) => this.#onSkipToFocus(e, 'top')}
         @click=${(e: Event & { target: HTMLSlotElement }) => this.#onSkipTo(e, 'end')}
       >
         ${msg('Skip to end of table')}</a
       >
-      <table part="table" aria-rowcount=${this.dataSource?.items.length || 0}>
+      <table part="table" aria-rowcount=${this.dataSource?.items.length || 0} class="table-anchor">
```
anna-lach, line 331: Is it ok that we have 'skip links' sometimes covered by the table (this is in chrome)?

image

anna-lach (approved):

packages/components/grid/src/grid.ts
Diaan: That shouldn't happen 🤔 I'll look into that

jpzwarte (approved):

PR #1699, opened 2025-01-09
Fix Inline messages reading

Diaan:


packages/components/inline-message/src/inline-message.ts
```diff
@@ -167,6 +202,7 @@ export class InlineMessage extends ScopedElementsMixin(LitElement) {
       'animationend',
       () => {
         this.dismissEvent.emit();
+        announce(msg(str`Message closed`));
```
anna-lach, line 205: Maybe we can add here a `title` or sth that would describe more which message has been closed?

Diaan: I actually already think this message is overkill, since it is announced after a direct user action (closing the message), so I assume that if a user has navigated to this button that is after reading the title/message and they decide to close it. So they know which message they are closing.

anna-lach (approved):

packages/locales/src/nl.ts
```diff
@@ -9,6 +9,7 @@ export const templates = {
   s129fac0ae1df7c8f: 'waarschuwing',
   s13fdff160ffad275: 'Wis tekst',
   s15abf4bcf7cba79f: 'items per pagina',
+  s197b12ac6e95ad9c: str`Bericht gesloten`,
```
jpzwarte, line 12: There should not be a `str` function here; that should only be there when there is a variable inside the text.

jpzwarte (approved):

PR #1708, opened 2025-01-15
Fix/1559 accordion with external toggle give no status for a screen reader

anna-lach:


website/src/categories/components/accordion/accessibility.md
```diff
@@ -41,3 +54,5 @@ WAI-ARIA roles, states, and properties for an accordion provide essential inform
 {.ds-table .ds-table-align-top}
 
 </div>
+
+</section>
```
jpzwarte, line 58: This was missing i assume?

jpzwarte (approved):

website/src/categories/components/accordion/accessibility.md
anna-lach: yes :)

Diaan (approved):

PR #1710, opened 2025-01-21
Fix/1658 refactoring custom icons set up

Diaan:


packages/components/icon/src/icon.stories.ts
```diff
@@ -129,9 +129,10 @@ export const All: Story = {
           justify-items: center;
         }
       </style>
+      change
```
jpzwarte, line 132: Remove?

scripts/output-pages-to-folders-svg.js
```diff
@@ -0,0 +1,32 @@
+import { kebabCase } from 'change-case';
```
jpzwarte, line 1: This is a generated file?

jpzwarte (approved):

scripts/output-pages-to-folders-svg.js
Diaan: What? Where this line is written or where i'm importing from?
`[output-pages-to-folders-svg.js](https://github.com/sl-design-system/components/pull/1710/files/88586898e867d7569bad06fd6ddc8d50ade116cb#diff-0b84967c7e2ded2889c1115abd13f148b3c29a829423dc22e8f2632480c276a8)` Is a script i wrote (or copied and changed) from the figma export bundle. There was no way to put the svgs of each page into a folder corresponding to the name of the page, so i made it myself.

jpzwarte: Check!

anna-lach (approved):

PR #1711, opened 2025-01-21
Fix/1490 accessibility features are not correctly marked in the select element

anna-lach:
Select component accessibility fixes


Diaan (approved):

packages/components/select/src/select-button.ts
```diff
@@ -81,7 +76,9 @@ export class SelectButton extends ScopedElementsMixin(LitElement) {
     }
 
     return html`
-      <div>${selected || this.placeholder || '\u00a0'}</div>
+      <div id="button-content" class=${this.placeholder && !selected ? 'placeholder' : ''}>
```
jpzwarte, line 79: What's the DOM id for?

jpzwarte (approved):

packages/components/select/src/select-button.ts
anna-lach: It's used for the `aria-labelledby` in the `listbox`.

jpzwarte: But that has to cross the shadow DOM, right? `aria-labelledby` doesn't support that.

anna-lach: Right, good one, I'll just add `aria-label` for the `listbox`.

PR #1777, opened 2025-02-05
Fix/1562 menu button submenu a11y

Diaan:


packages/components/menu/src/menu.stories.ts
```diff
@@ -74,6 +87,15 @@ export const Basic: Story = {
   }
 };
 
+export const Bold: Story = {
+  args: {
```
anna-lach, line 91: emphasis: 'bold' should be set in the story, right?

anna-lach (approved):

packages/components/menu/src/menu-button.ts
```diff
@@ -83,7 +83,7 @@ export class MenuButton extends ObserveAttributesMixin(ScopedElementsMixin(LitEl
 
     this.setAttributesTarget(this.button);
 
-    this.button.setAttribute('aria-details', this.menu.id);
+    this.button.setAttribute('aria-controls', this.menu.id);
```
anna-lach, line 86: Can you please also update the documentation?

packages/components/menu/src/menu-button.ts
```diff
@@ -99,6 +99,7 @@ export class MenuButton extends ObserveAttributesMixin(ScopedElementsMixin(LitEl
         @keydown=${this.#onKeydown}
         ?disabled=${this.disabled}
         aria-expanded="false"
+        aria-haspopup="menu"
```
anna-lach, line 102: Can you please also update the documentation?

packages/components/menu/src/menu-item.ts
```diff
@@ -96,6 +103,14 @@ export class MenuItem extends ScopedElementsMixin(LitElement) {
         this.#shortcut.unbind();
       }
     }
+
+    if (changes.has('selectable')) {
+      const selectMode = this.parentElement?.matches('[selects="single"]') ? 'menuitemradio' : 'menuitemcheckbox';
+      this.role = this.selectable ? selectMode : 'menuitem';
+    }
+    if (changes.has('selected')) {
+      this.setAttribute('aria-checked', (this.selected || false).toString());
```
anna-lach, line 112: Can you please also update the documentation?

.changeset/afraid-trees-marry.md
```diff
@@ -0,0 +1,10 @@
+---
+'@sl-design-system/shared': patch
+'@sl-design-system/menu': patch
+---
+
+Fixed several accessibility issues; 
+ - Improved VoiceOver in supoprt Chrome
```
jpzwarte, line 7: Typo

.changeset/afraid-trees-marry.md
```diff
@@ -0,0 +1,10 @@
+---
+'@sl-design-system/shared': patch
```
jpzwarte, line 2: I'm not entirely sure how this applies to the `shared` package? I see you removed a single line there? Can you elaborate?

packages/components/menu/src/menu-item.scss
```diff
@@ -56,34 +67,34 @@ sl-icon[name='chevron-right'] {
...
-  background: var(--sl-elevation-surface-raised-default-idle);
+  background: color-mix(in srgb, var(--_bg-color), var(--_bg-mix-color) calc(100% * var(--_bg-opacity)));
   border-radius: var(--sl-size-borderRadius-default);
   cursor: pointer;
   display: flex;
   gap: var(--sl-space-new-md);
   padding: var(--sl-space-new-sm) var(--sl-space-new-md);
   padding-inline-start: calc(
-    var(--sl-menu-item-indent, 0) * (var(--sl-size-icon-sm) + var(--sl-space-new-md)) + var(--sl-space-new-md)
+    var(--sl-menu-item-indent, 0) * (var(--sl-size-new-icon-sm) + var(--sl-space-new-md)) + var(--sl-space-new-md)
```
jpzwarte, line 83: Please don't use `--sl-space-*` tokens, but use `--sl-size-<number>` instead. The former will eventually be removed/not exported to CSS anymore.

jpzwarte: This applies to all `--sl-space-*` tokens used in this PR :)

packages/components/menu/src/menu-item.ts
```diff
@@ -136,8 +151,13 @@ export class MenuItem extends ScopedElementsMixin(LitElement) {
        */
       setTimeout(() => this.#showSubMenu(), 100);
     } else if (this.selectable) {
-      this.selected = !this.selected;
-      this.selectEvent.emit(this.selected);
+      if (
+        !this.parentElement?.matches('[selects="single"]') ||
```
jpzwarte, line 155: Move `this.parentElement?.matches('[selects="single"]')` into a variable so you don't have to repeat yourself?

packages/components/menu/src/menu-item.ts
```diff
@@ -199,14 +221,19 @@ export class MenuItem extends ScopedElementsMixin(LitElement) {
...
       this.submenu?.focus();
     }
   }
 
   #hideSubMenu(): void {
+    if (!this.submenu) {
+      return;
+    }
     this.submenu?.hidePopover();
+    this.setAttribute('aria-expanded', 'false');
```
jpzwarte, line 236: You are setting `aria-expanded` here directly after show/hide popover. An alternative could be to set it from a `beforetoggle` event listener on `this.submenu`??? That way if a user presses Escape, it'll also update the `aria-expanded` attribute?

packages/components/menu/src/menu.stories.ts
```diff
@@ -43,12 +43,19 @@ export default {
...
   args: {
-    maxWidth: '200px'
+    maxWidth: '200px',
+    emphasis: 'subtle'
+  },
+  argTypes: {
+    emphasis: {
+      control: 'inline-radio',
+      options: ['subtle', 'bold']
+    }
```
jpzwarte, line 53: Should probably add `menuItems: { table: { disable: true }}` here as well

packages/components/menu/src/menu.stories.ts
```diff
@@ -74,6 +87,16 @@ export const Basic: Story = {
   }
 };
 
+export const Bold: Story = {
+  args: {
+    emphasis: 'bold',
+    menuItems: () => html`
+      <sl-menu-item selected>Rename...</sl-menu-item>
```
jpzwarte, line 94: Should this always have the `selected` attribute?

packages/components/menu/src/menu.ts
```diff
@@ -76,6 +78,12 @@ export class Menu extends LitElement {
   /** Determines whether if and how many menu items can be selected. */
   @property() selects?: 'single' | 'multiple';
 
+  /**
+   * The emphasis of the menu.
+   * @default subtle
```
jpzwarte, line 83: I think it should be `@default 'subtle'` since it's a string?

packages/components/menu/src/menu.ts
```diff
@@ -188,6 +200,22 @@ export class Menu extends LitElement {
       );
     });
 
+    this.#propagateEmphasis();
+
     this.#rovingTabindexController.clearElementCache();
   }
+
+  #propagateEmphasis(): void {
+    if (this.#menuItems) {
```
jpzwarte, line 209: Remove the `if` and do `this.menuItems?.forEach(....)`?

packages/components/shared/src/controllers/focus-group.ts
```diff
@@ -238,8 +238,6 @@ export class FocusGroupController<T extends HTMLElement> implements ReactiveCont
       return targetIndex !== -1;
     });
     this.currentIndex = targetIndex > -1 ? targetIndex : this.currentIndex;
-
-    this.focusToElement(this.currentIndex);
```
jpzwarte, line 242: Please elaborate

jpzwarte (approved):

packages/components/menu/src/menu.stories.ts
Diaan: in this scenario, where we want to showcase the styling of the `bold` emphasis i think it should

PR #1788, opened 2025-02-12
Fix bug where a path with an empty string value renders as "No path set"

jpzwarte:


packages/components/grid/src/column.ts
```diff
@@ -187,7 +187,7 @@ export class GridColumn<T = any> extends LitElement {
         </td>
       `;
     } else {
-      return html`<td class=${classes.join(' ')} part=${parts.join(' ')}>${data || 'No path set'}</td>`;
+      return html`<td class=${classes.join(' ')} part=${parts.join(' ')}>${data ?? 'No path set'}</td>`;
```
anna-lach, line 190: Maybe we should use translation here?

anna-lach (approved):

packages/components/grid/src/column.ts
jpzwarte: I don't think so; this is more of a developer message. Not meant to be seen by end-users.

Issue #1216, opened 2024-04-24
[bug][sl-select] Options container Ignores z-index, overlaps sticky header.


kssc4567:
**Describe the bug**
On a page with scrollable content and a sticky header, the sl-select button scrolls under the header but the options scroll above the container, overlapping it.
Header (yellow) has z-index: 9999 and position one of: static, sticky
Options div listbox popover (red) has position: fixed (in addition with enforced z-index: 1)

**To Reproduce**
1. A page with a sticky header and scrolling content and sl-select
2. Click the sl-select to show options
3. Scroll down so sl-select and its options are cropped by the window

**Actual behavior**
sl-select button is behind the header
sl-select-option popover is in front of the header

**Expected behavior**
sl-select button is behind the header
sl-select-option popover is behind the header

**Screenshots**
<img width="799" alt="image" src="https://github.com/sl-design-system/components/assets/163280493/936f30cc-6b68-42f9-bbfc-cf3748f048f8">

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser Chrome
 - Version 123

**Additional context**
- Angular 15

```[tasklist]
### Tasks
- [ ] hide panel on scroll
```

jpzwarte:
@kssc4567 a `[popover]` element is by default display on the "top layer". This is a new layer in the browser where modal dialogs, popovers etc. are shown. This top-layer is by default *above* any other element not in the top-layer (no matter the `z-index`). See https://developer.chrome.com/blog/what-is-the-top-layer

Diaan:
There is quite some discussion about what the correct behaviour is. Researching this I came across 3 options:
1. Keep as is (popover is always on top of everything else)
2. block scrolling when popover is open
3. close popover on scrolling.

Options 2 and 3 are both used in the native HTML-select component. Option 2. in all browsers on MacOS and Firefox on Windows, option 3 in other browsers on Windows.
Options 2 and 3 will also impact the fact that a popover placed in position where it can possibly move around because of scrolling can never be higher than 50% of the viewport; When the trigger (the select in this case) is in the center of the screen the popover can not go past the edge of the screen, because you can't scroll to see the last option.

From a usability viewpoint option 1 sounds like the best option. So the question is, how important is the fact that the dropdown on top of the header might look a little bit funny.

Diaan:
We found a 4th option :D 
[...]

PR #1794, opened 2025-02-17
Refactor select

### LINKED ISSUE #1761: [Contextual Tokens] Refactoring select in code

- [ ] refactor component in code

### 👤 Your name

DR

### 🧢 Your Product/Team

SLDS

### LINKED ISSUE #1783: [select] Refactor to use `<sl-option>`, `<sl-listbox>`; same as combobox

`<sl-select>` was created before `<sl-listbox>`, `<sl-option>`, `<sl-option-group>` etc.

We need to refactor `<sl-select>` and use those components. This should simplify the implementation of select and at the same time improve a11y.

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

jpzwarte:
- Use `<sl-listbox>`, `<sl-option>` and `<sl-option-group>` components instead of custom ones
- Add `clearable` property for clearing the selection
- Hide the listbox popover when focus leaves the `<sl-select>` component
- Show the listbox popover immediately; only animate it when closing

Fixes #1783 
Fixes #1761


packages/components/select/src/select.ts
```diff
@@ -249,9 +253,9 @@ export class Select<T = unknown> extends ObserveAttributesMixin(FormControlMixin
     super.firstUpdated(changes);
 
     requestAnimationFrame(() => {
-      this.button.setAttribute('aria-controls', this.listbox.id);
```
anna-lach, line 252: Why there is no `aria-controls`?

anna-lach: If we really want to remove `aria-controls`, please remove it as well from the documentation, from the accessibility WAI ARIA description. :)

.changeset/dull-penguins-lay.md
```diff
@@ -0,0 +1,17 @@
...
+BREAKING: Remove `<sl-select-option>` and `<sl-select-option-group>` and use `<sl-option>` and `<sl-option-group>` from the `@sl-design-system/listbox` package instead.
+
+This change was made to align the select component with the listbox component. The listbox component is a more flexible and powerful component that can be used to create a variety of custom select-like components.
+
+To update your code, replace all instances of `<sl-select-option>` with `<sl-option>` and `<sl-select-option-group>` with `<sl-option-group>`. Make sure you have those custom elements loaded. If not, you can import them from the `@sl-design-system/listbox` package.
+
+Various improvements:
+- Add `clearable` property for clearing the selection
+- Add `aria-hidden="true"` to the placeholder content
+- Add `aria-placeholder` to the `<sl-select-button>` when the placeholder is shown
```
anna-lach, line 14: Please describe it in the documentation, in the accessibility tab (WAI-ARIA part) as well.

packages/components/select/src/select-button.ts
```diff
@@ -13,6 +19,9 @@ declare global {
   }
 }
 
+Icon.register(faCircleXmark, fasCircleXmark);
```
anna-lach, line 22: Maybe those icons need to be added to the `core.json` file instead of registering them here?

image

packages/components/select/src/select.stories.ts
```diff
@@ -49,39 +51,40 @@ export default {
           ${slot?.() ??
           html`
             <sl-select
+              ?clearable=${clearable}
```
anna-lach, line 54: Clicking on the `label` in the `form-field` should focus the select component (select-button), right?
Just like it works with other form fields.
Right now it's not working.

packages/components/select/src/select.ts
```diff
@@ -265,25 +269,23 @@ export class Select<T = unknown> extends ObserveAttributesMixin(FormControlMixin
...
       <slot name="button"></slot>
-      <div
+      <sl-listbox
         ${anchor({
           element: this.button,
           offset: Select.offset,
           position: 'bottom-start',
-          viewportMargin: Select.viewportMargin,
-          rootMarginTop: this.rootMarginTop
+          rootMarginTop: this.rootMarginTop,
```
anna-lach, line 277: In the current version of the select, when we scroll down and the select is opened, when it's not visible, the dropdown is closing and when we scroll up and the select is visible again, the dropdown is visible again as well (when the select was not closed explicitly by the user):

https://storybook.sanomalearning.design/?path=/story/form-select--hide-when-out-of-view

In the version from this PR it's not working the same, the listbox is not being shown again, when I scroll up:

https://kind-plant-088739303-1794.westeurope.1.azurestaticapps.net/?path=/story/form-select--hide-when-out-of-view

Diaan (approved):

Diaan:
image
it is possible to have 2 selected options??
https://kind-plant-088739303-1794.westeurope.1.azurestaticapps.net/?path=/story/form-select--groups

packages/components/select/src/select.ts
jpzwarte: I removed `aria-controls` because it won't work: `<sl-select-button>` is in the light DOM. `<sl-listbox>` is in the shadow DOM. And ARIA can't cross it. Even though those attributes are present on `main`, it doesn't work there. So to avoid confusion, i've removed it entirely here.

.changeset/dull-penguins-lay.md
jpzwarte: I'll add it for now, since that is the current practice.

packages/components/select/src/select.ts
jpzwarte: I think we should simplify things and just hide the popover when a user starts scrolling *outside* the popover. That's what Spectrum does. Native select doesn't allow *any* scrolling when a context menu is open.
But have we discussed this before? I can't recall.

anna-lach: I think we were discussing it some time ago, @Diaan right?

anna-lach: https://github.com/sl-design-system/components/issues/1216

jpzwarte: I think i broke the behavior after using `allow-discrete` here: ```  @media (prefers-reduced-motion: no-preference) {
    transition: 0.2s cubic-bezier(0.25, 0, 0.3, 1);
    transition-behavior: allow-discrete;
    transition-property: display, opacity;
  }
```

Comment (jpzwarte) [reply packages/components/select/src/select.ts:278]: Because the popover is now closed when this code runs: 
```ts
export const updatePopoverVisibility = (host: HTMLElement | undefined, outOfView: boolean): void => {
  if (outOfView) {
    host?.style.setProperty('display', 'none');
  } else {
    host?.style.removeProperty('display');
  }
};
```

jpzwarte: Disregard this; it's the `focusout` event handler that hides the popover.

packages/components/select/src/select-button.ts
```diff
@@ -48,19 +66,27 @@ export class SelectButton extends ScopedElementsMixin(LitElement) {
...
   }
 
   override updated(changes: PropertyValues<this>): void {
     super.updated(changes);
 
-    if (changes.has('disabled')) {
-      this.tabIndex = this.disabled ? -1 : 0;
+    if (changes.has('placeholder') || changes.has('selected')) {
+      if (this.placeholder && !this.selected) {
+        this.setAttribute('aria-placeholder', this.placeholder);
```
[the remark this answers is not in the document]
jpzwarte: Reverted back to the old behavior where the placeholder is read, but not marked as a placeholder.

website/src/categories/components/select/accessibility.md
```diff
@@ -39,38 +39,37 @@ WAI-ARIA Roles, States, and Properties for a select component provide essential
 |-|-|-|-|
 |`role`|`'combobox'`|Identifies the select-button as a combobox.|no|
 |`role`|`'listbox'`|Identifies the list of options as a listbox.|no|
-|`aria-controls`|string|Refers to the `id` of the listbox that the button controls.|no|
+|`aria-placeholder`|string|Defines a string that labels the action that will be performed when the user interacts with the select.|no|
```
anna-lach, line 42: `aria-placeholder` has been removed from the component, right?

anna-lach (approved):

anna-lach:
Clicking the `label` in the `form-field` is not focusing the `select` when I select an option in it.

Diaan (approved):

PR #1813, opened 2025-02-28
Tag design tweaks

jpzwarte:


packages/components/tag/src/tag-list.ts
```diff
@@ -85,13 +94,13 @@ export class TagList extends ScopedElementsMixin(LitElement) {
...
 
-  override async connectedCallback(): Promise<void> {
+  override connectedCallback(): void {
     super.connectedCallback();
 
     this.setAttribute('role', 'list');
 
     // Calculate the max inline size of the stack *before* we start the observer
-    this.#maxStackInlineSize = await this.#getMaxStackInlineSize();
+    // this.#maxStackInlineSize = await this.#getMaxStackInlineSize();
```
anna-lach, line 103: Do we want to keep it?

packages/components/tag/src/tag-list.ts
```diff
@@ -41,14 +40,11 @@ export class TagList extends ScopedElementsMixin(LitElement) {
...
-  #maxStackInlineSize = 0;
-
   /**
-   * Observe changes to the size of the tag-list so we can determine when to display
-   * a counter with amount of hidden tags.
+   * Observe size changes so we can determine when to display a counter
+   * with the amount of hidden tags.
    */
-  #resizeObserver = new ResizeObserver(() => this.#updateVisibility());
+  #resizeObserver = new ResizeObserver(entries => this.#onResize(entries));
```
anna-lach, line 47: I cannot focus any tag in the list, all tags have "tabindex="-1":
https://kind-plant-088739303-1813.westeurope.1.azurestaticapps.net/?path=/story/feedback-status-tag-list--stacked-removable

but I found it only when I go eg. to the story https://kind-plant-088739303-1813.westeurope.1.azurestaticapps.net/?path=/story/feedback-status-tag-list--stacked 
and then go back to this one:
https://kind-plant-088739303-1813.westeurope.1.azurestaticapps.net/?path=/story/feedback-status-tag-list--stacked-removable
so it looks like sth is not being refreshed.

jpzwarte: The inability to focus is related to resizing the tag-list. We'll create a separate issue for that (#1820)

Diaan (approved):

Diaan:
Isn't `+1` more logical than `1+` on the stack? 
`+1` means "this PLUS one more. while
`1+` means "more than one"

PR #1819, opened 2025-03-04
Refactor inline message styles

jpzwarte:


packages/components/inline-message/src/inline-message.ts
```diff
@@ -77,16 +70,48 @@ export class InlineMessage extends ScopedElementsMixin(LitElement) {
...
+  get size(): InlineMessageSize {
+    return this.#size;
+  }
+
+  /**
+   * The size of the inline message.
+   * @default 'auto'
+   */
+  @property({ reflect: true })
+  set size(size: InlineMessageSize) {
```
anna-lach, line 101: It looks like there is sth wrong with the `sm` and `md` sizes?

image

Also the close button should be aligned to top?

packages/components/inline-message/src/inline-message.ts
```diff
@@ -45,29 +47,20 @@ export class InlineMessage extends ScopedElementsMixin(LitElement) {
...
-  #maxInlineSize = 0;
-  /** The previous width of the message, to check if only the that has changed (changes in height shouldn't trigger a recheck) */
-  #previousInlineSize = 0;
+  /** The original size, set by the user, before any automatic behavior was applied. */
+  #originalSize: InlineMessageSize = 'auto';
 
-  /** @internal Emits when the inline message is dismissed. */
-  @event({ name: 'sl-dismiss' }) dismissEvent!: EventEmitter<SlDismissEvent>;
+  /** The current size. */
+  #size: InlineMessageSize = 'auto';
```
anna-lach, line 54: Why `auto` and not just making `md` default like we have in other components?

packages/components/inline-message/src/inline-message.ts
jpzwarte: Hmm, i don't think a title is meant to be used with the small size.

jpzwarte: I'll check with Danielle

packages/components/inline-message/src/inline-message.scss
```diff
@@ -46,147 +50,94 @@ $variants: info, success, warning, danger;
...
-  block-size: var(--_header-block-size);
+  block-size: 1lh;
   display: inline-flex;
-  fill: var(--_fill);
 }
 
 [part='title'] {
-  font: var(--_title-font);
+  align-self: start;
+  font-size: 1.2em;
```
Diaan, line 125: should this not be a token?

packages/components/inline-message/src/inline-message.ts
```diff
@@ -77,16 +70,48 @@ export class InlineMessage extends ScopedElementsMixin(LitElement) {
       case 'warning':
         return 'octagon-exclamation-solid';
       case 'danger':
-        return 'triangle-exclamation-solid';
+        return 'diamond-exclamation-solid';
```
Diaan, line 76: Why is this changed? Didn't we discuss this and decide that a triangle is a more natural warning sign shape? Since that is also used in road signs?

Diaan (approved):

packages/components/inline-message/src/inline-message.ts
jpzwarte: Yes, this behavior is as designed. We need to document this properly.

packages/components/inline-message/src/inline-message.scss
jpzwarte: Yes, but we don't have typescale tokens *yet*. We need those; Roald is aware & agrees.

packages/components/inline-message/src/inline-message.ts
jpzwarte: So i'm resolving this as this is a choice by the designers.

PR #1831, opened 2025-03-07
Fix transparent backgrounds due to obsolete tokens

jpzwarte:


packages/components/panel/src/panel.scss
```diff
@@ -14,11 +14,11 @@
 }
 
 :host([elevation='raised']) {
-  background: var(--sl-elevation-surface-raised-alternative-idle);
```
anna-lach, line 17: Can you please also update this token in the `All` story?

image

anna-lach (approved):

PR #1836, opened 2025-03-10
Dialog and message dialog improvements

### LINKED ISSUE #1554: User focus can escape Dialog when navigating with tab

When using keyboard or screen reader and navigating with tab, the modal dialog is not a closed environment. When using keyboard or screen reader, the focus in an element with the role “dialog” should be contained in the element. Though a bit counterintuitive, this also includes the browser elements. 

### 🤔 Expected Behavior?

When using keyboard or screen reader and navigating with tab, the focus stays inside the modal, and doesn’t “leak” outside of it to the browser elements or the web page below the modal.

### 😯 Current Behavior

The user can access the browser elements and the page below when a modal is open and using tab.

### 💁 Possible Solution

When in a modal the focus should be locked using the following method:
•	Find the first and last focusable element programmatically
•	When user has focus in the last element of the modal, if the user presses the “Tab”-key, move focus to the first focusable element
•	When user has focus in the first element of the modal, if the user pressed “Shift + Tab” key combination, move focus to the last focusable element

### 🖥️ Steps to Reproduce

1. Open a page with the modal element (e.g. [Dialog Basic](Open https://storybook.sanomalearning.design/?path=/story/overlay-dialog--basic)
2. Use the tab key (or shift+tab to reverse) to navigate the element (optionally start a screen reader)
3. When in the first or last focusable element of the modal, keep pressing tab (or shift-tab) to escape the modal

Chrome

### If other, please specify.

_No response_

Windows 10

### 👤 Your name

Valtteri Järvinen

### 🧢 Your product/team

External

### 🎨 Your Theme(s)

Sanoma Learning

### Tasks
- [ ] https://github.com/sl-design-system/components/issues/1555
- [ ] https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#accessibility
- [ ] https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/
- [ ] update statement in issue
- [ ] add deviations in wcag messege dialog page and explain why we don't follow the 'wcag rules'
- [x] Discuss with Valterri in upcoming meeting
- [x] implement focus trap within dialog

jpzwarte:
Fixes #1554 and #1859.


.changeset/wise-poets-turn.md
```diff
@@ -0,0 +1,26 @@
...
+
+This was done to make it easier to position the actions at the top of the dialog on mobile devices. On mobile, primary actions are positioned on both sides of the title, and secondary actions are positioned at the bottom of the body of the dialog. On desktop, the secondary actions are automatically positioned at the bottom left of the dialog.
+
+This change also includes new `renderPrimaryActions()` and `renderSecondaryActions()` methods that can be overridden by developers who need to customize the rendering of the dialog.
+
+The dialog automatically sets the `fill` property on the primary action buttons, depending on the size of the viewport.
+
+- Make the dialog header and footer behave as sticky when scrolling
+
+- Animate the dialog using only CSS
```
anna-lach, line 17: It looks like the animation of the dialog (even on opening the dialog) is not working in Safari and Firefox.

packages/components/message-dialog/src/message-dialog.scss
```diff
@@ -1,39 +1,91 @@
...
-    }
+dialog {
+  background: var(--sl-elevation-surface-raised-default);
+  border: 0;
+  border-radius: var(--sl-size-borderRadius-default);
+  box-sizing: border-box;
+  color: var(--sl-color-foreground-plain);
+  flex-direction: column;
+  gap: var(--sl-size-200);
+  inline-size: 30rem;
```
anna-lach, line 13: Can we use a token here?

packages/components/message-dialog/src/message-dialog.scss
```diff
@@ -1,39 +1,91 @@
...
+  dialog,
+  dialog::backdrop {
+    transition: 0.4s ease-in-out;
+    transition-behavior: allow-discrete;
+    transition-property: display, opacity, overlay, scale;
 
-    &::part(footer-bar) {
-      --sl-button-bar-vertical: var(--sl-OFF);
+    @supports not (overlay: auto) {
+      transition-duration: 0s;
```
anna-lach, line 63: Or maybe just `0`?

packages/tokens/src/bingel-dc/light-new.json
```diff
@@ -1549,25 +1549,25 @@
...
-    "plain": {
-      "value": "{color.background.accent.grey.subtle}",
-      "type": "color",
-      "description": "Use for skeleton loading states."
     },
-    "subtle": {
-      "value": "{color.background.accent.grey.subtlest}",
-      "type": "color",
-      "description": "Use for the pulse or shimmer effect in skeleton loading states."
+    "blanket": {
```
anna-lach, line 1553: Can you please also add changeset for tokens update?

.changeset/wise-poets-turn.md
jpzwarte: Fixed; now only the exit animation is "disabled" when `overlay` is not supported.

packages/components/message-dialog/src/message-dialog.scss
jpzwarte: Nope, then it would be unitless. Linting warns about that.

packages/tokens/src/bingel-dc/light-new.json
jpzwarte: Done

packages/components/dialog/src/dialog.stories.ts
```diff
@@ -147,40 +155,56 @@ export const Lazy: Story = {
 };
 
 export const Mobile: Story = {
-  ...Basic,
   parameters: {
     viewport: {
-      defaultViewport: 'iphone13'
+      defaultViewport: 'iphone5'
```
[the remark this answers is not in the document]
jpzwarte: It's only the viewport. And iphone5 is the smallest (and fits easiest in storybook :)) From the component's pov there is no difference between an iphone13 or iphone5.

Diaan:
The dialog in the website is missing the button:
image

Diaan:
The focus ring is cropped:
image

Diaan:
In other examples where we have just a button here we use a primary variant
image

anna-lach (approved):

PR #1837, opened 2025-03-10
Updated selected color etsl (from teal to blue)

DanielleRameau:


jpzwarte (approved):
Can you add a changeset with a `patch` for the `@sl-design-system/editorial-suite` package?

PR #1840, opened 2025-03-11
Fix lifecycle loop warnings in the console for `<sl-number-field>` and `<sl-search-field>`

jpzwarte:


packages/components/number-field/src/number-field.ts
```diff
@@ -108,23 +111,21 @@ export class NumberField extends LocaleMixin(TextField) {
       this.input.min = this.min.toString();
     }
 
-    if (changes.has('locale') || changes.has('formatOptions')) {
+    if (changes.has('formatOptions') || changes.has('locale')) {
```
anna-lach, line 114: When I open the story
https://kind-plant-088739303-1840.westeurope.1.azurestaticapps.net/?path=/story/form-number-field--format-currency
and there is a value like `€9.90` in the number field:

image

and I change the locale in options e.g. to `de` the value is being removed from the number field 
and when I type in a value, the currency symbol is not being added, the value is not being formatted there and the value is invalid (on `main` it works fine)
image

anna-lach: Also with the percentage, when I type in sth the value is not being formatted and the `%` is not added there
https://kind-plant-088739303-1840.westeurope.1.azurestaticapps.net/?path=/story/form-number-field--format-percent&args=locale:de

anna-lach: same with unit
https://kind-plant-088739303-1840.westeurope.1.azurestaticapps.net/?path=/story/form-number-field--format-unit&args=locale:de

anna-lach (changes requested):

anna-lach (approved):

Diaan (approved):

PR #1856, opened 2025-03-17
Fix/1843 [tabs] tab using btn more does not trigger routerlink

anna-lach:
Fixes triggering tabs with links from `sl-menu-button`.


packages/components/tabs/src/tab-group.ts
```diff
@@ -277,7 +277,11 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
                   <sl-icon name="ellipsis" slot="button"></sl-icon>
                   ${this.menuItems?.map(
                     menuItem => html`
-                      <sl-menu-item @click=${() => this.#onMenuItemClick(menuItem.tab)} ?disabled=${menuItem.disabled}>
+                      <sl-menu-item
+                        @keydown=${this.#onKeydown}
```
jpzwarte, line 281: Why the keydown event handler? I thought menu-item already does `this.click()` if you're using the keyboard?

jpzwarte (approved):

packages/components/tabs/src/tab-group.ts
anna-lach: It was not working properly with link.

jpzwarte: Just so i understand: using the keyboard with menu-item and pressing enter or space, *doesn't* trigger the click event?

anna-lach: My bad, you're right, keydown is unnecessary here in the menu-item :)

Diaan (approved):

PR #1882, opened 2025-03-21
Fix/1855 sl checkbox enters an infite loop if the value is null

anna-lach:


packages/components/checkbox/src/checkbox-group.ts
```diff
@@ -137,10 +137,14 @@ export class CheckboxGroup<T = any> extends FormControlMixin(LitElement) {
     if (changes.has('value')) {
       this.#observer.disconnect();
       this.boxes?.forEach((box, index) => {
-        if (box.value !== undefined) {
+        if (box.value != null) {
```
jpzwarte, line 140: Is it possible to write a unit test so we don't regress in the future?

jpzwarte (approved):

PR #1905, opened 2025-03-26
Feature/1839 refactoring popover

Diaan:


packages/components/grid/src/filter.scss
```diff
@@ -4,6 +4,7 @@
   --_icon-color: currentcolor;
 
   display: block;
+  font-weight: normal; // when using 'inherit' it inherits from the table header, which is bold
```
anna-lach, line 7: Maybe it should be a token `--sl-text-typeset-fontWeight-regular`?

packages/components/popover/src/popover.scss
```diff
@@ -83,4 +84,5 @@
   inline-size: var(--_arrow-inline-size);
   position: absolute;
   stroke: var(--sl-color-border-plain);
+  stroke-width: 2px;
```
anna-lach, line 87: Can you please use a token?

packages/components/tooltip/src/tooltip.scss
```diff
@@ -1,28 +1,21 @@
 :host {
-  --_arrow-block-size: 6px;
-  --_arrow-inline-size: 16px;
-  --_background: var(--sl-color-tooltip-background);
-  --_border-radius: var(--sl-border-radius-default);
-  --_border: 0;
-  --_color: var(--sl-color-tooltip-foreground);
-  --_font: var(--sl-text-tooltip-text-tip);
+  --_arrow-block-size: 5px;
```
anna-lach, line 2: Can you please use a token? :)

packages/components/tooltip/src/tooltip.scss
```diff
@@ -1,28 +1,21 @@
...
+  --_arrow-inline-size: 10px;
```
anna-lach, line 3: Can you please use a token?

packages/components/tooltip/src/tooltip.scss
Diaan: There is no token for 5px 🤷

.changeset/rude-jobs-share.md
```diff
@@ -0,0 +1,21 @@
...
+'@sl-design-system/editorial-suite': patch
+'@sl-design-system/my-digital-book': patch
+'@sl-design-system/sanoma-learning': patch
+'@sl-design-system/myvanin': patch
+'@sl-design-system/popover': patch
+'@sl-design-system/tooltip': patch
+'@sl-design-system/itslearning': patch
+'@sl-design-system/bingel-int': patch
+'@sl-design-system/bingel-dc': patch
+'@sl-design-system/grid': patch
```
jpzwarte, line 12: Grid should be excluded here? Since it has its own changeset?

packages/components/grid/src/filter.scss
jpzwarte: Perhaps use `initial` instead of `normal`? (not 100% sure that is correct)

packages/components/grid/src/filter.scss
```diff
@@ -27,24 +28,24 @@ sl-popover {
 
 header {
   align-items: center;
+  background: var(--sl-elevation-surface-raised-default);
   border-block-end: 1px solid var(--sl-color-elevation-border-raised);
   display: flex;
+  font-weight: var(--sl-font-weight-semibold);
   gap: 0.5rem;
```
jpzwarte, line 37: Please use `var(--sl-size-100)`

packages/components/grid/src/filter.scss
```diff
@@ -27,24 +28,24 @@ sl-popover {
...
+  inset-block-start: calc(var(--sl-size-300) * -1);
   justify-content: space-between;
-  padding-block: 0.25rem;
-  padding-inline: 0.5rem;
+  margin-block: calc(var(--sl-size-300) * -1) var(--sl-size-200);
+  padding-block-start: var(--sl-size-300);
+  position: sticky;
+  z-index: 1;
```
jpzwarte, line 43: I've started documenting the value of `z-index`, so it's clear why it is a certain value.

packages/components/tooltip/src/tooltip.scss
anna-lach: Maybe it would be helpful to have such?

packages/components/popover/src/popover.scss
```diff
@@ -1,18 +1,18 @@
...
-  --_arrow-inline-size: var(--sl-size-200);
+  --_arrow-block-size: var(--sl-size-125);
+  --_arrow-inline-size: var(--sl-size-250);
 
   background: var(--sl-elevation-surface-raised-default);
   border: var(--sl-size-borderWidth-default) solid var(--sl-color-border-plain);
   border-radius: var(--sl-size-borderRadius-default);
   box-shadow: var(--sl-elevation-shadow-overlay);
-  box-sizing: border-box;
+  box-sizing: content-box;
```
jpzwarte, line 9: Isn't `content-box` the default?

packages/components/popover/src/popover.scss
```diff
@@ -1,18 +1,18 @@
...
   color: var(--sl-color-foreground-plain);
   margin: 0;
   max-inline-size: var(--sl-popover-max-inline-size, 80vw);
   opacity: 0;
   overflow: visible;
-  padding: var(--sl-size-100) var(--sl-size-150);
+  padding: 0;
```
jpzwarte, line 15: Isn't this the default?

packages/components/popover/src/popover.scss
jpzwarte: `var(--sl-size-025)`

packages/components/tooltip/src/tooltip.ts
```diff
@@ -118,16 +118,28 @@ export class Tooltip extends LitElement {
...
       this.hidePopover();
     }
   };
 
   #onShow = ({ target }: Event): void => {
     if (this.#matchesAnchor(target as HTMLElement)) {
       this.anchorElement = target as HTMLElement;
+
       this.showPopover();
+      setTimeout(() => {
```
jpzwarte, line 139: Since this is only `10ms`, is `requestAnimationFrame()` also an option?

jpzwarte (approved):

packages/components/popover/src/popover.scss
Diaan: It appears there is a default padding on popovers from the user agent stylesheet:
image

anna-lach (approved):

PR #1909, opened 2025-03-27
Log a warning when you use `<sl-select>` without having `<sl-option>` registered

### LINKED ISSUE #1894: [SELECT][LISTBOX] Select requires listbox dependency to work

Select options are broken when @sl-design-system/listbox is not installed

### 🤔 Expected Behavior?

@sl-design-system/select should either:
- work standalone without listbox
- throw error during npm install that it requires listbox

It would be nice to mentiont that within the documentation

### 😯 Current Behavior

When you open a select options error is thrown within the console and you can see that something is wrong.

### 💁 Possible Solution

_No response_

### 🖥️ Steps to Reproduce

1. Checkout on angular-demo main
2. remove  "@sl-design-system/listbox" "^0.1.0"
3. add schemas: [CUSTOM_ELEMENTS_SCHEMA], to form.component.ts

Image

Chrome

### If other, please specify.

_No response_

Windows

### 👤 Your name

Jakub Wachowiak

### 🧢 Your product/team

ETSL

### 🎨 Your Theme(s)

Editorial Suite

jpzwarte:
Fixed #1894

### DIFF
--- a/packages/components/select/src/select.ts
+++ b/packages/components/select/src/select.ts
@@ -402,6 +402,8 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
   }
 
   #onSlotchange(): void {
+    this.#verifyRegisteredListboxElements();
+
     this.options.forEach(option => option.setAttribute('aria-selected', 'false'));
 
     if (this.value !== undefined && this.value !== null) {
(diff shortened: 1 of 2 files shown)


anna-lach (approved):

packages/components/select/src/select.ts
```diff
@@ -481,4 +483,14 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
 
     this.updateValidity();
   }
+
+  #verifyRegisteredListboxElements(): void {
+    const option = this.querySelector('sl-option');
+
+    if (option && !(option instanceof Option)) {
```
Diaan, line 490: We have more components where we send warnings if the dev is doing something wrong/is missing something, but we only send those when in dev mode. Is that something we want here as well?

Diaan (approved):

packages/components/select/src/select.ts
jpzwarte: I think in this case it is fine to have it in production as well. It is something that really should be fixed, even in production.

PR #1926, opened 2025-03-31
Fix/1892 sl grid filter lowercase in filter

Diaan:


packages/components/grid/src/stories/filtering.stories.ts
```diff
@@ -90,8 +92,16 @@ export const Custom: Story = {
...
         <sl-grid-column path="lastName"></sl-grid-column>
-        <sl-grid-filter-column path="status"></sl-grid-filter-column>
-        <sl-grid-filter-column path="membership"></sl-grid-filter-column>
+        <sl-grid-filter-column path="membership" filter-label="DUO membership status"></sl-grid-filter-column>
+        <sl-grid-filter-column
+          path="status"
+          .header=${html`
+            Status
+            <sl-icon aria-describedby="tooltip" name="info"></sl-icon>
+            <sl-tooltip id="tooltip">Status is based on current calendar activity</sl-tooltip>
```
anna-lach, line 101: I know this is just an example :) but it looks like the tooltip is not visible on hover/focus?

anna-lach (approved):

packages/components/grid/src/filter.ts
```diff
@@ -129,6 +132,21 @@ export class GridFilter<T = any> extends ScopedElementsMixin(LitElement) {
...
+    let header = this.column.header;
+    if (typeof header !== 'string' && header !== undefined) {
+      const div = document.createElement('div');
+      div.innerHTML = (header as unknown as TemplateResult).strings[0];
+      const textNodes = Array.from(div.childNodes)
+        .filter(node => node.nodeType !== Node.ELEMENT_NODE && node.textContent?.trim())
+        .map(node => node.textContent?.trim());
+      header = textNodes.join(' ');
+    }
+    return (
```
jpzwarte, line 145: Can you change this so it returns early if there is a filter label? And do the "hard work" otherwise?

packages/components/grid/src/filter.ts
```diff
@@ -137,9 +155,7 @@ export class GridFilter<T = any> extends ScopedElementsMixin(LitElement) {
       </sl-button>
       <sl-popover anchor="anchor" position="bottom">
         <header>
-          <h1 id="title">
-            ${msg('Filter by')} <span>${this.column.header?.toString() || getNameByPath(this.column.path)}</span>
-          </h1>
+          <h1 id="title">${msg('Filter by')} <span>${this.getFilterHeaderValue()}</span></h1>
```
jpzwarte, line 158: I'm not sure if we shouldn't include the filter header value in the `msg()` call. Other languages may have a different order?

jpzwarte (approved):

packages/components/grid/src/filter.ts
Diaan: There might even be languages where you'd have the name of the filter at the beginning of the sentence 😬 Causing a whole new logic for when to have upper or lower case...

packages/components/grid/src/stories/filtering.stories.ts
Diaan: I guess that doesn't work when filtering is active... I removes/overwrites all other interactions

Diaan: I guess that doesn't work when filtering is active... I removes/overwrites all other interactions

PR #1934, opened 2025-04-04
color update in sl light

Diaan:


anna-lach (approved):
Can you please add a changeset? :)

PR #1957, opened 2025-04-10
Minor improvements to grid filtering

### LINKED ISSUE #1949: [GRID] Filter column options are not reset to all available options when filtered at datasource

When there is a filter set on the datasource, ...

for example: 
`dataSource.addFilter('studyName', 'studyName', ['L BRUG 3', 'L HAVO 4']);`

... only these two studyNames are available in the filter.

After deselecting both I would expect to see all the available filter options (studies). 

 

N.B. Example filtering datasource Storybook is broken.. (no filter for status or membership)
https://storybook.sanomalearning.design/?path=/story/grid-filtering--filtered-data-source

### 🤔 Expected Behavior?

After deselecting both I would expect to see all the available filter options (studies). Or at least have a hook or an event to clear the filter myself.. 

Image

### 😯 Current Behavior

-

### 💁 Possible Solution

_No response_

### 🖥️ Steps to Reproduce

`dataSource.addFilter('studyName', 'studyName', ['L BRUG 3', 'L HAVO 4']);`

Chrome

### If other, please specify.

_No response_

MacOS

### 👤 Your name

Ronald van der Veen

### 🧢 Your product/team

Magister

### 🎨 Your Theme(s)

Magister

jpzwarte:
- Add an `originalItems` property to `ListDataSource` so it uses the unfiltered list to determine the filter options
- Add a `labelPath` property to `<sl-grid-filter-column>` so you can use a different path for displaying the value in the column

Fixes #1949


packages/tokens/src/sanoma-learning/light-new.json
```diff
@@ -1538,12 +1538,12 @@
       },
       "interactive": {
         "bold": {
-          "value": "{color.palette.grey.800}",
+          "value": "{color.background.neutral.interactive.bold}",
```
anna-lach, line 1541: Can you please add a changeset for this token update as well?

anna-lach (approved):

PR #1972, opened 2025-04-15
Fix/1820 bug tag list problems with focusing removable tags in the tag list

anna-lach:


packages/components/tag/src/tag-list.ts
```diff
@@ -238,6 +238,8 @@ export class TagList extends ScopedElementsMixin(LitElement) {
     // Calculate the stack size based on the visibility of the tags
     this.stackSize = this.tags.reduce((acc, tag) => (tag.style.display === 'none' ? acc + 1 : acc), 0);
     this.stack.style.display = this.stackSize === 0 ? 'none' : '';
+    const stackTag = this.stack.querySelector('sl-tag') as Tag;
```
jpzwarte, line 241: The `as Tag` should not be necessary

jpzwarte (approved):

Diaan (approved):

PR #1975, opened 2025-04-16
Improve grid headers

jpzwarte:


Diaan (approved):

Diaan:
The menu button looks a bit weird when it has the focus: 
image
https://kind-plant-088739303-1975.westeurope.1.azurestaticapps.net/?path=/story/grid-basics--menu-button

Lazy load and Skeleton are broken
https://kind-plant-088739303-1975.westeurope.1.azurestaticapps.net/?path=/story/grid-basics--lazy-load
https://kind-plant-088739303-1975.westeurope.1.azurestaticapps.net/?path=/story/grid-basics--skeleton

Dragging and dropping doesn't have any effect on the order (doesn't work on main either, so not sure it's part of this story)
https://kind-plant-088739303-1975.westeurope.1.azurestaticapps.net/?path=/story/grid-drag-and-drop--between

Filtering with datasource seems work on id, not on the displayed text:
image

Filtering with selection is broken (on main as well)
https://kind-plant-088739303-1975.westeurope.1.azurestaticapps.net/?path=/story/grid-filtering--filtering-with-selection

Is the alignment correct when not every column has a filter option?
image

The width of the table doesn't really work anymore :D 
https://kind-plant-088739303-1975.westeurope.1.azurestaticapps.net/?path=/story/grid-filtering--empty-values

Grouping is broken: it only adds 1 item in each group, and puts the rest in the last group:
https://kind-plant-088739303-1975.westeurope.1.azurestaticapps.net/?path=/story/grid-grouping--basic
https://github.com/user-attachments/assets/a8ac844e-51dc-4431-bd1c-1fe2db1db375

Horizontal scrolling has 2 scrollbars:
image
https://kind-plant-088739303-1975.westeurope.1.azurestaticapps.net/?path=/story/grid-scrolling--horizontal

Not sure a tri-state toggle button is very good accessibility wise. I don't think the change in label is being announced, so it will be a surprise what the sorting will be when you press it a second time.
https://kind-plant-088739303-1975.westeurope.1.azurestaticapps.net/?path=/story/grid-sorting--basic

jpzwarte:
I've fixed everything except:
- grouping (will look at this in another PR)
- a11y of sort header; afaict there is nothing to improve there

packages/components/data-source/src/fetch-list-data-source.ts
```diff
@@ -89,9 +89,12 @@ export class FetchListDataSource<T = any> extends ListDataSource<T> {
     if (options.placeholder) {
       this.placeholder = options.placeholder;
     }
+
+    // Initialize the items array, but do not emit an event yet
+    this.update(false);
   }
 
-  update(): void {
+  update(emitEvent = true): void {
```
anna-lach, line 97: So when we are filtering and there is a pagination added, we don't move to the first page anymore?
image

packages/components/grid/package.json
```diff
@@ -44,10 +45,12 @@
     "@sl-design-system/icon": "^1.2.0",
     "@sl-design-system/popover": "^1.2.2",
     "@sl-design-system/scrollbar": "^0.1.1",
```
anna-lach, line 47: It looks like the `listbox` dependency is missing?

packages/components/grid/src/grid.scss
```diff
@@ -447,3 +485,47 @@ tfoot {
...
+    padding-block: var(--sl-size-050);
+  }
+
+  slot[name='bulk-actions']::slotted(sl-button:first-of-type) {
+    margin-inline-start: var(--sl-size-600);
+  }
+}
+
+@position-try --viewport-block-end {
+  inset-block: auto 1rem;
```
anna-lach, line 528: Maybe we can use a token instead of `1rem` here?

packages/components/grid/src/grid.scss
```diff
@@ -447,3 +485,47 @@ tfoot {
...
 }
+
+[part='bulk-actions'] {
+  align-items: center;
+  background: var(--sl-color-background-selected-bold);
+  border: 0;
+  border-radius: var(--sl-size-borderRadius-default);
+  box-shadow: var(--sl-elevation-shadow-overlay);
+  color: var(--sl-color-foreground-inverted-plain);
+  inset-block-start: calc(anchor(bottom) + 1rem);
```
anna-lach, line 496: Maybe we can use a token instead of `1rem` here?

packages/components/grid/src/grid.ts
```diff
@@ -326,34 +344,42 @@ export class Grid<T = any> extends ScopedElementsMixin(LitElement) {
...
                 </tr>
-              `
-            : nothing}
-        </tfoot>
+              </tfoot>
+            `
+          : nothing}
       </table>
 
+      <div part="bulk-actions" popover="manual">
```
anna-lach, line 376: When I selected few rows I couldn't focus anything inside bulk-actions, I couldn't get there with keyboard to focus anything inside.

image

packages/components/grid/src/stories/selection.stories.ts
```diff
@@ -19,6 +24,8 @@ export default {
   }
 };
 
+Icon.register(faCopy, faTrash);
+
 export const ClickableRow: Story = {
```
anna-lach, line 29: When we are able to click and select a row with mouse, I guess we should be able to select a row also with keyboard? Right now it's not possible

image

anna-lach (approved):

packages/components/grid/package.json
jpzwarte: Fixed

packages/components/data-source/src/fetch-list-data-source.ts
jpzwarte: This is happening on `main` as well; i've created #2021 for this

packages/components/grid/src/grid.ts
jpzwarte: So the `focus-visible` is broken in the SL theme. If you switch to the Magister theme it works. Once you've tabbed through all the checkboxes, focus will go to the buttons in the popover.

PR #1977, opened 2025-04-16
Feature/1976 color property default progress bar

Diaan:


.changeset/free-rooms-type.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sl-design-system/progress-bar': patch
+---
+
+Adds color property for changing the color of the default variant. These colors don't add any semantical meaning or extra labels or icons.
```
jpzwarte, line 5: Not sure "semantical" is an actual work? "semantic" instead?

jpzwarte (approved):

PR #1991, opened 2025-04-24
Fix/1843 (tabs) tab using btn more does not trigger router link

anna-lach:


Diaan (approved):

.changeset/few-maps-design.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sl-design-system/tabs': patch
+---
+
+Fix selecting tab when clicking menu item.
```
jpzwarte, line 5: Can you elaborate a bit more?

PR #1995, opened 2025-04-25
Add new `eslint-plugin-slds` package for linting rules

jpzwarte:
This adds a new `@sl-design-system/eslint-plugin-slds` package for SL Design System specific rules:
- `slds/button-has-label`: checks if an `<sl-button>` has a label, either via text content inside the button, or via `aria-label` or `aria-labelledby`

CleanShot 2025-04-25 at 14 21 06@2x

- `slds/multiline-html-template`: this rule makes sure that multiline tagged templates have a newline after/before the opening/closing backtick; this ensures that prettier formats the template properly

CleanShot 2025-04-28 at 09 59 04@2x

- `slds/singleline-html-template-trimmed`: this rule trims whitespace from singleline templates; sometimes when prettier formats code, it will collapse a multiline template into a single line, but it won't trim the whitespace. This rule does.

CleanShot 2025-04-28 at 10 01 07@2x


packages/components/card/src/card.spec.ts
```diff
@@ -51,14 +53,16 @@ describe('sl-card', () => {
...
-        html`<sl-card>
+      el = await fixture(html`
+        <sl-card>
           <h2>${title}</h2>
           <h3 slot="header">${subHeader}</h3>
           <p slot="body">${bodyCopy}</p>
-          <sl-button icon-only slot="actions" fill="ghost"><sl-icon name="ellipsis"></sl-icon></sl-button>
-        </sl-card>`
-      );
+          <sl-button aria-label="Actions" icon-only slot="actions" fill="ghost">
```
anna-lach, line 61: The `icon-only` attribute is not necessary, right? It's added automatically.

packages/components/card/src/card.spec.ts
```diff
@@ -72,15 +76,17 @@ describe('sl-card', () => {
...
+      el = await fixture(html`
+        <sl-card style="--sl-card-horizontal-breakpoint:1800px">
           <img slot="media" src=${image} />
           <h2>${title}</h2>
           <h3 slot="header">${subHeader}</h3>
           <p slot="body">${bodyCopy}</p>
-          <sl-button icon-only slot="actions" fill="ghost"><sl-icon name="ellipsis"></sl-icon></sl-button>
-        </sl-card>`
-      );
+          <sl-button aria-label="Actions" icon-only slot="actions" fill="ghost">
```
anna-lach, line 85: The `icon-only` attribute is not necessary, right? It's added automatically.

packages/components/card/src/card.spec.ts
```diff
@@ -12,16 +12,18 @@ describe('sl-card', () => {
...
+        <sl-card>
           <img slot="media" src=${image} />
           <sl-icon name="pinata" slot="icon"></sl-icon>
           <h2>${title}</h2>
           <h3 slot="header">${subHeader}</h3>
           <p slot="body">${bodyCopy}</p>
-          <sl-button icon-only slot="actions" fill="ghost"><sl-icon name="ellipsis"></sl-icon></sl-button>
-        </sl-card>`
-      );
+          <sl-button aria-label="Actions" icon-only slot="actions" fill="ghost">
```
anna-lach, line 22: The `icon-only` attribute is not necessary, right? It's added automatically.

anna-lach (approved):

packages/components/calendar/src/select-day.ts
```diff
@@ -131,11 +131,16 @@ export class SelectDay extends LocaleMixin(ScopedElementsMixin(LitElement)) {
   override render(): TemplateResult {
     return html`
       <div part="header">
-        <sl-button @click=${this.#onToggleMonthSelect} class="current-month" fill="link">
+        <sl-button
+          @click=${this.#onToggleMonthSelect}
+          aria-label="Change current month"
```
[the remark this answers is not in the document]
jpzwarte: Hmm, good point. Perhaps the rule should only check specifically for `<sl-icon>`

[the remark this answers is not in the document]
jpzwarte: Fixed

PR #1997, opened 2025-04-25
Does this fix some chromatic issues?

Diaan:


packages/components/avatar/src/avatar.scss
```diff
@@ -112,8 +112,8 @@
   text-align: center;
 
   [part='name'] {
+    overflow-wrap: anywhere;
```
jpzwarte, line 115: Can you add a changeset for this?

anna-lach (approved):

PR #2013, opened 2025-05-01
fix/1999 Add possibility to change background color for tab bar

Diaan:
use mask-image with inherited bg instead of background gradient with explicit colour

### DIFF
--- a/packages/components/tabs/src/tab-group.scss
+++ b/packages/components/tabs/src/tab-group.scss
@@ -61,13 +61,17 @@
   }
 
   .fade-start {
-    background: linear-gradient(0deg, transparent, var(--_background));
+    background: inherit;
     inset-block-start: 0;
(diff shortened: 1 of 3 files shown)


packages/components/tabs/src/tab-group.scss
```diff
@@ -61,13 +61,17 @@
   }
 
   .fade-start {
-    background: linear-gradient(0deg, transparent, var(--_background));
+    background: inherit;
     inset-block-start: 0;
+    /* stylelint-disable-next-line color-no-hex */
+    mask-image: linear-gradient(0deg, transparent, #000);
```
jpzwarte, line 67: Replace `#000` with a variable, so you don't need to disable the stylelint rule everywhere?

jpzwarte (approved):

PR #2017, opened 2025-05-05
Fix/1477 shared rovingtabindexcontroller has trouble with both light dark elements

anna-lach:


jpzwarte (approved):
Can you add a specific test for the "tabindex trouble" it previously had as a regression test?

Diaan (approved):

PR #2067, opened 2025-05-28
Fix/1875 combobox required validation

Diaan:


packages/components/combobox/src/combobox.ts
```diff
@@ -256,7 +262,8 @@ export class Combobox<T = any, U = T> extends FormControlMixin(ScopedElementsMix
 
     this.#observer.observe(this);
 
-    this.setFormControlElement(this.input);
+    this.internals.ariaRequired = this.required ? 'true' : 'false';
```
jpzwarte, line 265: Should this be done here? Or in `willUpdate`/`updated`?

jpzwarte (approved):
Can you add a unit test for it?

PR #2081, opened 2025-06-11
Various tool-bar fixes

jpzwarte:


Diaan:
Toolbars with overflow have some weird scrolling behaviour:
image
It's like the container that hides the button still has overflow that causes the parent to grow bigger
https://jolly-stone-0f7fffc03-2081.westeurope.6.azurestaticapps.net/?path=/story/actions-tool-bar--overflow

Diaan (approved):

PR #2086, opened 2025-06-13
Various form fixes

### LINKED ISSUE #1230: Dialog & form issues
I tried to build a design of a form using just SLDS. Below you can see the original and design and what I built, along with what I encountered.

Screenshot 2024-04-30 at 09 44 26

CleanShot 2024-05-05 at 11 27 17@2x

1. The default slot has `overflow: auto;`, which clips the focus outline
2. The border radius is much less than in the original design
3. There is much more spacing in the dialog vs the original design
4. This is an `<sl-text-field>` with the euro sign in the `prefix` slot and `0,00` as `placeholder`; looks good, but this really needs to be an `<sl-number-field>` imo
5. I added this to the rental period `<sl-form-field>`, but not sure that is the best approach. Since `<sl-form-field>` is meant for a single form control, this feels weird; need to investigate what the best approach is
6. The optional mark works in this case, but would break if used on the rental amount  field. Should automatically use a less wide version if needed?

### LINKED ISSUE #2128: [NUMBER FIELD] sl-change and sl-validatie events do not fire

In a SL number field `@sl-change` and `@sl-validate` events doesn't work

### 🤔 Expected Behavior?

`@sl-change` and `@sl-validate` events should work

### 😯 Current Behavior

nothing happens

### 💁 Possible Solution

_No response_

### 🖥️ Steps to Reproduce

```
        <sl-form-field>
          <sl-number-field @sl-change=${(event: CustomEvent<string>) => {
              console.log(event.detail);
            }}>
          </sl-number-field>
        </sl-form-field>
```

Chrome

### If other, please specify.

_No response_

WSL

### 👤 Your name

Frans de Boer

### 🧢 Your product/team

Magister

### 🎨 Your Theme(s)

Magister

### LINKED ISSUE #2483: [Number field] Setting a `number` via the form causes errors

The numberfield has a `value` and a `rawValue` property; the `value` is the value from the input; a string, and the parsed number is the `rawValue`
When you set the initial value, a `number` via the form, like so:
```typescript
this.#form.value.initialWeightingFactor = this.gradeColumnDetail.initialWeightingFactor ?? 0
``` 
you will get an error:
```
number-parser.js:13 Uncaught TypeError: value.replace is not a function
    at NumberParser.parse (number-parser.js:13:32)
    at set value (number-field.js:56:56)
```

because the `parse` function expects  a string, but now a number is given

### 🤔 Expected Behavior?

make component more robust; it should work when you put in an actual number as well

### 😯 Current Behavior

error

### 💁 Possible Solution

_No response_

### 🖥️ Steps to Reproduce

see description

Chrome

### If other, please specify.

_No response_

mac

### 👤 Your name

Diana

### 🧢 Your product/team

slds

### 🎨 Your Theme(s)

Sanoma Learning

jpzwarte:
Fixes #1230 
Fixes #2483 
Fixes #2128


packages/components/text-field/src/text-field.stories.ts
```diff
@@ -206,73 +205,6 @@ export const CustomAsyncValidity: Story = {
...
-        <custom-text-field
-          @sl-validate=${onValidate}
-          placeholder="Enter a DD-MM-YYYY value"
-          required
-        ></custom-text-field>
-      `;
-    }
-  }
-};
-
```
Diaan, line 275: Don't we support custom components anymore?

jpzwarte:
> When you edit a formatted number you get a validation error;

Fixed!

packages/components/number-field/src/number-parser.spec.ts
```diff
@@ -45,22 +53,8 @@ describe('NumberParser', () => {
       expect(parser.parse('$1,234.56')).to.equal(1234.56);
     });
 
-    it('should return undefined for invalid currency value', () => {
-      expect(parser.parse('$abc')).to.be.undefined;
-    });
-  });
-
-  describe('unit', () => {
```
anna-lach, line 53: So we don't want to test units anymore?

packages/components/text-field/src/text-field.ts
```diff
@@ -140,13 +140,13 @@ export class TextField<T extends { toString(): string } = string>
    */
   @property() type: 'email' | 'number' | 'tel' | 'text' | 'url' | 'password' = 'text';
```
anna-lach, line 141: Can you please also remove 'number' from here since we already have a number-field? :)

anna-lach (approved):

PR #2097, opened 2025-06-16
Don't hide on move to child nodes

Diaan:


packages/components/tooltip/src/tooltip.ts
```diff
@@ -118,13 +118,33 @@ export class Tooltip extends LitElement {
     );
   };
 
+  #getParentsUntil = (el: Element, selector: string) => {
```
anna-lach, line 121: Can you please also add a changeset?

packages/components/tooltip/src/tooltip.stories.ts
```diff
@@ -115,6 +116,30 @@ export const Shared: Story = {
   }
 };
 
+export const NestedChildren: Story = {
+  args: {
+    example: ({ message }) => html`
+      <div
```
anna-lach, line 122: That `div` is not focusable, so it should not have a tooltip, right? Or maybe make it focusable?

jpzwarte (approved):
Can you add a changeset plz?

packages/components/tooltip/src/tooltip.ts
```diff
@@ -118,13 +118,33 @@ export class Tooltip extends LitElement {
...
+    const parents: Element[] = [];
+    let _el = el?.parentNode as HTMLElement | null;
```
anna-lach, line 123: or maybe `let element = ...` ?

anna-lach (approved):

PR #2105, opened 2025-06-23
Fix/2100 form run time error reportvalidity

Diaan:


packages/components/form/src/form.ts
```diff
@@ -177,7 +177,7 @@ export class Form<T extends Record<string, any> = Record<string, any>> extends L
   }
 
   #onFormControl(event: SlFormControlEvent): void {
-    const control = event.target;
+    const control = event.composedPath()[0] as EventTarget & HTMLElement & FormControl;
```
jpzwarte, line 180: The type casting without any checks feels a little dangerous. Anyway you could improve this using `instanceof` perhaps?

jpzwarte (approved):

packages/components/form/src/form.ts
```diff
@@ -177,7 +177,10 @@ export class Form<T extends Record<string, any> = Record<string, any>> extends L
   }
 
   #onFormControl(event: SlFormControlEvent): void {
-    const control = event.target;
+    if (!(event.composedPath()[0] instanceof EventTarget) || !(event.composedPath()[0] instanceof HTMLElement)) {
```
anna-lach, line 180: I would add type guard below that method, sth like:

```
#isFormControl(element: unknown): element is HTMLElement & FormControl {
    return element instanceof HTMLElement && 'formControlElement' in element;
  }
  ```
  
  and here it could look like 
  ```suggestion
    if (
      !(event.composedPath()[0] instanceof EventTarget) ||
      !(event.composedPath()[0] instanceof HTMLElement) ||
      !this.#isFormControl(event.composedPath()[0])
    ) {
```

anna-lach (approved):

PR #2131, opened 2025-07-14
Fix/2122 date field does not remove required validation

anna-lach:


packages/components/date-field/src/date-field.ts
```diff
@@ -166,6 +166,29 @@ export class DateField extends LocaleMixin(FormControlMixin(ScopedElementsMixin(
 
     if (changes.has('value')) {
       this.input.value = this.value && this.#formatter ? this.#formatter.format(this.value) : '';
+      this.#onTextFieldChange(new CustomEvent('sl-change', { detail: this.value, bubbles: true, composed: true }));
```
jpzwarte, line 169: This is a little weird, calling an event handler like that.

packages/components/date-field/src/date-field.ts
```diff
@@ -166,6 +166,29 @@ export class DateField extends LocaleMixin(FormControlMixin(ScopedElementsMixin(
...
+    }
+
+    if (changes.has('showValid') || changes.has('showValidity')) {
+      if (this.renderRoot.querySelector('sl-text-field')) {
```
jpzwarte, line 173: We should probably cache the text field element in a class variable.

jpzwarte (approved):

PR #2140, opened 2025-07-24
Fix/2100 form run time error reportvalidity

Diaan:


Diaan:
Review was done on original PR but 1 forgot to push the latest commit before I merged it

PR #2481, opened 2025-08-25
refactor: update focus ring styles across multiple components

Diaan:


packages/components/button/src/button.scss
```diff
@@ -14,8 +14,8 @@
   font-weight: var(--sl-text-new-typeset-fontWeight-semiBold);
   gap: var(--sl-size-075);
   justify-content: center;
-  outline: transparent solid var(--sl-size-outlineWidth-default);
-  outline-offset: var(--sl-size-outlineOffset-default);
+  outline: transparent solid var(--sl-size-borderWidth-focusRing);
```
anna-lach, line 17: Maybe for the `inverted` buttons there should be a different color token used?

anna-lach (approved):

PR #2535, opened 2025-09-02
Add missing exports for theme packages

### LINKED ISSUE #2119: The path "**.css" is not exported by the package

When we try to install themes like:
import @sl-design-system/sanoma-learning/light.css it says the path is not exported by the package

Image

### 🤔 Expected Behavior?

It should allow you to import it directly from the file or probably just with: @sl-design-system/sanoma-learning

### 😯 Current Behavior

It doesn't let you import it in that way throwing an error:

Image

### 💁 Possible Solution

Exporting files in the package.json from the theme/package. I've seen that it's happening in sanoma-learning theme, however it might also happening in the rest of the themes.

### 🖥️ Steps to Reproduce

1. I created a new test angular project
2. Installed theme dep
3. Added to the angular.json file in the styles array

Then the error appears when I try to run it. It also happens if we try to import it in a html file. Actually the only possible thing is to import it through node_modules/@sl.... etc.

We actually use editorial-suite, bingel-dc, sanoma-learning themes.

Chrome

### If other, please specify.

_No response_

Windows

### 👤 Your name

Óscar Ledesma

### 🧢 Your product/team

Pokemon

### 🎨 Your Theme(s)

Editorial Suite

jpzwarte:
Fixes #2119

### DIFF
--- a/website/src/categories/getting-started/developers.md
+++ b/website/src/categories/getting-started/developers.md
@@ -111,6 +111,12 @@ The simplest way to include the theme is by including the theme stylesheet in yo
 
 </div>
 
+The theme packages also support package exports, meaning you can import the stylesheet directly from the package (if your build system supports it):
+
+```scss
+@import '@sl-design-system/sanoma-learning/light.css';
+```
+
 If your theme supports dark mode you can also use `dark.css`.
 
 If you need to have more control over theming, you can use the split files in the `css` and `scss` folders of the theme package. These folders contain split files for the base, light and dark theme. You can include these files in your application as you see fit.


.changeset/wicked-ducks-wink.md
```diff
@@ -0,0 +1,17 @@
...
+'@sl-design-system/clickedu': patch
+'@sl-design-system/magister': patch
+'@sl-design-system/kampus': patch
+'@sl-design-system/neon': patch
+'@sl-design-system/teas': patch
+'@sl-design-system/max': patch
+'@sl-design-system/tig': patch
+---
+
+Add missing package exports for CSS files
```
anna-lach, line 17: Can you please also update Getting started page for developers on the documentation website:
https://sanomalearning.design/categories/getting-started/developers/ ?

I mean this section:
https://sanomalearning.design/categories/getting-started/developers/#setup-a-theme
that styling can be used also by importing explicitly those files (not only from node_modules as it is described right now).

And also this section:
https://sanomalearning.design/categories/getting-started/developers/#how-do-i-use-the-dark-mode-of-the-theme

PR #2577, opened 2025-09-04
Avatar improvements

jpzwarte:


DanielleRameau:
font-size of heading isn't consistent in figma heading is 14px (md) in sm, md, lg and xl.
font-size in subheading should be 14px in XL, 2XL and 3XL (this is currently not in Figma the case but asked Roald to change this)

Diaan (approved):

packages/components/avatar/src/avatar.ts
```diff
@@ -49,15 +52,27 @@ export class Avatar extends ScopedElementsMixin(LitElement) {
...
 
-  /** The clip-path cutout for the badge. */
+  /** @internal The clip-path cutout for the badge. */
   @state() clipPath?: string;
 
+  /**
+   * The color of the avatar.
+   * @default grey
+   */
+  @property({ reflect: true }) color?: AvatarColor;
```
anna-lach, line 62: This is probably not the right place, but it looks like colors are not described on the documentation website (in the `usage` tab)?
 https://salmon-hill-037b61403-2577.westeurope.6.azurestaticapps.net/categories/components/avatar/usage#figma-options
 
image

anna-lach: And probably `badge` 'Figma options' are not updated as well?

https://salmon-hill-037b61403-2577.westeurope.6.azurestaticapps.net/categories/components/badge/usage

PR #2597, opened 2025-09-10
Fix the ability to release new theme packages

jpzwarte:
- Stop automatically importing tokens during the build
- Store the imported tokens in `packages/tokens/src` folder
- Build the themes from `packages/tokens/src`


packages/tokens/src/placeholder/light-new.json
```diff
@@ -0,0 +1,1684 @@
+{
+	"placeholder": {
```
anna-lach, line 2: Do we want to keep this `placeholder` structure? Is it a proper naming?

.changeset/tall-carrots-win.md
```diff
@@ -0,0 +1,17 @@
+---
+'@sl-design-system/bingel-dc': minor
```
anna-lach, line 2: It's `minor`, so we will start with ver. `0.1.0`?

packages/tokens/src/I/B/base-new.json
```diff
@@ -0,0 +1,236 @@
+{
+	"I": {
```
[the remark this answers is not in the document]
jpzwarte: That is the structure within Tokens Studio. You won't see that in the exported themes. I chose to keep the structure in `packages/tokens/src` the same as in Tokens Studio.

PR #2618, opened 2025-09-12
New `<sl-time-field>` component

### LINKED ISSUE #2178: [Time field] Development
### 1. Structure & elements

- [ ]  Render container with focus ring wrapping input + action-icon slot
- [ ]  Support placeholder, time-icon, and clear-icon slots/props
- [ ]  Show time icon when field is empty; reserve right-padding for clear icon

### 2. Clear-button logic

- [ ]  Clear icon appears on hover / focus only when a value is present
- [ ]  Click / Enter / Space on clear icon empties the field and returns focus to input
- [ ]  Mirrors Date Field implementation for consistency

### 3. Opening the dropdown

- [ ]  Mouse: click anywhere in field (or icon) → open picker
- [ ]  Keyboard: with input focused, Enter / Space opens picker; Esc closes and restores focus

### 4. Keyboard navigation inside picker

- [ ]  Arrow keys move between options (or columns)
- [ ]  Tab cycles columns (hours → minutes) in split mode
- [ ]  Esc and clicking outside closes picker and returns focus to input

### 5. Dropdown modes

- [ ]  Split mode: three independent columns (hours / minutes) for wide ranges
- [ ]  Expose prop to switch modes and set interval

### 6. Time formats & masking

- [ ]  Display in 24-hour (00:00) or 12-hour (12:00 am/pm) — decided by CFA consumer prop
- [ ]  Numeric input masking guides entry (e.g., __:__)
- [ ]  Arrow ↑ / ↓ inside input increment / decrement time according to format granularity

jpzwarte:
Fixes #2178

### DIFF
--- a/.changeset/afraid-tools-lay.md
+++ b/.changeset/afraid-tools-lay.md
@@ -1,0 +1,6 @@
+---
+'@sl-design-system/tooltip': minor
+---
+
+Adds support for passing a config object to the tooltip directive (same as `Tooltip.lazy`).
+The directive now accepts either a string (content) or a `TooltipConfig` with configurable properties like: `position`, `maxWidth` and `ariaRelation`.
--- a/.changeset/upset-ghosts-call.md
+++ b/.changeset/upset-ghosts-call.md
@@ -1,0 +1,5 @@
+---
+'@sl-design-system/eslint-plugin-slds': minor
+---
+
+Improved 'button-has-label' rule: it now also accepts the tooltip directive when `ariaRelation: 'label'` is set.
--- a/packages/components/tooltip/src/tooltip-directive.spec.ts
+++ b/packages/components/tooltip/src/tooltip-directive.spec.ts
@@ -1,6 +1,7 @@
 import { type ScopedElementsMap, ScopedElementsMixin } from '@open-wc/scoped-elements/lit-element.js';
 import { expect, fixture } from '@open-wc/testing';
 import { Button } from '@sl-design-system/button';
+import { type PopoverPosition } from '@sl-design-system/shared';
 import { LitElement, type TemplateResult, html } from 'lit';
 import { spy, stub } from 'sinon';
 import { tooltip } from './tooltip-directive.js';
@@ -8,12 +9,14 @@ import { Tooltip } from './tooltip.js';
 
 describe('tooltip()', () => {
   it('should create a lazy tooltip on the host element', async () => {
-    spy(Tooltip, 'lazy');
+    const lazySpy = spy(Tooltip, 'lazy');
 
     const el = await fixture(html`<div ${tooltip('content')}>Host</div>`);
 
     expect(Tooltip.lazy).to.have.been.calledOnce;
     expect(Tooltip.lazy).to.have.been.calledWith(el);
+
+    lazySpy.restore();
   });
 
   it('should log a warning if the custom element is not defined on the document', async () => {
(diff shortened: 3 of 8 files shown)


packages/components/time-field/src/time-field.ts
```diff
@@ -0,0 +1,610 @@
...
+  static viewportMargin = 8;
+
+  /**
+   * Flag indicating whether the popover was just closed. We need to know this so we can
+   * properly handle button clicks that close the popover. If the popover was just closed,
+   * we don't want to show it again when the button click event fires.
+   */
+  #popoverJustClosed = false;
+
+  /** The start time. */
```
Diaan, line 55:
```suggestion
  /** The start time; the time that has the initial focus when the picker is opened when there isn't a value. */
```

The term `startTime` might be confusing because you will likely use the time-picker to let the user set the start time of something.

packages/components/time-field/src/time-field.ts
```diff
@@ -0,0 +1,610 @@
...
+  @property({ type: Boolean }) readonly?: boolean;
+
+  /**
+   * Whether the time field is a required field.
+   * @default false
+   */
+  @property({ type: Boolean, reflect: true }) override required?: boolean;
+
+  /** The start time of the field. If not set, it will use the current time. */
+  @property() start?: string;
```
Diaan, line 122: same; is `start` really the best word? What about `initial`?

Diaan (approved):

Diaan:
When using keyboard navigation, after you have picked an hour and you use the right arrow key to navigate to the minutes, the hour reel jumps to that your selected hour is in the 2nd position. Can this be prevented? it's a bit distracting. 
https://github.com/user-attachments/assets/8eda6809-6c15-4cd1-88fe-53bf9ae5749f

packages/components/time-field/src/time-field.ts
```diff
@@ -0,0 +1,610 @@
...
+          tabindex="-1"
+        >
+          ${this.renderMinutes()}
+        </ul>
+      </dialog>
+    `;
+  }
+
+  renderHours(): TemplateResult[] {
+    let hours = Array.from({ length: 24 / this.hourStep }, (_, i) => i * this.hourStep);
```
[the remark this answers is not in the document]
jpzwarte: No, this requirements was removed.

packages/components/time-field/src/time-field.ts
```diff
@@ -0,0 +1,610 @@
...
+        ?readonly=${this.readonly}
+        ?required=${this.required}
+        input-size="10"
+        part="text-field"
+        placeholder=${ifDefined(this.placeholder)}
+        show-validity=${ifDefined(this.showValidity)}
+        value=${ifDefined(this.value)}
+      >
+        <slot name="input" slot="input"></slot>
+        <sl-field-button
```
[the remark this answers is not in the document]
jpzwarte: This was removed after discussion with the designers.

packages/components/time-field/src/time-field.ts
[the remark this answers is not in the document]
anna-lach: Can you please leave a comment in the issue then with information which elements were removed from the requirements and why?

packages/components/time-field/src/time-field.ts
jpzwarte: I think `start` is (only) slightly better than `initial`. Sticking with that for now.

packages/components/time-field/src/time-field.stories.ts
```diff
@@ -0,0 +1,107 @@
...
+  },
+  render: ({ disabled, hint, label, max, min, placeholder, readonly, reportValidity, required, start, value }) => {
+    const onClick = (event: Event & { target: HTMLElement }): void => {
+      event.target.closest('sl-form')?.reportValidity();
+    };
+
+    return html`
+      <sl-form>
+        <sl-form-field .hint=${hint} .label=${label}>
+          <sl-time-field
```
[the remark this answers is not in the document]
jpzwarte: Fixed

packages/components/time-field/src/time-field.stories.ts
```diff
@@ -0,0 +1,156 @@
...
+};
+
+export const Readonly: Story = {
+  args: {
+    readonly: true,
+    value: '13:30'
+  }
+};
+
+export const Required: Story = {
```
anna-lach, line 129: In this case, the field should be invalid right? It's not set as invalid (even when clicked reportValidity) and invalid state is not being shown
image

packages/components/time-field/src/time-field.stories.ts
```diff
@@ -0,0 +1,156 @@
...
+};
+
+export const Start: Story = {
+  args: {
+    hint: 'The start time is set to 08:00, so that will be the default time shown when opening the listbox',
+    start: '08:00'
+  }
+};
+
+export const Steps: Story = {
```
anna-lach, line 144: Is it ok, when we have steps, those set are only steps for the time picker, not for the input as well?

So sometimes the value is shown in the dropdown, sometimes not
image
image

packages/components/time-field/src/time-field.stories.ts
```diff
@@ -0,0 +1,156 @@
...
+  args: {
+    readonly: true,
+    value: '13:30'
+  }
+};
+
+export const Required: Story = {
+  args: {
+    hint: 'This field is required, if you leave it empty you will see an error message when clicking the button',
+    reportValidity: true,
```
anna-lach, line 132: When I type in a wrong value and click `report validity` in this case it shows that the field is invalid and this is fine
image
and when I type in again a value (but a proper value this time) and leave the field, it's still showing invalid state
image

only after clicking once again `report validity` the field is valid
image

So this is not how it works in our other form controls. When we call `reportValidity()` and we change the value from invalid to valid, the field should be automatically set as valid, without calling `reportValidity()` once again, right?

anna-lach (approved):

packages/components/time-field/src/time-field.stories.ts
jpzwarte: The `<input>` can be valid while the time-field isn't.

packages/components/time-field/src/time-field.stories.ts
jpzwarte: As discussed on Slack, we're keeping the behavior for now.

PR #2621, opened 2025-09-12
Clickedu theme fixes

jpzwarte:


anna-lach (approved):
Just a question, is it a proper backdrop color set?

Image

DanielleRameau:
@anna-lach I don't know, the CFA doesn't have time to review all the components right now. We will improve the values in another story. So we can leave it as is for now.

PR #2655, opened 2025-09-22
Fix issue with `selectAll` flag not resetting after deselecting all items

### LINKED ISSUE #2625: [Grid - Selection] Select all in grouped grid not fully functional when deselecting all groups

If you have a grid with two groups en you check the box for "select all", every item in the grid is checked as it's supposed to.
If i now uncheck both groups (wether it be on record level or group level), all checkboxes will be unchecked, even the Select All checkbox. The blue bar with the selected items will disappear. This is all good.

But function isSelectAllToggled() on the datasource is still true and the set with selection is still filled with all items in the grid, this is not correct

### 🤔 Expected Behavior?

isSelectAllToggled() should return false and selection on the datasource should be empty

### 😯 Current Behavior

isSelectAllToggled() returns true (checkbox is unchecked) and datasource.selection is still filled with all id's, this should be an empty set

### 💁 Possible Solution

_No response_

### 🖥️ Steps to Reproduce

If you have a grid with two groups en you check the box for "select all", every item in the grid is checked as it's supposed to.
If i now uncheck both groups (wether it be on record level or group level), all checkboxes will be unchecked, even the Select All checkbox. The blue bar with the selected items will disappear. This is all good.

But function isSelectAllToggled() on the datasource is still true and the set with selection is still filled with all items in the grid, this is not correct

Chrome

### If other, please specify.

_No response_

Mac OS

### 👤 Your name

Menno Homan

### 🧢 Your product/team

Team us heit

### 🎨 Your Theme(s)

Magister

jpzwarte:
Fixes #2625


Windigo (approved):

Diaan:
Not really related, but: 
image
The icon is missing/incorrect

Diaan (approved):

PR #2677, opened 2025-09-26
Migrate from `@web/test-runner` to `vitest`

### LINKED ISSUE #2676: [Unit tests] Migrate from `@web/test-runner` to `vitest`

Some problems we're having with wtr:
- using `sinon` in tests that then fail will cause a 120second timeout in `wtr --watch`; only workaround is running in debug mode with Chrome open
- `@open-wc/testing` causes typing issues and as a result linting errors that are unexplainable
- wtr is unmaintained; vitest is not
- better vscode integration with vitest
- better storybook integration with vitest

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

jpzwarte:
Fixes #2676


packages/components/dialog/src/dialog.spec.ts
```diff
@@ -107,6 +108,9 @@ describe('sl-dialog', () => {
       dialog = el.renderRoot.querySelector('dialog')!;
 
       el.showModal();
+
+      // Wait for the dialog to open
+      await new Promise(resolve => setTimeout(resolve));
```
anna-lach, line 113: Do we still need this one?

vitest.shims.d.ts
```diff
@@ -0,0 +1,3 @@
+/// <reference types="@types/chai-dom" />
```
anna-lach, line 1: Maybe this file should be added to the .gitignore?

jpzwarte: Why?

PR #2714, opened 2025-10-01
Add autofocus support to sl-form and update dialog story to include s…

Diaan:
…l-form


anna-lach (approved):
Maybe it's worth to add a story with autofocus to Angular storybook forms examples?

PR #2734, opened 2025-10-07
Feature/2583 split the css output files into old and new tokens

Diaan:


packages/components/badge/src/badge.scss
```diff
@@ -105,7 +105,30 @@
...
+  background: transparent;
+  color: inherit;
+  font-size: 1em;
+  line-height: unset;
+}
+
+:host([size='sm'][round])::before {
+  background: var(--_background, var(--sl-color-background-accent-grey-subtle));
+  block-size: var(--sl-size-100);
+  border-radius: var(--sl-size-borderRadius-full);
```
jpzwarte: I don't think it makes sense to use a token here; in other places i just use `50%`.

scripts/build-themes.js
```diff
@@ -225,104 +226,115 @@ const build = async (production = false, path) => {
...
+            outputReferences: !production
+          }
+        }
+      ];
+
+      if (production) {
+        files.push(
+          {
+            destination: old ? `${themeBase}/${theme}/css/base-deprecated.css` : `${themeBase}/${theme}/css/base.css`,
+            // filter: excludeSpaceTokens,
```
jpzwarte: Does it make sense to now uncomment this line?

website/src/categories/getting-started/developers.md
```diff
@@ -137,6 +137,12 @@ To initialize the theme you need to run theme's `setup` function in a global JS
 
 When you're using Angular this can be done in `main.ts` in the root folder of your application for example.
 
+### Deprecated tokens
+
+Sometimes we new insights or techniques cause us to no long use a certain token and to keep things clean and maintainable we will remove this after a deprecation period of at least a year. The latest version of a component is, at the time of release, always compatible with the (at that time) latest version of a theme.
```
jpzwarte:
```suggestion
Sometimes new insights or techniques cause us to no longer use a certain token. To keep things clean and maintainable we will remove this after a deprecation period of at least a year. The latest version of a component is, at the time of release, always compatible with the (at that time) latest version of a theme.
```

PR #2842, opened 2025-11-18
Feature/2725 callout develop component

anna-lach:


packages/components/callout/src/callout.scss
```diff
@@ -0,0 +1,74 @@
+:host {
+  align-items: center;
+  background: var(--sl-color-background-info-subtlest);
+  border-radius: var(--sl-size-borderRadius-default);
+  color: var(--sl-color-foreground-info-bold);
+  display: grid;
+  gap: 0 var(--sl-size-100);
+  grid-template-columns: auto 1fr;
+  overflow: clip;
```
jpzwarte: Is this necessary?

packages/components/callout/src/callout.scss
```diff
@@ -0,0 +1,74 @@
...
+  padding-block: var(--sl-size-150);
+  padding-inline: var(--sl-size-200);
+  position: relative;
```
jpzwarte: Is this necessary? I don't see any absolutely positioned elements below

packages/components/callout/src/callout.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    });
+
+    it('should have positive variant when set', async () => {
+      el.variant = 'positive';
+      await el.updateComplete;
+
+      expect(el).to.have.attribute('variant', 'positive');
+    });
+
+    it('should have caution variant when set', async () => {
```
jpzwarte: Isn't 1 test for `variant` enough? Cauting & negative don't add very much?

packages/components/callout/src/callout.ts
```diff
@@ -0,0 +1,93 @@
...
+        return 'info';
+    }
+  }
+
+  /** @internal If the title is missing, the content needs to be placed where the title should be. */
+  @property({ type: Boolean, attribute: 'no-title', reflect: true }) noTitle = true;
+
+  /**
+   * The density of the callout.
+   * @default plain
```
jpzwarte: `@default 'plain'`

.changeset/some-rabbits-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sl-design-system/callout': patch
+---
+
+New `callout` component.
```
jpzwarte: Can you add a little bit more information on *why* we have callout and how it is different from inline message? The text here ends up in `CHANGELOG.md` :)

packages/components/callout/src/callout.stories.ts
```diff
@@ -38,18 +37,25 @@ export default {
         disable: true
       }
     },
-    size: {
+    density: {
       control: 'inline-radio',
-      options: ['md', 'lg']
+      options: ['plain', 'comfortable']
```
Diaan: Roald uses "compact", "default" and "relaxed" in his example...
https://www.figma.com/board/5s9wbKRDR9Lw9bfa1eL0QY/Untitled?node-id=0-1

anna-lach: We keep 'plain' and 'comfortable' for the backward compatibility for now.

PR #2865, opened 2025-11-28
Chore/2857 angular make it possible to use slds with angular v21

Diaan:
Tested in the demo-app, using Portal Protocol, and i don't see any warnings, and it seems to work fine in the angular storybook as well.


packages/angular/package.json
```diff
@@ -12,6 +12,11 @@
     "directory": "packages/angular"
   },
   "type": "module",
+  "peerDependencies": {
```
anna-lach: I think there should be no such peerDependencies?

PeerDependencies should be updated in the file `packages/angular/src/package.json`?

Image

PR #2867, opened 2025-12-02
Feature/2508 toolbar development

anna-lach:


packages/components/tool-bar/src/tool-bar.stories.ts
```diff
@@ -199,18 +254,30 @@ export const Inverted: Story = {
...
-      'This example shows a tool bar without a border. Without a border, it also does not have any inline padding. Please make sure that the focus outlines of the buttons are still visible.',
-    noBorder: true
+    contained: true,
+    description: html`
+      This example shows a contained tool bar with inverted buttons. You have to set the
+      <code>inverted</code> attribute on the tool bar, otherwise the menu button will not be inverted.
+    `,
+    inverted: true,
+    items: () => html`
+      <sl-button fill="outline" variant="inverted">Action 1</sl-button>
```
Diaan: can we not do this smarter? So that all buttons inherit the inverted variant?

arecuenco-dsgn:
I saw an overlap when the toolbar started to overflow.

Screenshot 2025-12-04 at 11 42 48

arecuenco-dsgn:
Can you make the container fit the toolbar in the inverted variants to show that they don't have any padding around? Right now, the container has padding, and you can see the real size.

Screenshot 2025-12-04 at 11 53 38

Like you have for the non-inverted variant:
Screenshot 2025-12-04 at 11 57 01

anna-lach:
> Can you make the container fit the toolbar in the inverted variants to show that they don't have any padding around? Right now, the container has padding, and you can see the real size.
> 
> Screenshot 2025-12-04 at 11 53 38

yes, it's there (applied in the examples) just to show that it needs some space for the focus ring, but maybe I can change it a little bit, to add the padding to the container, not the tool-bar there

packages/components/tool-bar/src/tool-bar.ts
```diff
@@ -165,7 +149,8 @@ export class ToolBar extends ScopedElementsMixin(LitElement) {
   /** @internal True when the tool-bar is empty. */
   @property({ type: Boolean, reflect: true }) empty?: boolean;
 
-  /** Use this if you want the menu button to use the "inverted" variant.
+  /**
+   * Use this if you want the menu button to use the "inverted" variant.
```
Diaan: it says `menu-button` but i guess that should be `toolbar`?

packages/components/tool-bar/src/tool-bar.scss
```diff
@@ -1,21 +1,19 @@
...
-  flex: 1; // Allow the wrapper to grow, but only when aligned to the end, when aligned to start the overflow is broken
+  flex: 1; /* Allow the wrapper to grow, but only when aligned to the end, when aligned to start the overflow is broken */
   flex-wrap: wrap;
   justify-content: end;
 }
 
 :host([contained]) {
   background: var(--sl-elevation-surface-base-default);
-  border: var(--sl-size-borderWidth-default) solid var(--_border-color);
+  border: var(--sl-size-borderWidth-default) solid var(--sl-color-foreground-secondary-onBold);
```
Diaan: you are using a foreground-color for a border.

PR #2874, opened 2025-12-07
Docs/2513 Paginator Documentation

arecuenco-dsgn:


website/src/categories/components/paginator/accessibility.md
```diff
@@ -0,0 +1,62 @@
...
+
+<section>
+
+## Announcements
+
+The paginator provides screen reader announcements for important state changes:
+
+- **Page changes**: When navigating between pages using previous/next buttons or selecting a page directly, the new page number and total pages are announced.
+- **Page size changes**: When changing the number of items per page, the new page size is announced to inform users of the updated display.
+- **Status updates**: The current page range and total item count are available to screen readers to provide context about the displayed content.
```
Diaan: Maybe also mention when this is read out loud.

website/src/categories/components/paginator/usage.md
```diff
@@ -0,0 +1,320 @@
...
+
+</div>
+</section>
+
+<section>
+
+## When to use
+
+### Large Datasets   
+When there is too much data to display within a single page or component view, and you want to break it into manageable chunks so users can scan, compare, and keep their place more easily. Also when loading all available data at once would be slow, heavy, or resource-intensive, and you need to keep performance predictable.
```
Diaan:
```suggestion
When there is too much data to display within a single page or component view, and you want to break it into manageable chunks so users can scan, compare, and keep their place more easily. Also when loading all available data at once would be slow, heavy or resource-intensive, and you need to keep performance predictable.
```

PR #2894, opened 2025-12-15
Fix/2852 make breadcrumbs more flexible v2

Diaan:


packages/components/breadcrumbs/src/breadcrumb-item.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * ```html
+ *   <sl-breadcrumb-item>
+ *     Breadcrumb label
+ *   </sl-breadcrumb-item>
+ * ```
+ *
+ * @slot default - The content to display inside the breadcrumb item.
+ */
+export class BreadcrumbItem extends ScopedElementsMixin(LitElement) {
```
anna-lach: Can you please add changeset?

PR #2955, opened 2026-01-29
Fix/2785 [Time field] Min time does not work with minutes 

anna-lach:


a11ymiko (approved):
**Accessibility check information:**
Keyboard user and screen reader user can only select hours and minutes that are active. 
Screen readers announce names, roles and states of option properly. 
Hours and minutes selected via keyboard are properly added as input value to Time field.

Diaan (approved):

PR #2982, opened 2026-02-09
fix(callout): update examples to use anchor tags instead of button links

Diaan:


michal-sanoma (approved):

packages/components/callout/src/callout.scss
```diff
@@ -7,6 +7,12 @@
   gap: 0 var(--sl-size-100);
   grid-template-columns: auto 1fr;
   padding: var(--sl-size-150) var(--sl-size-200);
+
+  --sl-color-component-link-idle: var(--sl-color-foreground-info-bold);
+}
+
+::slotted(a[href]) {
+  color: var(--sl-color-component-link-idle);
```
Diaan: This is for when the CFA doesn't use global.css

When they do have global styling with the exact same selector other than from our global.css or when the link is not directly in the root of the `sl-callout` it will break. I can't think of a way to access a deeper link without the global.css...

packages/components/callout/src/callout.scss
```diff
@@ -26,16 +32,22 @@
 :host([variant='positive']) {
   background: var(--sl-color-background-positive-subtlest);
   color: var(--sl-color-foreground-positive-bold);
+
+  --sl-color-component-link-idle: var(--sl-color-foreground-positive-bold);
```
Diaan: This way we can use it in other components as well... Like breadcrumbs.

PR #3012, opened 2026-02-16
Refactor date-field

### LINKED ISSUE #1863: [date-field] Finish component

- [ ] Handle basic keyboard input of date
- [ ] Determine behavior of keyboard input (shortcuts etc.)
- [ ] Implement behavior for keyboard input
- [ ] Write unit tests

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

### LINKED ISSUE #2043: [date field] manual user input doesn't work

define a date field (inside an `sl-form`)
`<sl-date-field name="effectiveDate"></sl-date-field>`

If a date is typed in the field (using the keyboard). The value is not retained (`this.#form.value?.effectiveDate` is undefined). The `sl-change` event not triggered either.

However, if a date is selected using the popover the form value (`this.#form.value?.effectiveDate`) is set correctly 

### 🤔 Expected Behavior?

should fill the form field value if a value is entered using the keyboard.

### 😯 Current Behavior

-

### 💁 Possible Solution

-

### 🖥️ Steps to Reproduce

-

Chrome

### If other, please specify.

_No response_

Windows

### 👤 Your name

Frans de Boer

### 🧢 Your product/team

Magister

### 🎨 Your Theme(s)

Magister

jpzwarte:
Fixes #1863 
Fixes #2043 

Storybook link: https://github.com/sl-design-system/components/pull/3012#issuecomment-3907921450

### DIFF
--- a/packages/components/date-field/src/date-field.ts
+++ b/packages/components/date-field/src/date-field.ts
@@ -534,12 +534,12 @@ export class DateField extends LocaleMixin(FormControlMixin(ScopedElementsMixin(
     const relatedTarget = event.relatedTarget as Node | null;
 
     if (!relatedTarget || (!this.contains(relatedTarget) && !this.renderRoot.contains(relatedTarget))) {
+      this.placeholderShown = !this.value && !!this.placeholder;
+
       this.blurEvent.emit();
       this.updateState({ touched: true });
       this.updateValidity();
     }
-
-    this.placeholderShown = !this.value && !!this.placeholder;
   };
 
   #onKeydown(event: KeyboardEvent): void {


packages/components/date-field/src/date-field.scss
```diff
@@ -1,13 +1,89 @@
 :host {
```
anna-lach: It looks like there is sth wrong with the width, maybe fit-content is missing?

Image

packages/components/date-field/src/date-field.stories.ts
anna-lach: When I typed in a wrong value like `31/04/2025` it behaves weird, when I leave the field the value disappears, but when I click back in the date field the value is still there.

https://github.com/user-attachments/assets/59276780-32b9-4ec3-b61b-29b0a97a8a8e

anna-lach: With arrow keys (used in the spinbutton) I can still set invalid date

Image

So I guess spinputtons should take into account min and max?

And I gues aria-valuemin and aria-valuemax in that case should be different, when we have `min` and `max` set?

Image

@a11ymiko what's your opinion here?

anna-lach: Using arrow keys in the spinbutton I can still set e.g. 31 of February, so it looks like the month is not taken into account in the day spinbutton

https://github.com/user-attachments/assets/0a907644-e440-47e4-8c0b-deb3189f4295

packages/components/date-field/src/date-field.ts
```diff
@@ -1,37 +1,56 @@
...
+}
+
+type DatePartType = 'day' | 'month' | 'year';
 
 /**
  * A form component that allows the user to pick a date from a calendar.
+ * Uses individual spinbutton inputs per date part for improved accessibility.
  */
 @localized()
 export class DateField extends LocaleMixin(FormControlMixin(ScopedElementsMixin(LitElement))) {
```
anna-lach: What about the Angular directive for the Date field?

.changeset/whole-cloths-crash.md
```diff
@@ -0,0 +1,9 @@
+---
+'@sl-design-system/date-field': minor
+---
+
+Various improvements:
+
+- Keyboard entry is now supported. Users can type in a date, but they can also use the arrow keys to navigate between date parts and increment/decrement values. This works for any locale
```
anna-lach: Maybe it's worth to mention that it's an accessibility improvement.

.storybook/preview.ts
```diff
@@ -59,13 +67,22 @@ const preview: Preview = {
...
+          { value: 'en-GB', right: '🇬🇧', title: 'English (UK)' },
+          { value: 'es', right: '🇪🇸', title: 'Español' },
+          { value: 'fr', right: '🇫🇷', title: 'Français' },
+          { value: 'it', right: '🇮🇹', title: 'Italiano' },
+          { value: 'nl', right: '🇳🇱', title: 'Nederlands' },
+          { value: 'nl-BE', right: '🇧🇪', title: 'Nederlands (België)' },
+          { value: 'no', right: '🇳🇴', title: 'Norsk' },
+          { value: 'pl', right: '🇵🇱', title: 'Polski' },
+          { value: 'fi', right: '🇫🇮', title: 'Suomi' },
+          { value: 'sv', right: '🇸🇪', title: 'Svenska' }
```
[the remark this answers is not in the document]
anna-lach: Or maybe it's worth to explain it here as well on the Welcome page:

Image

That we currently have translation files for Dutch and English only?

packages/components/button-bar/src/button-bar.ts
```diff
@@ -22,17 +22,28 @@ export type ButtonBarAlign = 'start' | 'center' | 'end' | 'space-between';
  * ```
  *
  * @slot default - Buttons to be grouped in the bar.
+ * @cssState icon-only - Set when all buttons in the bar are icon-only.
```
anna-lach: Those `cssState` are not being rendered on the website, probably sth we need to look into in a separate issue?

Image

jpzwarte: Created https://github.com/sl-design-system/components/issues/3064

PR #3023, opened 2026-02-19
Chore/2996 menu align menu styling with figma

Diaan:


Diaan:
removed to save space on azure

PR #3055, opened 2026-03-02
3028 [Time field] Accessible name toggle button

Diaan:
Improve accessibility for time field with updated placeholder and aria-label

### DIFF
--- a/packages/components/time-field/src/time-field.stories.ts
+++ b/packages/components/time-field/src/time-field.stories.ts
@@ -106,7 +106,7 @@ export const Disabled: Story = {
 export const Placeholder: Story = {
   args: {
     placeholder: 'Add a time',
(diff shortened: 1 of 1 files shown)


a11ymiko (approved):

packages/components/time-field/src/time-field.stories.ts
```diff
@@ -31,7 +31,7 @@ export default {
   args: {
     disabled: false,
     label: 'Time',
-    placeholder: 'Select time',
+    placeholder: 'HH:MM',
```
anna-lach: For Finnish it should be `HH.MM` so it depends on locale.

Diaan: Yes, but is is a placeholder an example. "Select time" is also something else in Finnish :D

PR #3077, opened 2026-03-09
Fix/2866 time field functionality needs to be clarified to assistive technologies and users

Diaan:


packages/components/date-field/src/date-field.stories.ts
anna-lach: When there is `min` and `max` set why I can't type wrong value like `18:55` but I can type another wrong value e.g.  `04:55` or `19:55` that are out of range? 

https://github.com/user-attachments/assets/13b19e24-48c0-4055-9246-b0fcfdf4ebd4

a11ymiko:
Would be it much additional work to add `aria-label` to each hour and minut menu item in "Select time" dialog? 
I think that it would be beneficial to users to get each hour/minute announces as "NUMBER hour, menu item" instead of "NUMBER, menu item". Example of such approach is presented: https://mui.com/x/react-date-pickers/time-picker/ where they add `aria-label='NUMBER hours'` to each menu item.

It's not mandatory for us to add this `aria-label` as we already have names for listboxes "Select hour"/"Select minutes" so users can get this information, but is not given as straightforward as it could. In some cases information "Select minutes" may be lost at first when user select hours and navigate to the next listbox.
If it's not that much work I would recommend to add that additional names for hours and minutes options.

image

PR #3092, opened 2026-03-13
Use spinbutton role for number-field

### LINKED ISSUE #2922: [Number field] Should have role="spinbutton"

Number field is currently a text field, which only allows numbers (at least when validating) and users can change values using arrow up or arrow down and has somewhat discrete ranges (e.g., number cant be negative or must be between N and M). The actions are not clear for users from the role. The number field is similar to <input type="number">, which allows numbers only, and users can use up or down arrow keys to change the value. Number field should have a role spinbutton, which would indicate the actions directly from the role. 

Personally I would also only allow users to type numbers.

### 🤔 Expected Behavior?

Users are aware of the input methods.

### 😯 Current Behavior

Users can discover the input methods through trial and error, but are not native to a text field.

### 💁 Possible Solution

Use role="spibutton", which indicates that the field allows for numbers and has increment/decrement using up/down arrow keys.

### 🖥️ Steps to Reproduce

1. Focus the element with a screen reader -> No indication of actions are available through role or otherwise (apart from some variant where a range is specified using aria-describedby).

Chrome

### If other, please specify.

_No response_

Windows 11

### 👤 Your name

Valtteri Järvinen

### 🧢 Your product/team

External

### 🎨 Your Theme(s)

Sanoma Learning

jpzwarte:
Fixes #2922


a11ymiko:
In Firefox on macOS empty number field's value is announced as "NaN" which can be unclear for nontechnical users. Is there a way to announce empty number field as with value "0"? that's how it's announced in different browsers.


https://github.com/user-attachments/assets/2a562bfa-396b-4d2c-87ef-9a9a62f2be41

PR #3173, opened 2026-04-03
feat(editor): add rich text editor toolbar

jpzwarte:
Add `<sl-editor-toolbar>` component that renders a formatting toolbar
inside `<sl-editor>` using `<sl-tool-bar>` and `<sl-button>`.

Toolbar groups:
- Text marks: Bold, Italic, Underline, Strikethrough, Inline code
- Block formats: Blockquote, Heading 1–3
- Lists: Bullet list, Ordered list
- History: Undo, Redo (disabled when unavailable)

Buttons reflect the current editor selection state via `aria-pressed`
and `fill="outline"` when active. The toolbar is schema-aware — buttons
only render when the corresponding mark/node is in the editor schema.

Set `toolbar=false` on `<sl-editor>` to hide the toolbar.

https://claude.ai/code/session_01D6g9VibYCJSMMmMp1y5qgm


Diaan:
closed because we need environments

PR #3192, opened 2026-04-08
Fix/3174 textarea   rows attributeprop does not affect textarea sizerows shown

### LINKED ISSUE #3174: [TextArea] - rows attribute/prop does not affect textarea size/rows shown

The component has an attribute/prop called "rows" by default browser sets it to 2. However adding it doesn't make the component change. Instead I need to use:

`    --sl-text-area-rows: 1;`

Is there an option to

### 🤔 Expected Behavior?

The rows attribute/prop should directly control the number of visible rows in the textarea, behaving consistently with the native <textarea> element. Applying rows should update the component height without requiring the use of a CSS custom property.

### 😯 Current Behavior

Setting the rows attribute or property on the component has no effect on the rendered textarea height. The component ignores this value and keeps the default size unless the CSS variable --sl-text-area-rows is used.

### 💁 Possible Solution

_No response_

### 🖥️ Steps to Reproduce

`        <sl-text-area
          rows="1"
          size="md"
          resize="vertical"
        ></sl-text-area>`

Chrome

### If other, please specify.

_No response_

Windows 11

### 👤 Your name

Oscar Ledesma

### 🧢 Your product/team

Bingel DC/Pokemon

### 🎨 Your Theme(s)

Editorial Suite

Diaan:
This pull request updates the `TextArea` component to standardize its default row behavior and improve how the minimum height is handled when resizing. The changes clarify and enforce that the default number of visible rows is 3 (instead of the browser default of 2), improve the CSS for minimum height calculation, and enhance documentation and storybook examples.

**Default rows behavior and minimum height handling:**

* The default number of rows for the textarea is now explicitly set to 3 if not provided or if an invalid value is given, both in the property and when rendering the textarea. [[1]](diffhunk://#diff-b1dcaede9b2911a3dcb6b027509b86f38b40494831597795ab92e25f7a943da8L100-R99) [[2]](diffhunk://#diff-b1dcaede9b2911a3dcb6b027509b86f38b40494831597795ab92e25f7a943da8L242-R248)
* The minimum block size for the textarea is now controlled via a CSS variable (`--_sl-text-area-rows`), which is set according to the number of rows, ensuring the minimum height matches the intended number of rows when `resize="vertical"` or `resize="auto"` is used. [[1]](diffhunk://#diff-93fa36dfb182542316302632ac6c9bcbf26ce5bef7498d706ca4291732d7ce7cR64-R67) [[2]](diffhunk://#diff-93fa36dfb182542316302632ac6c9bcbf26ce5bef7498d706ca4291732d7ce7cL81-R84) [[3]](diffhunk://#diff-b1dcaede9b2911a3dcb6b027509b86f38b40494831597795ab92e25f7a943da8L242-R248)
* The CSS custom property `--sl-text-area-rows` was removed as it is no longer used.

**Resize logic improvements:**

* The resize logic in the component was updated to properly clear or reset the textarea height depending on the `resize` property, ensuring consistent behavior for all resize modes.

**Documentation and storybook updates:**

* Storybook stories and hint texts were updated to clarify the default row behavior, and a new story was added to demonstrate setting a custom number of rows.
* The outdated JSDoc comment for the removed CSS property was deleted.

### DIFF
--- a/packages/components/text-area/src/text-area.scss
+++ b/packages/components/text-area/src/text-area.scss
@@ -62,7 +62,7 @@
 }
 
 :host(:where([resize='vertical'], [resize='auto'])) slot[name='textarea']::slotted(textarea) {
-  min-block-size: calc(var(--sl-text-area-rows) * 1lh);
+  min-block-size: calc(var(--_sl-text-area-rows, 1) * 1lh);
 }
 
 @media (prefers-reduced-motion: no-preference) {
--- a/packages/components/text-area/src/text-area.scss
+++ b/packages/components/text-area/src/text-area.scss
@@ -3,7 +3,8 @@
   --_bg-mix-color: var(--sl-color-background-input-interactive);
   --_bg-opacity: var(--sl-opacity-interactive-plain-idle);
   --_border: var(--sl-color-border-input);
-  --sl-text-area-rows: 3;
+
+  // --sl-text-area-rows: 3;
 
   align-items: start;
   background: color-mix(in srgb, var(--_bg-color), var(--_bg-mix-color) calc(100% * var(--_bg-opacity)));
@@ -62,9 +63,9 @@
   }
 }
 
-:host(:where([resize='vertical'], [resize='auto'])) slot[name='textarea']::slotted(textarea) {
-  min-block-size: calc(var(--sl-text-area-rows) * 1lh);
-}
+// :host(:where([resize='vertical'], [resize='auto'])) slot[name='textarea']::slotted(textarea) {
+//   min-block-size: calc(var(--sl-text-area-rows) * 1lh);
+// }
 
 @media (prefers-reduced-motion: no-preference) {
   :host(:where(:hover, :focus-within)) {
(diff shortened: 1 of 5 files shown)


packages/components/text-area/src/text-area.scss
```diff
@@ -62,6 +61,10 @@
   }
 }
 
+:host(:where([resize='vertical'], [resize='auto'])) slot[name='textarea']::slotted(textarea) {
+  min-block-size: calc(var(--_sl-text-area-rows, 1) * 1lh);
```
jpzwarte: Should this not be at least `2`?

jpzwarte: Or 3?

jpzwarte (approved):

packages/components/text-area/src/text-area.scss
Diaan: I don't think it really matters, because --_sl-text-area-rows will always be set by the script. So it's just fallback.

a11ymiko (approved):

michal-sanoma (approved):

PR #3197, opened 2026-04-08
Create infotip component

### LINKED ISSUE #2111: [Infotip] Create webcomponent

- [ ] Create webcomponent using popover and button component.
- [ ] use info-circle icon as default, but give possibility to overwrite this.

###  Testing scenarios:
There are 2 scenarios that now have an infotip that is not accessible. Make sure to fix/test these scenarios and close the stories: 
- [ ] https://github.com/sl-design-system/components/issues/3180
- [ ] https://github.com/sl-design-system/components/issues/3185

### 👤 Your name

Diana

### 🧢 Your Product/Team

_No response_

jpzwarte:
## Summary

Adds the new `sl-infotip` component — an info icon button that triggers a popover showing contextual content. It is designed to be placed in the `infotip` slot of `<sl-label>` inside a `<sl-form-field>`.

Closes #2111

### `@sl-design-system/infotip` (new package)

- `sl-infotip` renders a ghost icon button (default: `circle-info`) that toggles an `sl-popover` on click or keyboard activation
- Default slot accepts plain text or rich HTML content; custom icon via `icon` slot
- Uses `slotAssignment: 'manual'` and a `MutationObserver` to keep a light DOM `<div>` content copy in sync with the original slotted content
- The content copy gets a unique `sl-infotip-content-<n>` id exposed via the `contentId` property — this is what other components use for `aria-describedby`

### `@sl-design-system/form`

- `<sl-label>` gains an `infotip` slot so `sl-infotip` can be placed alongside the label text
- `<sl-form-field>` detects an `sl-infotip` in the label slot and adds its `contentId` to `aria-describedby` on the associated form control's native input, ensuring screen readers announce the infotip content when the input receives focus

### Other

- Storybook stories: Basic, CustomIcon, RichContent, InContext
- Vitest unit tests covering defaults, toggle behavior, custom icon, and the `aria-describedby` accessibility wiring
- Documentation pages (overview, usage, accessibility, code, specs) added to the website
- Changeset added for both `infotip` and `form`

## Usage

```html
<sl-form-field>
  <sl-label>
    Username
    <sl-infotip slot="infotip">This field requires a unique identifier used for account login.</sl-infotip>
  </sl-label>
  <sl-text-field placeholder="Username"></sl-text-field>
</sl-form-field>
```


packages/components/form/src/form-field.ts
```diff
@@ -264,7 +267,25 @@ export class FormField extends ScopedElementsMixin(LitElement) {
...
     const assignedElements = event.target.assignedElements({ flatten: true }),
-      label = assignedElements.find((el): el is Label => el instanceof Label);
+      label = assignedElements.find((el): el is Label => el instanceof Label),
+      infotip = label?.querySelector<Element & { contentId?: string }>('sl-infotip');
+
+    if (infotip) {
+      this.#infotip = infotip;
+
+      if (this.control && infotip.contentId) {
+        const describedby = this.control.formControlElement.getAttribute('aria-describedby'),
```
michal-sanoma: **issue:** I think there is a stale `aria-describedby` issue in the `sl-form-field` integration

`#onLabelSlotchange()` only appends the infotip `contentId` to the control, but it never removes the previous ID again. It also only reacts to `slotchange` on the `label` slot, so removing or replacing `<sl-infotip>` inside the same `<sl-label>` does not trigger any cleanup

That means the input can keep an old `aria-describedby` value that points to a node that no longer exists, because `sl-infotip` removes its copied content on disconnect

jpzwarte: Fixed

michal-sanoma (approved):

RoaldBoerema:
For info we use our own custom svg icon `svg-info` 
Scherm­afbeelding 2026-04-20 om 09 36 19

PR #3211, opened 2026-04-13
Fix/3104 select some select components do not have visible label

### LINKED ISSUE #3010: [Form field] Combobox input does not have properly added label
## Current behavior
[Form Field Combobox](https://storybook.sanomalearning.design/?path=/story/form-form-field--combobox) has text `<label>` that is associated with `<sl-combobox>` using `for` and `id` attributes.
`<sl-combobox>` also has `aria-describedby` with hint text.
Both `<label>` and `aria-describedby` shouldn't be added to `sl-combobox` as users won't place their focus on `<sl-combobox>`.
That is why hint is not read now with use of Tab key to navigate page.

Image

## Proposed solution
Change value of `for` in `<label>` to `id` value of `<input>`. 
Remove `aria-describedby` from `<sl-combobox>` and add it to `<input>`.

Image

This way users using Tab to navigate page will get both label name and hint text on the same element.

### 👤 Your name

Kacper

### 🧢 Your Product/Team

SLDS

### LINKED ISSUE #3016: [Switch] Custom Icons switches do not have any labels
## Curent behavior
3 switches inside story [Switch Custom Icons](https://storybook.sanomalearning.design/?path=/story/form-switch--custom-icons) do not have visible or invisible label added. 
That results in screen reader announcing each switch as "off, switch" without information about its name.
It also results in issue 'Form label' being reported with a11y add-on in storybook.

Image

## Proposed solution
Adding `aria-label` with value eg. 'Switch name' or 'Switch label' to each `<input type='checkbox'>` should resolve this issue as screen reader would announce "Switch name, off, switch" and there would not be issues reported by a11y add-on in storybook.

### 👤 Your name

Kacper

### 🧢 Your Product/Team

SLDS

### LINKED ISSUE #3104: [Select] Some select components do not have visible label
## Current behavior 
There is no external or internal visible label for examples on [documentation website](https://sanomalearning.design/categories/components/select/usage/) and for Storybook variants:

1. [Options Styling](https://storybook.sanomalearning.design/?path=/story/form-select--options-styling)
2. [No Visible Label](https://storybook.sanomalearning.design/?path=/story/form-select--no-visible-label)
3.  [Display Inline Block](https://storybook.sanomalearning.design/?path=/story/form-select--display-inline-block)

Having visible label can be very helpful especially for users with cognitive disabilities. It's also WCAG 3.3.2 requirement. 

Image

Image

## Suggested solution

I think in documentation website we should present the most accessible way to use Select which is using visible label.

For Options Styling and Display Inline Block visible labels can be used as well but in that cases label can be also provided with not selectable option, that is being displayed when user doesn't choose anything. This option should explain what user can select inside select. Using first selectable option as only visible label for select is not sufficient. Example of that is used in [Custom Styling](https://storybook.sanomalearning.design/?path=/story/form-select--custom-styling).

For No Visible label Variant additional information about how to met WCAG with this component should be added. Whenever select with no labels is used in application it is important to make sure some visual cues, preferably using text, indicating select purpose are added in close proximity of that select. By doing this WCAG 3.3.2 would be met and component would be more accessible.

### 👤 Your name

Kacper 

### 🧢 Your Product/Team

SLDS

Diaan:
This pull request focuses on improving accessibility and usability for the `combobox`, `select`, and `switch` components, as well as enhancing Storybook configuration and documentation. The main changes include adding `aria-label` attributes to various component stories for better accessibility, updating Storybook sidebar and story sorting, and making internal improvements to the `combobox` component to better handle observed attributes.

**Accessibility Improvements:**

* Added descriptive `aria-label` attributes to `sl-combobox`, `sl-select`, and `sl-switch` instances in their respective story files to enhance accessibility for screen readers. This includes various states and sizes (e.g., "Medium Empty", "Large Selected", "Switch with custom icons small"). [[1]](diffhunk://#diff-a0a328aa7c91cffc1c73edf219f5ef68834395e7c6245bd3742b3a1c37abb686L69-R183) [[2]](diffhunk://#diff-a0a328aa7c91cffc1c73edf219f5ef68834395e7c6245bd3742b3a1c37abb686R197) [[3]](diffhunk://#diff-ba95f29ea90adcfd2558b02bd21136cda1cb0fea2b1fc53201828c5fef92b15bL378-R394) [[4]](diffhunk://#diff-ba95f29ea90adcfd2558b02bd21136cda1cb0fea2b1fc53201828c5fef92b15bL528-R650) [[5]](diffhunk://#diff-336ec258c4a579be8c490f28d80bd2406b7ae731f66cb859777b3787220d249eL90-R102)

**Component Enhancements:**

* Updated the `Combobox` component to use the `ObserveAttributesMixin`, allowing it to observe changes to the `aria-label` attribute and forward them to the internal input element for improved accessibility. [[1]](diffhunk://#diff-8f9a70134090f1e8edb4e27b08e8b9849ea6c405ed1572a298eeb85a3e3fac7bR10) [[2]](diffhunk://#diff-8f9a70134090f1e8edb4e27b08e8b9849ea6c405ed1572a298eeb85a3e3fac7bL66-R70) [[3]](diffhunk://#diff-8f9a70134090f1e8edb4e27b08e8b9849ea6c405ed1572a298eeb85a3e3fac7bR281-R282)

**Storybook Configuration and Documentation:**

* Improved Storybook sidebar organization by collapsing specific roots and updated the story sorting order to prioritize "Welcome" and "Getting started" sections. [[1]](diffhunk://#diff-1f1a9343f22a44ec21bc8c695dc736e00feac850609f633c8fa8d1c446ebfb1aR58) [[2]](diffhunk://#diff-d42715fd3297e575fb61faba39bba1b83739d3fd533a719ccfb0d81f64862b15L160-R160)
* Enhanced `select` component stories by wrapping custom styling and option styling examples in `sl-form` and `sl-form-field` for better documentation and structure. Also added a label to the embedded components story. [[1]](diffhunk://#diff-ba95f29ea90adcfd2558b02bd21136cda1cb0fea2b1fc53201828c5fef92b15bR114) [[2]](diffhunk://#diff-ba95f29ea90adcfd2558b02bd21136cda1cb0fea2b1fc53201828c5fef92b15bL177-R179) [[3]](diffhunk://#diff-ba95f29ea90adcfd2558b02bd21136cda1cb0fea2b1fc53201828c5fef92b15bR194-R195) [[4]](diffhunk://#diff-ba95f29ea90adcfd2558b02bd21136cda1cb0fea2b1fc53201828c5fef92b15bR258-R262)


packages/components/combobox/src/combobox.ts
```diff
@@ -543,7 +571,7 @@ export class Combobox<T = any, U = T> extends FormControlMixin(ScopedElementsMix
   }
 
   #onFocus(): void {
-    if (!this.#pointerDown) {
+    if (this.#pointerDown) {
```
michal-sanoma, line 630: **issue:** This looks like a regression: the condition in `#onFocus` was flipped from !pointerDown to pointerDown. That blocks opening the list when focused via keyboard (Tab), and the comment above `#pointerDown` suggests the opposite behavior

Was this intentional? 

https://github.com/user-attachments/assets/ce2fcea8-6fd9-4a75-bbd3-68922cdb0796

packages/components/select/src/select.stories.ts
```diff
@@ -374,15 +382,15 @@ export const DisplayInlineBlock: Story = {
       }
     </style>
     <section>
-      <sl-select value="2">
+      <sl-select value="2" aria-label="Select step 1">
```
michal-sanoma, line 406: **question:** shouldn't we add here labels also? 

```
<sl-form-field label="Step 1">
    <sl-select value="2">
        <sl-option value="1">short</sl-option>
         <sl-option value="2">very very long option text</sl-option>
     </sl-select>
</sl-form-field>
```

packages/checklist/src/sanoma-learning-theme.scss.ts
```diff
@@ -545,6 +550,7 @@ export default css`
       --sl-color-background-inverted-interactive-plain: var(
         --sl-color-palette-grey-200
       ); /** Used as transparent layer on top of the muted, subtlest and subtle background to indicate the hover and pressed state. */
+      --sl-color-background-highlight: var(--sl-color-palette-yellow-100);
```
michal-sanoma, line 553: **question:** are we smuggling here some extra changes? 😃

packages/components/combobox/src/combobox.ts
Diaan: * Flag indicating whether pointerdown event has happened. We need to know this so
   * we know if we need to show the popover when the input receives focus. If pointerdown
   * has happened, we know that the input is being focused as a result of a click, and we
   * will show the popover in the click event handler, not the focus event handler.
   
   so;
pointerdown = true -> click -> show popover
pointerdown = false -> focus -> don't show popover.

That is what is happening now (in this PR), right?

packages/components/select/src/select.stories.ts
Diaan: No, this is a specific scenario where you want them inline, so there is no room for a label.

packages/checklist/src/sanoma-learning-theme.scss.ts
Diaan: This is not in a theme package file, but in the checklist component. That should work without a theme being loaded as well. So it's a (bit dirty) copy of a theme file 😅

packages/components/combobox/src/combobox.ts
Diaan: CoPilot had the same comment EVERY review it did by the way 😅

michal-sanoma: gotcha, so we had a misalignment so far with this Tab focus which was opening a popover. It's really visible when testing prod vs testEnv hence the comment

michal-sanoma (approved):

a11ymiko (approved):

anna-lach (approved):

PR #3214, opened 2026-04-14
fix(tag tag-list): Improve tag-list accessibility

### LINKED ISSUE #2845: [Tag] Improve accessibility of the removable tag

The removable `tag` right now can be removed when pressing `Backspace`/`Delete` key, but it should be possible to remove it with `Space`/`Enter` key instead (so just like normal action on a button). It would be more clear for the assistive technology users, they will understand it's an action. 

We probably won't need the `aria-describedby` there anymore (?)

### 👤 Your name

Anna Sobczak

### 🧢 Your Product/Team

SLDS

### LINKED ISSUE #2868: [Tag] Tags with interactions should have button roles and native button interactions

_Overflow_ and _Removable_ variants have interactions, but they don't have proper roles. Every interactable element should have a programmatic role. They do have a description "Press the backspace or delete the items", but this does not include a role, though it does make it clearer for some users that the item is interactable. 

The only way to remove the items is through using backspace or delete, or tapping/clicking on the "X" button, which is aria-hidden. Thus some screen reader users without a keyboard (e.g., on mobile) cannot remove the items.

### 🤔 Expected Behavior?

Screen readers announce the tag as "button", indicating an interactable element, along with its visible label.

Users can delete the item by activating it e.g., double tapping).

### 😯 Current Behavior

Users hear "Label, press the delete or [...]" but no role. Some screen readers might ignore the description because of the lack of role.

Users on mobile cannot perform the delete action without using a keyboard.

### 💁 Possible Solution

1. Add role button to the tags.
2. Add native button interactions (mouse click, enter and space keys) to cover all styles of use.

Note: Having the backspace or delete keys as options is okay.

### 🖥️ Steps to Reproduce

1. Focus the item with a screen reader -> reads label, and (sometimes) the description
2. (Mobile) perform action (e.g., double tap) -> item does not get removed.

Chrome

### If other, please specify.

_No response_

Windows 11

### 👤 Your name

Valtteri Järvinen

### 🧢 Your product/team

External

### 🎨 Your Theme(s)

Sanoma Learning

### LINKED ISSUE #2869: [Tag list] Clarification for interaction for assistive technology users

Currently the tag list is a list. The element has an interaction, where when focused, users can move between tag items using arrow keys. This is a great idea, especially when there are many items in the tag list, so users have only one tab stop, but the list remains interactive. However, the interaction isn't that clear to assistive technology users. 

The list role is not interactive semantically (this is the reason, why for example NVDA users must set focus to the list items using NVDA key + Space, rather than having the list focusable and interactable when entering in focus mode). List can have interactable elements inside the list (e.g., buttons: #2868), but the list itself typically is just a container, with the added semantics of the items being in a list, and therefore providing additional information of their position in the list. In this case the list contains the the interactable items in one tab stop, so some users might think the items in the list are broken, as the rest cannot be focused using tab. 

Fixing issue #2868 does help with this somewhat, because the button role helps users understand that the items are interactable, but the interaction isn't obvious from the roles.

### 🤔 Expected Behavior?

Users have a clear understanding of the list actions.

### 😯 Current Behavior

Users must use trial and error to understand the interaction.

### 💁 Possible Solution

Some form of clarification for the list interaction should help. Adding a clarifying text with aria-describedby would suffice. E.g., "Use arrow keys to move between list items when focused". This would also be read to users using assistive technology without a keyboard, might need some thinking in how to present it to users.

This element acts basically like a [Layout Grid (W3C ARIA Authoring Practices)](https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/layout-grids/). It could be refactored as one to make it more obvious, that the interaction is present. **NOTE:** The grid role expects that all the items in the grid are interactive, so this doesn't work with the Mixed variant

### 🖥️ Steps to Reproduce

**NVDA**
1. Press arrow keys to move through the list -> reads list item labels
2. Enter focus mode (NVDA key + Space) -> Shows focus, but still uses NVDA interactions with arrow keys
3. Set focus mode again (NVDA key + Space) -> Actually focuses the buttons and can switch the items in the list using arrow keys

Chrome

### If other, please specify.

_No response_

Windows 11

### 👤 Your name

Valtteri Järvinen

### 🧢 Your product/team

External

### 🎨 Your Theme(s)

Sanoma Learning

jpzwarte:
## What changed

This PR improves the accessibility of `<sl-tag>` and `<sl-tag-list>`, especially for removable tags, disabled removable tags, keyboard navigation, and screen reader users.

Closes #2868 #2869

### `<sl-tag>`

- Reworked removable tags so the remove action is handled by a native `<button>` instead of the host tag.
- Added a localized accessible name to the remove button: `Remove tag '<label>'`.
- Removed `aria-hidden="true"` from the remove button.
- Added `type="button"` to prevent accidental form submissions.
- Kept the remove button rendered for disabled removable tags and uses `aria-disabled="true"` instead of native `disabled`, so the action remains discoverable to assistive technologies.
- Guarded disabled remove actions so click and keyboard activation do not remove the tag.
- Moved Backspace/Delete handling to the remove button and prevents default behavior/propagation for those shortcuts.
- Added a custom `focus()` override:
  - removable tags delegate focus to the remove button,
  - overflow/static tags delegate focus to the label when applicable,
  - host-focus cases such as stacked counter tags remain focusable.
- Reworked focus styling:
  - removable tags show the focus ring on the remove button,
  - non-removable/host-focused tags keep a visible host focus ring,
  - mouse focus does not force the custom focus-visible state.
- Wrapped the default slot in `<div part="label">`.
- Added `@csspart label` and `@csspart button` documentation.
- Reflects the `removable` property to support CSS selectors.
- Reworked overflow tooltip rendering to use reactive state and `aria-describedby` on the label part.
- Derives the tag label from all slotted text content, including nested elements.
- Updates label and tooltip overflow state when slotted text changes without resize.
- Ensures clicking the label area does not remove a removable tag.

### `<sl-tag-list>`

- Sets `role="list"` on the tag list and `role="listitem"` on slotted tags.
- Adds screen reader navigation instructions to removable tag remove buttons.
- Keeps disabled removable tags in roving tabindex, while their remove buttons remain `aria-disabled`.
- Prevents disabled stacked counter tags from becoming tab stops.
- Excludes disabled stacked counter tags from roving tabindex candidates.
- Keeps tag properties synchronized on update, including localized navigation instructions after runtime locale changes.
- Keeps keyboard navigation scoped to visible removable tags in stacked lists.
- Added `@customElement sl-tag-list` JSDoc.

### `<sl-combobox>` integration with removable tags

- Updated the multi-select combobox integration with `<sl-tag-list>` so keyboard focus uses the real remove buttons from the tag list instead of competing with the combobox's older fake tag focus state.
- Prevented combobox input key handling from reacting to keyboard events that originate from the tag list remove buttons.
- Cleared the combobox fake `focusedTag` state when focus enters the tag list, avoiding two visible focus indicators at once.
- Improved focus behavior after removing selected tags from a multi-select combobox:
  - when a selected tag is removed, focus moves to the next visible removable tag,
  - if there is no next tag, focus moves to the previous visible removable tag,
  - if the removed tag was the only selected tag, focus returns to the combobox input.
- Added regression coverage for combobox tag focus behavior and removing multiple selected tags with the keyboard.

### Locales

- Removed the unused `sl.tag.removalInstructions` key.
- Added `sl.tag.remove`.
- Added `sl.tagList.navigationInstructions`.
- Updated locale files and XLF files for `es-ES`, `it`, `nl`, and `pl`.

### Stories and docs

- Added `OverflowRemovable` story.
- Added `RemovableDisabled` stories.
- Removed mixed tag list story.
- Updated tag examples to separate static and removable tag lists.
- Updated docs to clarify:
  - the remove button is the only interactive part of a removable tag,
  - static and removable tags should not be mixed in one list,
  - disabled tags should only be used for removable tags.

### Website accessibility follow-ups

- Updated website vertical navigation tabs from pseudo-tab semantics to navigation link semantics using `aria-current="page"`.
- Adjusted website link/tab colors for contrast.
- Narrowed the website axe exclusion for the known `sl-tab` page navigation issue so page content remains covered.

### Tests

- Updated and expanded `<sl-tag>` tests for:
  - remove button focus,
  - accessible labels,
  - slotted element labels,
  - disabled remove behavior,
  - label click behavior,
  - Backspace/Delete isolation,
  - tooltip updates,
  - host focus-visible fallback.

- Updated and expanded `<sl-tag-list>` tests for:
  - disabled removable tags in roving tabindex,
  - navigation descriptions,
  - localized description resync,
  - stacked disabled counter behavior,
  - all-regular-tags-hidden stacked edge case.


michal-sanoma (approved):
**suggestion:** Small a11y/UX concern: because `#onFocus()` always adds focus-visible, the host behaves like `:focus` instead of native `:focus-visible`. Would it make sense to track input source and apply the ring only for keyboard interaction?

a11ymiko (commented):
For Stacked variants keyboard user cannot set focus on tag with tooltip. `sl-tag` has `tabindex=0` but keyboard focus cannot be place on this tag.

a11ymiko (commented):
For Stacked Removable variant user cannot place focus on any tag. This happens on Safari, Chrome and Firefox on macOS and Edge and Chrome on Win11. On Firefox on Win11 user can place focus on some removable tag, but it's not the first one in the list.

https://github.com/user-attachments/assets/433ad638-b1dd-498e-abe3-54616d78bbbb

a11ymiko (commented):
Minor issue but on Chrome on macOS tag with Overflow has it's content announced twice (once from tag text and once from tooltip text). Not adding 'aria-describedby' to `div` in Overflow `sl-tag` will fix that issue in Chrome without making an issue for other OSes and browsers.

https://github.com/user-attachments/assets/25990edf-0d3e-461a-acdf-0b579ca6cf58

a11ymiko (commented):
For Stacked tag lists the first tag (with number of stacked items) doesn't have `role=listitem`. It should have `role=listitem` because it's inside `sl-tag-list role="list"`.

packages/locales/src/nl.xlf
```diff
@@ -426,6 +422,10 @@
   <source>No later than <x id="0" equiv-text="${format(this.max, this.locale, this.#helperTextFormatOptions)}"/></source>
   <target>Uiterlijk <x id="0" equiv-text="${format(this.max, this.locale, this.#helperTextFormatOptions)}"/></target>
 </trans-unit>
+<trans-unit id="sl.tag.remove">
```
Diaan, line 457: should we not add it to spanish and polish as well?

packages/locales/src/nl.xlf
```diff
@@ -10,10 +10,6 @@
   <source>Loading</source>
   <target>Laden</target>
 </trans-unit>
-<trans-unit id="sl.tag.removalInstructions">
-  <source>Press the delete or backspace key to remove this item</source>
-  <target>Druk op de delete- of backspacetoets om dit item te verwijderen</target>
-</trans-unit>
```
Diaan, line 16: should we keep this in as deprecated? If people update the locales but don't update the tag list component it will break the translation.

packages/components/tag/src/tag.ts
```diff
@@ -106,62 +114,59 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
     `;
   }
 
+  #onBlur(): void {
+    this.#internals.states.delete('focus-visible');
+  }
+
+  #onFocus(): void {
+    this.#internals.states.add('focus-visible');
+  }
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag.ts
```diff
@@ -106,62 +114,59 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
-  }
-
   override render(): TemplateResult {
+    const hasTabindex = !this.disabled && !this.removable && this.tooltip;
+
     return html`
-      <slot @slotchange=${this.#onSlotChange} part="label"></slot>
-      ${this.removable && !this.disabled
+      ${this.tooltip ? html`<sl-tooltip id="tooltip">${this.label}</sl-tooltip>` : nothing}
+      <div
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag.scss
```diff
@@ -43,25 +43,13 @@
...
-    background: var(--sl-color-background-disabled);
-  }
-}
-
-:host(:focus-visible) {
+:host(:state(focus-visible)) {
   outline-color: var(--sl-color-border-focused);
   position: relative;
   z-index: 1; // Make sure the focus ring is above other elements
 }
```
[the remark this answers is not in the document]
michal-sanoma: this is intentionally done like this

packages/components/tag/src/tag.ts
```diff
@@ -180,39 +187,17 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
   }
 
   #onSlotChange(event: Event & { target: HTMLSlotElement }): void {
     this.label = event.target
       .assignedNodes({ flatten: true })
-      .filter(node => node.nodeType === Node.TEXT_NODE)
-      .map(node => node.textContent?.trim())
-      .join('');
+      .map(node => node.textContent ?? '')
+      .join('')
```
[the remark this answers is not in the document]
michal-sanoma: fixed

website/src/ts/components/vertical-tabs/vertical-tabs.ts
```diff
@@ -187,11 +183,11 @@ export class VerticalTabs extends LitElement {
...
 
   #setActiveTab(verticalTab: HTMLElement): void {
-    const currentVerticalTabLink = this.renderRoot.querySelector('[aria-selected="true"]'),
+    const currentVerticalTabLink = this.renderRoot.querySelector('[aria-current="true"]'),
       verticalTabs = this.renderRoot.querySelectorAll('.ds-tab--vertical');
 
-    currentVerticalTabLink?.setAttribute('aria-selected', 'false');
-    verticalTab.setAttribute('aria-selected', 'true');
+    currentVerticalTabLink?.removeAttribute('aria-current');
+    verticalTab.setAttribute('aria-current', 'true');
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag.ts
```diff
@@ -101,67 +112,68 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
         ? html`
             <button
+              @blur=${this.#onBlur}
               @click=${this.#onRemove}
-              ?disabled=${this.disabled}
-              aria-hidden="true"
-              part="button"
-              tabindex="-1">
+              @focus=${this.#onFocus}
+              @keydown=${this.#onKeydown}
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag.ts
```diff
@@ -101,67 +112,69 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
+      this.#internals.states.add('focus-visible');
+    }
+  }
+
   #onKeydown(event: KeyboardEvent): void {
-    if (this.removable && (event.key === 'Backspace' || event.key === 'Delete')) {
+    if (event.key === 'Backspace' || event.key === 'Delete') {
       this.#onRemove(event);
     }
   }
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/locales/src/nl.xlf
michal-sanoma: fixed

packages/locales/src/nl.xlf
michal-sanoma: fixed

packages/components/tag/src/tag-list.ts
```diff
@@ -344,11 +350,17 @@ export class TagList extends ScopedElementsMixin(LitElement) {
...
+
     this.tags = Array.from(event.target.assignedElements({ flatten: true })).filter(
       (el): el is Tag => el instanceof Tag
     );
 
     this.tags.forEach(tag => {
+      tag.navigationDescription = msg('Use arrow keys to move between removable tags.', {
+        id: 'sl.tagList.navigationInstructions'
+      });
+      tag.role = 'listitem';
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag.scss
```diff
@@ -43,25 +43,13 @@
...
-    background: var(--sl-color-background-disabled);
-  }
-}
-
-:host(:focus-visible) {
+:host(:not([removable]):state(focus-visible)) {
   outline-color: var(--sl-color-border-focused);
   position: relative;
   z-index: 1; // Make sure the focus ring is above other elements
 }
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag-list.ts
```diff
@@ -135,14 +137,18 @@ export class TagList extends ScopedElementsMixin(LitElement) {
...
-    focusInIndex: (elements: Tag[]) => elements.findIndex(el => !el.disabled),
+    focusInIndex: (elements: Tag[]) => {
+      const index = elements.findIndex(el => el !== this.stackTag || !el.disabled);
+
+      return index === -1 ? 0 : index;
+    }
     elements: () => [
       ...(this.stacked && this.stackTag && this.stackTag.style.display !== 'none'
         ? [this.stackTag]
         : []),
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag.ts
```diff
@@ -54,14 +53,14 @@ export class Tag extends ScopedElementsMixin(LitElement) {
   /** @internal */
   static override styles: CSSResultGroup = styles;
 
-  // eslint-disable-next-line no-unused-private-class-members
-  #events = new EventsController(this, { keydown: this.#onKeydown });
+  /** @internal */
+  #internals = this.attachInternals();
```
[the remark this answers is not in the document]
michal-sanoma: description fixed

packages/components/tag/src/tag-list.ts
```diff
@@ -372,6 +382,20 @@ export class TagList extends ScopedElementsMixin(LitElement) {
...
+    });
+
+    this.tags.forEach(tag => {
+      tag.navigationDescription = navigationDescription;
+      tag.role = 'listitem';
+      tag.size = this.size;
+      tag.variant = this.variant;
+      tag.setAttribute('role', 'listitem');
+    });
+  }
```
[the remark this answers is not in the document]
michal-sanoma: imo to complex proposition to add for a such simple component

a11ymiko (commented):
Tag and Tags list works fine now as separate components. But tags list have some new issues when it's used inside Combobox: 
When navigating with keyboard user can place focus on "x" used for removal of tag. That button have visible focus indicator. But it's also possible to have visible focus indicator on whole tag and it's possible to have this indicator not on the same item that is currently receiving focus (presented on screen capture below).
This happens due to loop between last and first tag receiving focus.

https://github.com/user-attachments/assets/0f212518-6867-4582-97aa-effd9432dbc0

Getting rid of focus indicator on the whole tag should fix this issue.

a11ymiko (commented):
In Combobox when keyboard-only user removes one tag their focus is automatically placed on Combobox input. To remove next tag user has to take focus back with `Shift + Tab`. This may be frustrating when user want to remove more than one tag.

Proposed behavior is for focus to move to next/previous tag after removal of one tag, as it works for the list of removable tag in storybook. Only when there was only one tag and it was removed user's focus should be placed on Combobox input automatically.

https://github.com/user-attachments/assets/751c3568-8d53-4d77-9365-83de160c8a11

website/src/ts/components/vertical-tabs/vertical-tabs.ts
```diff
@@ -140,22 +140,18 @@ export class VerticalTabs extends LitElement {
...
                 html` <a
                   href=${`#${variant?.id}`}
                   class="ds-tab--vertical"
-                  role="tab"
-                  tabindex="0"
                   id=${`ds-vertical-tab-${this.nextUniqueId++}`}
-                  aria-selected="false"
-                  aria-controls=${(variant as HTMLElement)?.id}
                   @click=${this.#onClick}
                   >${(variant as HTMLElement)?.textContent || variant?.getAttribute('link-in-navigation-text')}</a
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag-list.ts
```diff
@@ -372,6 +382,20 @@ export class TagList extends ScopedElementsMixin(LitElement) {
...
+    const navigationDescription = msg('Use arrow keys to move between removable tags.', {
+      id: 'sl.tagList.navigationInstructions'
+    });
+
+    this.tags.forEach(tag => {
+      tag.navigationDescription = navigationDescription;
+      tag.role = 'listitem';
+      tag.size = this.size;
+      tag.variant = this.variant;
+      tag.setAttribute('role', 'listitem');
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag.ts
```diff
@@ -101,67 +106,86 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
+                  this.tooltip ? 'tooltip' : undefined,
+                  this.navigationDescription ? 'navigation-description' : undefined
+                ]
+                  .filter(Boolean)
+                  .join(' ') || undefined
+              )}
               part="button"
-              tabindex="-1">
+              type="button">
               <sl-icon name="xmark"></sl-icon>
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag.ts
```diff
@@ -101,67 +106,88 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
-              aria-hidden="true"
+              @focus=${this.#onFocus}
+              @keydown=${this.#onKeydown}
+              aria-describedby=${ifDefined(buttonDescription || undefined)}
+              aria-disabled=${ifDefined(this.disabled ? 'true' : undefined)}
+              aria-label=${msg(str`Remove tag '${this.label}'`, { id: 'sl.tag.remove' })}
               part="button"
-              tabindex="-1">
+              type="button">
               <sl-icon name="xmark"></sl-icon>
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag.ts
```diff
@@ -97,71 +108,106 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
+  protected override updated(): void {
+    this.#syncButtonTabIndex();
   }
 
   override render(): TemplateResult {
+    const hasTabindex =
+        !this.disabled && !this.removable && (this.tooltip || this.hasAttribute('tabindex')),
+      buttonDescription = [
+        this.tooltip ? 'tooltip' : undefined,
+        this.navigationDescription ? 'navigation-description' : undefined
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag.ts
```diff
@@ -179,40 +231,46 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
-      this.removeAttribute('tabindex');
+    if (this.navigationDescription) {
+      if (button.matches(':focus') && this.tabIndex === -1) {
+        return;
+      }
+
+      button.tabIndex = this.tabIndex;
+    } else {
+      button.removeAttribute('tabindex');
     }
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag.ts
```diff
@@ -179,40 +231,42 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
-    // so the user can tab to it.
-    if (!this.disabled && (this.removable || this.#tooltip)) {
-      this.setAttribute('tabindex', '0');
-    } else if (!this.hasAttribute('aria-labelledby')) {
-      this.removeAttribute('tabindex');
+    if (this.navigationDescription) {
+      button.tabIndex = this.tabIndex;
+    } else {
+      button.removeAttribute('tabindex');
     }
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag-list.ts
```diff
@@ -220,9 +234,7 @@ export class TagList extends ScopedElementsMixin(LitElement) {
   override updated(changes: PropertyValues<this>): void {
     super.updated(changes);
 
-    if (changes.has('size')) {
-      this.tags?.forEach(tag => (tag.size = this.size));
-    }
+    this.#syncTags();
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag-list.ts
```diff
@@ -372,6 +389,19 @@ export class TagList extends ScopedElementsMixin(LitElement) {
     });
   }
 
+  #syncTags(): void {
+    const navigationDescription = msg('Use arrow keys to move between removable tags.', {
```
a11ymiko, line 394: I'm a huge fan of that fix for tag list! 
Last touch that need to be done before review is to add `aria-hidden='true'` to `<span>` with navigationDescription. Then it still will be announced when user place their focus on close button of each tag but when user will be using different navigation modes there won't be that much repeating, not necessary elements with the same text.

michal-sanoma: good catch, added

packages/components/tag/src/tag.ts
```diff
@@ -97,71 +108,113 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
               part="button"
-              tabindex="-1">
+              type="button">
               <sl-icon name="xmark"></sl-icon>
             </button>
+            ${this.navigationDescription
+              ? html`
+                  <span id="navigation-description" class="sr-only" aria-hidden="true"
+                    >${this.navigationDescription}</span
+                  >
```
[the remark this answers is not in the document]
michal-sanoma: discussed with Kacper, this can stay

packages/components/tag/src/tag-list.spec.ts
```diff
@@ -85,6 +101,90 @@ describe('sl-tag-list', () => {
...
+
+    it('should describe arrow key navigation on remove buttons', () => {
+      const button = el.querySelector('sl-tag')?.renderRoot.querySelector('button'),
+        description = el
+          .querySelector('sl-tag')
+          ?.renderRoot.querySelector('#navigation-description');
+
+      expect(button).to.have.attribute('aria-describedby', 'navigation-description');
+      expect(description).to.have.attribute('aria-hidden', 'true');
+      expect(description).to.have.trimmed.text('Use arrow keys to move between removable tags.');
```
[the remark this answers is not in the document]
michal-sanoma: discussed with Kacper, this can stay

a11ymiko (approved):

packages/components/tag/src/tag-list.ts
```diff
@@ -473,7 +503,7 @@ export class TagList extends ScopedElementsMixin(LitElement) {
       }
     });
 
-    this.#rovingTabindexController.clearElementCache();
+    this.#clearRovingTabindexCache();
```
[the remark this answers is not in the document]
michal-sanoma: added

packages/components/tag/src/tag-list.ts
```diff
@@ -372,6 +389,19 @@ export class TagList extends ScopedElementsMixin(LitElement) {
...
+    const navigationDescription = msg('Use arrow keys to move between removable tags.', {
+      id: 'sl.tagList.navigationInstructions'
+    });
+
+    this.tags.forEach(tag => {
+      tag.navigationDescription = navigationDescription;
+      tag.size = this.size;
+      tag.variant = this.variant;
+      tag.setAttribute('role', 'listitem');
+    });
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/button/CHANGELOG.md
```diff
@@ -20,12 +20,10 @@
 ### Patch Changes
 
 - [#3299](https://github.com/sl-design-system/components/pull/3299) [`78e7333`](https://github.com/sl-design-system/components/commit/78e733338fd67ef59797b3e02b22907fe0f5c638) - Minor style fixes:
-
```
anna-lach: It looks like those formatting changes are not necessary, can you please revert those formatting changes in all the changelog files?

.storybook/stories/links.stories.ts
```diff
@@ -14,7 +14,9 @@ export default {
     In the theme is a global.css file containing styling for links. This works as a global style for
     links in your application and links in the light dom of SLDS components.
     <br />
-    <a href="https://sanomalearning.design" target="_blank" rel="noopener noreferrer">Sanoma Learning Design system</a>
```
anna-lach: This formatting change is probably not necessary?

packages/components/tag/src/tag.scss
```diff
@@ -84,28 +79,54 @@ button {
...
   }
+
+  &[aria-disabled='true'] {
+    background: var(--sl-color-background-disabled);
+    color: var(--sl-color-foreground-disabled);
+    cursor: default;
+  }
+
+  &:focus-visible {
+    border-radius: 0 var(--sl-size-borderRadius-focusRing-inside)
```
anna-lach, line 118: Focus ring looks different than in Figma, 
This one is in Storybook:


Image

and in Figma:

Image

anna-lach: Also the icon is smaller in the code than in Figma.

packages/components/tag/src/tag.scss
```diff
@@ -37,36 +37,31 @@
...
-  button {
-    background: var(--sl-color-background-disabled);
-  }
+:host([variant='info']) button:not([aria-disabled='true']) {
+  color: var(--sl-color-foreground-info-bold);
 }
 
-:host(:focus-visible) {
+:host(:not([removable]):where(:focus-visible, :state(focus-visible))) {
   outline-color: var(--sl-color-border-focused);
```
anna-lach, line 56: Does it look ok? Isn't this focus ring inside the combobox a little bit weird and as designed? 

Image

packages/components/tag/src/tag-list.ts
```diff
@@ -37,6 +37,8 @@ declare global {
  * </sl-tag-list>
  * ```
  *
+ * @customElement sl-tag-list
```
anna-lach: Do we really need it here?

packages/components/tag/src/tag.scss
```diff
@@ -84,28 +79,54 @@ button {
...
+    border-radius: 0 var(--sl-size-borderRadius-focusRing-inside)
+      var(--sl-size-borderRadius-focusRing-inside) 0;
+    outline: var(--sl-size-borderWidth-focusRing) solid var(--sl-color-border-focused);
+    outline-offset: calc(var(--sl-size-borderWidth-focusRing) * -1);
+    position: relative;
+    z-index: 1; // Make sure the focus ring is above other elements
+  }
+}
+
+.sr-only {
```
anna-lach: In grid for example we are using class name 'visually-hidden' for hidden elements, only visible for screen readers, maybe it's worth to use that name here as well and keep it consistent?

packages/components/tag/src/tag.spec.ts
```diff
@@ -47,66 +47,155 @@ describe('sl-tag', () => {
...
+    it('should have the focus-visible state when focused', async () => {
+      el.focus();
+      await el.updateComplete;
+
+      expect(el).to.match(':state(focus-visible)');
     });
 
-    it('should have a tabindex of -1 when disabled', async () => {
-      el.disabled = true;
+    it('should not have the focus-visible state when focus is not focus-visible', async () => {
```
anna-lach: Sounds a bit weird? :)

packages/components/tag/src/tag.stories.ts
```diff
@@ -72,8 +65,29 @@ export const Overflow: Story = {
...
 };
+
+export const InfoRemovable: Story = {
+  args: {
+    removable: true,
+    variant: 'info'
+  }
+};
+
+export const RemovableDisabled: Story = {
```
anna-lach, line 88: Does it make sense to inform a user that it can be removed when it's disabled? Or maybe this description makes sense with combination of 'dimmed' here?

Image

packages/components/tag/src/tag.ts
```diff
@@ -39,7 +33,12 @@ export type TagVariant = 'neutral' | 'info';
  * <sl-tag>Tag label</sl-tag>
  * ```
  *
+ * @customElement sl-tag
```
anna-lach: Do we need it here?

packages/components/tag/src/tag.ts
```diff
@@ -97,71 +108,113 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
+        .filter(Boolean)
+        .join(' ');
+
     return html`
-      <slot @slotchange=${this.#onSlotChange} part="label"></slot>
-      ${this.removable && !this.disabled
+      ${this.tooltip ? html`<sl-tooltip id="tooltip">${this.label}</sl-tooltip>` : nothing}
+      <div
+        @blur=${this.#onBlur}
+        @focus=${this.#onFocus}
```
anna-lach, line 171: Maybe better to use focusin and focusout?

packages/components/tag/src/tag.ts
```diff
@@ -97,71 +108,113 @@ export class Tag extends ScopedElementsMixin(LitElement) {
...
+        aria-describedby=${ifDefined(this.tooltip ? 'tooltip' : undefined)}
+        part="label"
+        tabindex=${ifDefined(labelTabIndex)}>
+        <slot @slotchange=${this.#onSlotChange}></slot>
+      </div>
+      ${this.removable
         ? html`
             <button
+              @blur=${this.#onBlur}
```
anna-lach, line 180: Maybe better to use focusin and focusout?

website/src/categories/components/tag/usage.md
```diff
@@ -9,14 +9,30 @@ eleventyNavigation:
 
 <div class="ds-example">
 <sl-tag>Mathematics</sl-tag>
+<sl-tag>History</sl-tag>
+<sl-tag>Science</sl-tag>
+</div>
+
+<div class="ds-code">
```
anna-lach, line 22: Is is a good idea to show two code examples in the Usage tab? We are not making it for other components

Image

website/src/categories/components/tag/usage.md
```diff
@@ -33,7 +49,10 @@ Use tags to categorize and label course content, enhancing organization, discove
 
 
 ### User-Generated Tags
-Enable users to create and manage their own custom labels, allowing for a personalized organization system. In a student dashboard, for example, users might tag their notes or assignments with custom labels like "Exam Prep," "Homework," "Group Project," or "To Review," facilitating better organization.
+Enable users to create and manage their own custom labels, allowing for a personalized organization system. In a student dashboard, for example, users might tag their notes or assignments with custom labels like "Exam Prep," "Homework," "Group Project," or "To Review," facilitating better organization. Use removable tags when users can edit or remove those labels.
```
anna-lach: Are those changes in the 'usage' tab consulted with @arecuenco-dsgn ? This part is typically done by Alvaro and it's more for Figma users.

website/src/categories/components/tag/usage.md
```diff
@@ -56,9 +75,9 @@ For tracking the status of tasks or items, such as "In Progress," "Completed," o
 
 |Item|Name| Description | Optional|
 |-|-|-|-|
-|1|Container |The container contains the label and close button |no|
+|1|Container |The container contains the label and optional close button |no|
```
anna-lach, line 74: Are those changes in the 'usage' tab consulted with @arecuenco-dsgn ? This part is typically done by Alvaro and it's more for Figma users.

website/src/categories/components/tag/accessibility.md
```diff
@@ -46,6 +47,8 @@ When the tag is focused and is removable, it can be removed by pressing the `Del
 
 {.ds-table .ds-table-align-top}
 
+Do not mix static and removable tags in the same tag list. Use a static tag list when no tags can be removed, and a removable tag list when users can remove tags.
```
anna-lach: Maybe it's better to add that one to the new section 'Accessibility considerations' and put it somewhere on the top of the page or below 'Keyboard interactions' section.

website/src/categories/components/tag/accessibility.md
```diff
@@ -29,7 +28,9 @@ When the tag is focused and is removable, it can be removed by pressing the `Del
 
 |Attribute|Value|Description|
 |-|-|-|
-|`role`|`'listitem', 'button'`|Identifies the tag element as a `listitem` when it's used inside the `sl-tag-list` - this role is added automatically. Please provide a role `button` when the tag is interactive, is used to perform an action or is removable and not used inside `sl-tag-list`.|
+|`role`|`'listitem'`|Identifies the tag element as a `listitem` when it's used inside the `sl-tag-list` - this role is added automatically. Removable tags contain a native remove button.|
+|`aria-label`|string|The remove button uses an accessible name that identifies which tag will be removed.|
```
anna-lach: But this is not sth that should be added by user, it's added by us, right? So it should not be added to that table then. We are trying to list only those ARIA atrributes that should be provided by a user.

packages/components/tag/src/tag.stories.ts
a11ymiko: I think it make sense right now when user gets name 'Remove tag X' and info about state 'dimmed'. Users are used to information being announced this way and it's also easier to implement in DS this way rather than changing accessible name of a button. 
So I would recommend to stay with what Michał provided, as there are every needed information there, announced in the way that is understandable and known by screen reader users

website/src/categories/components/tag/usage.md
arecuenco-dsgn: Yes, I agree. I understand why Michal added (to show both alternatives).

Maybe we can have only one code example and add 2 sl-tag_list elements with only 2 tags each.

website/src/categories/components/tag/usage.md
arecuenco-dsgn: Who changed this?

website/src/categories/components/tag/usage.md
arecuenco-dsgn: Who changed this?

website/src/categories/components/tag/usage.md
michal-sanoma: I added this

website/src/categories/components/tag/usage.md
michal-sanoma: I added this

packages/components/tag/src/tag.scss
arecuenco-dsgn: That's correct, Figma was an early version.

Example of Focus within the tags


https://www.figma.com/design/CHpKrPIdXdbV2u7X8vizKI/Components-2.0?node-id=17338-7982&t=5009iM1g30W5XLZ1-4

packages/components/tag/src/tag.scss
arecuenco-dsgn: I show in the comment below. This version of the focus in the tag counter is wrong. We need to apply the same inside focus as the close button.

Example of Focus within the tags

Here is the link to the example in Figma:
https://www.figma.com/design/CHpKrPIdXdbV2u7X8vizKI/Components-2.0?node-id=17338-7982&t=5009iM1g30W5XLZ1-4

website/src/categories/components/tag/usage.md
arecuenco-dsgn: Delete it. It repeats what was already mentioned.

website/src/categories/components/tag/usage.md
arecuenco-dsgn: Keep it.

website/src/categories/components/tag/accessibility.md
arecuenco-dsgn: This isn't on my side. I don't cover what's in the a11y tab.

website/src/categories/components/tag/accessibility.md
arecuenco-dsgn: This isn't on my side. I don't cover what's in the a11y tab.

packages/components/button/CHANGELOG.md
michal-sanoma: We introduced some time ago new rules for prettier and every time when  a file is saved (even with auto save) those rules are applied automatically. I think it doesn't make sense to revert this as this is new standard

.storybook/stories/links.stories.ts
michal-sanoma: https://github.com/sl-design-system/components/pull/3214#discussion_r3485915968

packages/components/tag/src/tag.scss
michal-sanoma: fixed

website/src/categories/components/tag/usage.md
michal-sanoma: fixed

packages/components/tag/src/tag.spec.ts
michal-sanoma: I rephrased this

packages/components/tag/src/tag.scss
michal-sanoma: changed to `visually-hidden`

packages/components/tag/src/tag-list.ts
michal-sanoma: good catch, removed

packages/components/tag/src/tag.ts
michal-sanoma: removed

packages/components/tag/src/tag.ts
michal-sanoma: Since the focus-visible state depends on whether the focused target itself matches `:focus-visible`, I’d prefer keeping focus/blur on the actual focus targets. focusin/focusout would not change the behavior here unless we move the listener to a shared parent

packages/components/tag/src/tag.ts
michal-sanoma: https://github.com/sl-design-system/components/pull/3214#discussion_r3485943234

website/src/categories/components/tag/accessibility.md
michal-sanoma: Makes sense, removed

website/src/categories/components/tag/accessibility.md
michal-sanoma: Fair enough, changed

website/src/categories/components/tag/usage.md
michal-sanoma: Changed to sth like this

Screenshot 2026-06-27 at 13 53 33

packages/components/tag/src/tag.scss
michal-sanoma: fixed

Screenshot 2026-06-27 at 13 59 56

packages/components/tag/src/tag.scss
```diff
@@ -37,36 +37,32 @@
...
-  color: var(--sl-color-foreground-disabled);
-  pointer-events: none;
-
-  slot,
-  button {
-    background: var(--sl-color-background-disabled);
-  }
+:host([variant='info']) button:not([aria-disabled='true']) {
+  color: var(--sl-color-foreground-info-bold);
 }
```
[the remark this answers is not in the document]
michal-sanoma: this is intentional and expected change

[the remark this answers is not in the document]
michal-sanoma: this is intentional and expected change

website/src/categories/components/tag/usage.md
```diff
@@ -8,17 +8,29 @@ eleventyNavigation:
 <section class="no-heading">
 
 <div class="ds-example">
-<sl-tag>Mathematics</sl-tag>
-<sl-tag removable>History</sl-tag>
-<sl-tag disabled removable>Science</sl-tag>
+<sl-tag-list>
```
anna-lach: Could you please add `aria-label` to the tag-list to have the example accessible?

website/src/categories/components/tag/usage.md
```diff
@@ -8,17 +8,29 @@ eleventyNavigation:
...
+  <sl-tag>Mathematics</sl-tag>
+  <sl-tag>History</sl-tag>
+</sl-tag-list>
+
+<sl-tag-list>
```
anna-lach: Could you please add `aria-label` to the tag-list to have the example accessible?

website/src/categories/components/tag/usage.md
```diff
@@ -8,17 +8,29 @@ eleventyNavigation:
...
+</sl-tag-list>
 </div>
 
 <div class="ds-code">
 
   ```html
-<sl-tag>Mathematics</sl-tag>
-<sl-tag removable>History</sl-tag>
-<sl-tag disabled removable>Science</sl-tag>
+<sl-tag-list>
```
anna-lach: Could you please add `aria-label` to the tag-list to have the example accessible?

website/src/categories/components/tag/usage.md
```diff
@@ -8,17 +8,29 @@ eleventyNavigation:
...
+  <sl-tag>Mathematics</sl-tag>
+  <sl-tag>History</sl-tag>
+</sl-tag-list>
+
+<sl-tag-list>
```
anna-lach: Could you please add `aria-label` to the tag-list to have the example accessible?

website/src/categories/components/tag/usage.md
anna-lach: Please also add that missing `aria-label` in the `Code` tab example as well.

packages/components/tag/src/tag-list.spec.ts
```diff
@@ -181,6 +298,55 @@ describe('sl-tag-list', () => {
...
+      `);
+
+      await new Promise(resolve => setTimeout(resolve, 60));
+
+      const stackTag = el.renderRoot.querySelector('sl-tag')!,
+        visibleTags = Array.from(el.querySelectorAll('sl-tag')).filter(
+          tag => getComputedStyle(tag).display !== 'none'
+        );
+
+      expect(stackTag).to.have.attribute('disabled');
```
anna-lach: stack tag should be never disabled (always focusable), right?

packages/components/combobox/src/combobox.ts
```diff
@@ -504,6 +504,7 @@ export class Combobox<T = any, U = T> extends ObserveAttributesMixin(
         ${this.multiple && this.selectedItems.length
           ? html`
               <sl-tag-list
+                @focusin=${this.#onTagListFocusIn}
```
anna-lach, line 507: When focus leaves the stack tag there is a focus ring outside of the tag visible for a while, it should not be visible in that case.

https://github.com/user-attachments/assets/111a9e29-a1e2-4689-9ea7-8e623f658f1b

packages/components/tag/src/tag.scss
```diff
@@ -84,32 +80,58 @@ button {
...
+  clip-path: inset(50%);
+  inline-size: 1px;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
 }
 
 sl-icon {
-  --sl-icon-size: var(--sl-size-new-icon-xs);
+  --sl-icon-size: var(--sl-size-new-icon-sm);
```
anna-lach: For 'md' size of tag there should be smaller icon used 12 x 12px

Image

packages/components/tag/src/tag.ts
```diff
@@ -97,71 +106,113 @@ export class Tag extends ScopedElementsMixin(LitElement) {
    */
   @property({ reflect: true }) variant?: TagVariant;
 
+  override get tabIndex(): number {
```
anna-lach, line 110: Probably should be marked as `@internal`?
Right now it's visible on the website:

Image

anna-lach (approved):

website/src/categories/components/tag/usage.md
michal-sanoma: added

website/src/categories/components/tag/usage.md
michal-sanoma: added

website/src/categories/components/tag/usage.md
michal-sanoma: added

website/src/categories/components/tag/usage.md
michal-sanoma: added

packages/components/tag/src/tag-list.spec.ts
michal-sanoma: fixed

packages/components/tag/src/tag.scss
michal-sanoma: fixed

packages/components/tag/src/tag.ts
michal-sanoma: fixed

packages/components/combobox/src/combobox.ts
michal-sanoma: fixed

packages/components/tag/src/tag-list.ts
```diff
@@ -135,17 +138,29 @@ export class TagList extends ScopedElementsMixin(LitElement) {
...
+        ...stackTags,
+        ...(this.tags ?? []).filter(t => t.style.display !== 'none' && !!t.removable)
+      ];
+    },
+    isFocusableElement: (el: Tag) => this.#isFocusableElement(el)
   });
 
-  /** Disables interaction with the tag list and renders the stacked tag as disabled. */
+  /** Disables interaction with the tag list. */
   @property({ type: Boolean }) disabled?: boolean;
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/combobox/src/combobox.ts
```diff
@@ -707,6 +707,10 @@ export class Combobox<T = any, U = T> extends ObserveAttributesMixin(
   }
 
   #onKeydown(event: KeyboardEvent): void {
+    if (!event.composedPath().includes(this.input)) {
+      return;
+    }
+
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/tag/src/tag.scss
```diff
@@ -17,56 +17,56 @@
...
   position: relative;
   z-index: 1; // Make sure the focus ring is above other elements
 }
 
-slot {
+[part='label'] {
   background: var(--_bg-color);
   display: block;
   flex: 1;
+  font: var(--sl-text-new-body-md);
```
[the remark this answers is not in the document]
michal-sanoma: fixed

PR #3219, opened 2026-04-15
fix(button): fix link styling in button

### LINKED ISSUE #3218: [Button] global `<a href>` styling breaks link button

Should look like this: 

Image

But looks like this:

Image

We should override any styling on `::slotted(a)`.

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

jpzwarte:
Fixes #3218

## Summary

Fixed styling of slotted `<a>` elements inside `<sl-button>`. Global CSS from the application could override the link color and text decoration, causing the button to render incorrectly. Styles are now explicitly reset using `::slotted(a)` to ensure the button's own color and text-decoration styles are always applied.

## Changes

- Added `::slotted(a)` rule to reset `color` and `text-decoration` on slotted anchor elements
- Moved `button:disabled { pointer-events: none }` inside the `button {}` rule scope
- Added a new `Link` story to demonstrate `<sl-button fill="link">` with a slotted `<a href>`
- Updated the `All` story to use `<a href>` elements inside link-fill buttons

### DIFF
--- a/packages/components/button/src/button.stories.ts
+++ b/packages/components/button/src/button.stories.ts
@@ -161,20 +161,6 @@ export const IconOnly: Story = {
   }
 };
 
-export const Link: Story = {
-  render: ({ disabled, shape, size, text, variant }) => html`
-    <sl-button
-      ?disabled=${disabled}
-      fill="link"
-      shape=${ifDefined(shape)}
-      size=${ifDefined(size)}
-      variant=${ifDefined(variant)}
-    >
-      <a href="https://example.com">${text}</a>
-    </sl-button>
-  `
-};
-
 export const Pill: Story = {
   args: {
     shape: 'pill'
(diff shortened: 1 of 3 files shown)


packages/components/button/src/button.stories.ts
```diff
@@ -161,6 +161,20 @@ export const IconOnly: Story = {
   }
 };
 
+export const Link: Story = {
```
Diaan: we certainly shouldn't add this story; we shouldn't encourage people to do this! 😁

packages/components/button/src/button.stories.ts
```diff
@@ -385,42 +399,42 @@ export const All: Story = {
         <span>Link</span>
         <sl-button fill="link" variant="primary">
           <sl-icon name="far-universal-access"></sl-icon>
-          Button
+          <a href="https://example.com">Button</a>
```
Diaan: 🙂‍↔️

Diaan (approved):

PR #3220, opened 2026-04-15
Fix/3013 avatar sizes variant has badge what does not have role specified

### LINKED ISSUE #3013: [Avatar] Sizes variant has badge what does not have `role` specified
## Current behavior
In [Avatar Sizes](https://jolly-stone-0f7fffc03-3000.westeurope.6.azurestaticapps.net/?path=/story/media-avatar--sizes) every `<sl-badge>` do not have `role` specified, which makes it forbidden to use `aria-label` on `<sl-badge>`.
Image

## Proposed solution
Adding `role="text"` to each `<sl-badge>` should fix the issue that is reported by a11y add-on in Storybook. 

### 👤 Your name

Kacper

### 🧢 Your Product/Team

SLDS

anna-lach:
This pull request enhances the accessibility of avatar badges, especially for dynamic content, and updates the Storybook stories and documentation to demonstrate and explain these improvements. The main focus is on ensuring that screen readers correctly announce badge updates, following best practices for ARIA roles and visually hidden text.

**Accessibility improvements for avatar badges:**

- Updated the `Sizes` story in `avatar.stories.ts` to demonstrate dynamic badge content for `sm` and `4xl` avatars. These badges now use `role="status"` and include a visually hidden `<span>` with descriptive text, ensuring screen readers announce count changes. The story dynamically updates the badge count every 5 seconds to showcase this behavior. [[1]](diffhunk://#diff-d6f9f2eec3a6a96b8c4f70fe5bfb6ca233642d91a0ac10ba559e3a07a505db2bL216-R217) [[2]](diffhunk://#diff-d6f9f2eec3a6a96b8c4f70fe5bfb6ca233642d91a0ac10ba559e3a07a505db2bL231-R319) [[3]](diffhunk://#diff-d6f9f2eec3a6a96b8c4f70fe5bfb6ca233642d91a0ac10ba559e3a07a505db2bL249-R351)
- For static badges in other sizes, removed unnecessary ARIA attributes and added a visually hidden `<span>` for context, so screen readers announce the badge content when navigated to, but do not treat it as a live region.

**Documentation updates:**

- Expanded the accessibility documentation (`avatar/accessibility.md`) with clear guidance and code examples for using `sl-badge` inside avatars, covering both dynamic and static badge content scenarios. Included links to Storybook examples for further reference. [[1]](diffhunk://#diff-7d035748ec7da665a2ce4494c7f425f23b679ea5fe1440116c8e922e138a320aR11-R48) [[2]](diffhunk://#diff-7d035748ec7da665a2ce4494c7f425f23b679ea5fe1440116c8e922e138a320aR60)


packages/components/avatar/src/avatar.stories.ts
```diff
@@ -203,52 +204,111 @@ export const Overflow: Story = {
...
+
+        if ((badge as HTMLElement & { size: string }).size !== 'sm') {
+          const textNode = Array.from(badge.childNodes).find(
+            node => node.nodeType === Node.TEXT_NODE && node.textContent?.trim() !== ''
+          );
+          if (textNode) {
+            textNode.textContent = `${count}`;
+          }
+        }
+      });
```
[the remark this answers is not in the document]
anna-lach: @a11ymiko what's your opinion? Maybe we should use our `announcer` component here?

packages/components/avatar/src/avatar.stories.ts
```diff
@@ -203,52 +204,109 @@ export const Overflow: Story = {
...
+        intervalPerCanvas.delete(canvasElement);
+        observer.disconnect();
+
+        return;
+      }
+
+      countPerCanvas.set(canvasElement, count);
+
+      const badges = canvasElement.querySelectorAll('sl-badge');
+      badges.forEach(badge => {
```
michal-sanoma: **suggestion:** I tested this with VoiceOver in Storybook and it looks like we’re duplicating announcements here. Since all badges in Sizes are role="status" and update together, VoiceOver reads out multiple very similar messages every 5 seconds. Maybe we should keep only one dynamic example and leave the rest static?

michal-sanoma: https://github.com/user-attachments/assets/56dfddbe-522e-4747-a2e5-1df378f5fee9

anna-lach: That's a good point, @a11ymiko what do you think?

a11ymiko: That's a good idea

packages/components/avatar/src/avatar.stories.ts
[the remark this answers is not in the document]
a11ymiko: yeah, using text for announcement would act more predictable with different assistive technologies. 
Could you please prepare version using announcer that is inside badge? 
I have to confirm that this combination would result in only announcer text being announced after each update and `aria-label` won't be announced as duplicate of announcer content but can be reached with screen reader manually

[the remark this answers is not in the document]
anna-lach: I've added the `announcer` in the `Sizes` story. @a11ymiko can you please take a look at it?

packages/components/avatar/src/avatar.stories.ts
anna-lach: I've added the `announcer` in the `Sizes` story, now the change is read once.

michal-sanoma (approved):

packages/components/avatar/src/avatar.stories.ts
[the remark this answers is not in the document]
anna-lach: Discussed with Kacper, we've changed the solution, so we are not using `announcer` here anymore.

packages/components/avatar/src/avatar.stories.ts
anna-lach: Discussed with Kacper, we've changed the solution, so we are not using `announcer` here anymore.
Right now the change is read twice, for `sm` and `4xl` example to show different scenarios, when there is a text inside the badge and no text in the badge.

michal-sanoma (approved):

a11ymiko (approved):

PR #3240, opened 2026-04-20
fix(select): observe and update selected option content

### LINKED ISSUE #3239: [Select] Selected option content does not update when mutated in runtime
### 😯 Current Behavior

When an `sl-option`'s text or child content **is mutated** after it has **already been selected**, the closed `sl-select` button continues to show the stale label. 

In addition - the selected option content inside the dropdown **is updated**.

Image

Selected option content is always up-to-date with how it was mutated in runtime

### 🖥️ Steps to Reproduce

1. Select an option in an `sl-select`
2. Modify the selected option's content in runtime

Yes

_No response_

_No response_

### ⚖️ Impact

### Framework

Angular

Chrome

### If other, please specify.

_No response_

_No response_

### Component version

2.1.3

### Theme version

_No response_

### 👤 Your name

Bartosz Wasilew

### 🧢 Your product/team

_No response_

### 🎨 Your Theme(s)

Sanoma Learning

bartosz-wasilew-sl:
Fixes #3239

### DIFF
--- a/packages/components/select/src/select.ts
+++ b/packages/components/select/src/select.ts
@@ -236,6 +236,7 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
 
     this.#observer.observe(this, { childList: true, subtree: true });
     this.#observeSelectedOptionContent();
+    this.#onSelectedOptionContentChange();
 
     // Listen for i18n updates and update the validation message
     this.#events.listen(window, LOCALE_STATUS_EVENT, this.#updateValueAndValidity);
@@ -760,7 +761,9 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
     this.#selectedOptionObserver.observe(this.selectedOption, {
       childList: true,
       subtree: true,
-      characterData: true
+      characterData: true,
+      attributes: true,
+      attributeFilter: ['value']
     });
   }
 
--- a/packages/components/select/src/select.ts
+++ b/packages/components/select/src/select.ts
@@ -99,6 +99,9 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
   /** Detect when options are added to the host, or a nested option group and clear the cache. */
   #observer = new MutationObserver(() => this.#rovingTabindexController.clearElementCache());
 
+  /** Detect when the selected option content changes, so the button can refresh its cloned content. */
+  #selectedOptionObserver = new MutationObserver(() => this.#onSelectedOptionContentChange());
+
   /** Since we can't use `popovertarget`, we need to monitor the closing state manually. */
   #popoverClosing = false;
 
@@ -236,6 +239,7 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
 
   override disconnectedCallback(): void {
     this.#observer.disconnect();
+    this.#selectedOptionObserver.disconnect();
 
     super.disconnectedCallback();
   }
@@ -414,6 +418,16 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
     }
 
     this.#lastRenderedOption = this.selectedOption;
+  }
+
+  #onSelectedOptionContentChange(): void {
+    if (!this.selectedOption) {
+      return;
+    }
+
+    this.#lastRenderedOption = undefined;
+    this.#renderSelectedContent();
+    this.#calculateLargestOptionWidth();
   }
 
   #onBeforetoggle({ newState }: ToggleEvent): void {
@@ -694,6 +708,8 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
   }
 
   #setSelectedOption(option?: Option<T>, emitEvent = true): void {
+    this.#selectedOptionObserver.disconnect();
+
     if (this.selectedOption) {
       this.selectedOption.selected = false;
       this.selectedOption.setAttribute('aria-selected', 'false');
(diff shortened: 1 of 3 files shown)


packages/components/select/src/select.ts
```diff
@@ -236,6 +239,7 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
 
   override disconnectedCallback(): void {
     this.#observer.disconnect();
+    this.#selectedOptionObserver.disconnect();
```
michal-sanoma, line 247: **suggestion:** Nice fix overall. One edge case: we disconnect `#selectedOptionObserver` in `disconnectedCallback()`, but I don’t see it being re-attached on `connectedCallback()` when `selectedOption` is already set. 

If the component is detached and re-attached without changing `value`, runtime mutations of the selected option may stop updating the button content. Could we re-observe on connect and add a regression test for detach/attach?

packages/components/select/src/select.ts
```diff
@@ -416,6 +420,16 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
...
   }
 
+  #onSelectedOptionContentChange(): void {
+    if (!this.selectedOption) {
+      return;
+    }
+
+    this.#lastRenderedOption = undefined;
+    this.#renderSelectedContent();
+    this.#calculateLargestOptionWidth();
```
michal-sanoma: **question:** Small performance question: `#onSelectedOptionContentChange()` recalculates largest option width on every mutation. For frequently changing content this could trigger repeated layout work. Would it make sense to batch/debounce this recalculation (e.g. microtask/rAF) or gate it more narrowly?

bartosz-wasilew-sl: That's a good one. Sure - performance-wise - it costs.

But I cannot imagine how it could work without re-calculating the largest width, when we (as per purpose of this feature) update content (that can be "very different" in terms of width, comparing to the one before an update). Example: runtime translations.

Do you see a scenario when the number of "needed re-calculations" is lower than "each time we change content"?

packages/components/select/src/select.ts
bartosz-wasilew-sl: Thanks.

Indeed we disconnect the `#selectedOptionObserver` as we disconnect the whole `Select` - treat it as a cleanup.

And indeed we re-attach/re-connect, but only when the `selectedOption` is available (`#selectedOptionObserver`).

The `connectedCallback()` "hook" concerns the `Select` overall - and yet, at the component init - there's no `selectedOption` === nothing to observe yet.

michal-sanoma: Thanks for clarifying, I'm still new to Lit.js so maybe I'm missing sth but we may still be missing the re-subscribe path for the detach/attach case when `value` does not change.

From what I see, we disconnect in `disconnectedCallback()`, and `observe(...)` is only re-attached inside `#setSelectedOption(...)`. If the same selected option remains selected across re-attach, `#setSelectedOption(...)` may not run, so runtime mutations would no longer refresh selected content.

packages/components/select/src/select.ts
michal-sanoma: Makes sense, and I agree we need width recalculation for correctness.

My concern is only about frequency, not removing recalculation: one logical content update can produce multiple MutationObserver records in the same tick (e.g. runtime i18n / framework patching several text nodes). In that case we currently run full width measurement multiple times, even though one final measurement would be enough

packages/components/select/src/select.ts
```diff
@@ -416,6 +420,16 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
...
+  }
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/select/src/select.ts
michal-sanoma: fixed here https://github.com/sl-design-system/components/pull/3240/changes/42384a4b9d064edcdebf8d5bd76cfeadc92045c7

packages/components/select/src/select.ts
michal-sanoma: fixed here: https://github.com/sl-design-system/components/pull/3240/changes/42384a4b9d064edcdebf8d5bd76cfeadc92045c7

packages/components/select/src/select.ts
```diff
@@ -721,6 +750,31 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
...
+    this.#selectedOptionObserver.disconnect();
+
+    if (!this.selectedOption) {
+      return;
+    }
+
+    this.#selectedOptionObserver.observe(this.selectedOption, {
+      childList: true,
+      subtree: true,
+      characterData: true
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/select/src/select.ts
```diff
@@ -229,13 +235,19 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
     this.setAttributesTarget(this.button);
 
     this.#observer.observe(this, { childList: true, subtree: true });
+    this.#observeSelectedOptionContent();
```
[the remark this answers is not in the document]
michal-sanoma: fixed

packages/components/select/src/select.ts
```diff
@@ -416,6 +429,22 @@ export class Select<T = any> extends ObserveAttributesMixin(FormControlMixin(Sco
...
+
+    const selectedOptionValue = this.selectedOption.value;
+    if (selectedOptionValue !== this.value) {
+      this.value = selectedOptionValue;
+      this.#updateValueAndValidity();
+    }
+
+    this.#lastRenderedOption = undefined;
+    this.#renderSelectedContent();
+    this.#scheduleLargestOptionWidthCalculation();
```
Diaan, line 451: does it need this? The list of options doesn't change here, right? Why do we need to recalculate the width?

Diaan (approved):

packages/components/select/src/select.ts
michal-sanoma: true, I've added an early return

PR #3251, opened 2026-04-24
Replace Prettier with oxfmt

### LINKED ISSUE #3188: [Maintenance] Investigate replacing prettier with oxfmt
## Background

[oxfmt](https://oxc.rs/docs/guide/usage/formatter.html) is a high-performance JavaScript/TypeScript formatter built on the [OXC](https://oxc.rs/) compiler stack. It may be a worthwhile replacement for Prettier given its speed improvements and additional built-in features.

## Motivation

- **~30x faster than Prettier** and ~2x faster than Biome according to their benchmarks — meaningful improvement for CI and local DX
- **100% Prettier conformance** for JavaScript and TypeScript, making migration low-risk
- **Built-in extras** that currently require separate tooling or plugins:
  - Import sorting
  - Tailwind CSS class organization
  - `package.json` field sorting
  - Embedded code formatting (CSS-in-JS, GraphQL, etc.)
- Broad language support: JS, TS, JSX, TSX, JSON, YAML, TOML, HTML, Vue, CSS preprocessors, Markdown, and more

## Tasks

- [ ] Evaluate current maturity/stability of oxfmt (it's still under active development)
- [ ] Check support for all file types we currently format with Prettier (`.ts`, `.html`, `.css`, `.json`, `.md`, etc.)
- [ ] Test output compatibility — confirm formatting output matches or is acceptable vs. current Prettier output
- [ ] Check editor integration (VS Code extension, etc.)
- [ ] Assess whether built-in import sorting could replace our current import sorter setup
- [ ] Assess Tailwind CSS class sorting support as a potential replacement for `prettier-plugin-tailwindcss`
- [ ] Plan migration path if deemed ready

## References

- oxfmt docs: https://oxc.rs/docs/guide/usage/formatter.html

jpzwarte:
Closes #3188

## Summary

Replaces Prettier with [oxfmt](https://oxc.rs/docs/guide/usage/formatter.html) as the code formatter across the repository. oxfmt is the Rust-based formatter from the OXC toolchain, offering significantly faster formatting speeds while producing consistent output.

### Tooling
- Removes `.prettierrc.cjs` and Prettier dependency from root `package.json`
- Removes Prettier from `@sl-design-system/eslint-config` and `@sl-design-system/stylelint-config`
- Adds `.oxfmtrc.json` with oxfmt configuration
- Updates `.lintstagedrc.mjs` to run `oxfmt format` on staged files instead of Prettier

### VS Code
- Adds oxfmt VS Code extension to `.vscode/extensions.json`
- Configures oxfmt as the default formatter in `.vscode/settings.json`

### Reformatting
- Reformats all TypeScript, JavaScript, JSON, SCSS, and Markdown files across the entire repository to match oxfmt output


package.json
```diff
@@ -133,7 +133,7 @@
       ]
     },
     "build:locales": {
-      "command": "node scripts/build-packages.js locales",
+      "command": "node scripts/build-packages.js locales && oxfmt packages/locales/src",
```
michal-sanoma, line 136: Nice migration! 🚀  One small thing I noticed: we run `oxfmt packages/locales/src` in both `build:locales:code` and `build:locales`. Won't this double the work?

.oxfmtrc.json
```diff
@@ -0,0 +1,18 @@
...
+  "ignorePatterns": [
+    "*dist",
+    "*generated",
+    "packages/components/shared/src/vendor",
+    "packages/tokens/src/tokens",
+    "website"
+  ],
+  "insertFinalNewline": true,
+  "jsdoc": true,
+  "printWidth": 100,
```
michal-sanoma, line 13: **question:** Quick question on formatting settings: `printWidth` changed from 120 (Prettier) to 100. Was that intentional?

.editorconfig
```diff
@@ -1,13 +0,0 @@
-root = true
```
michal-sanoma, line 1: **thought:** removing `.editorconfig` makes sense if we fully standardize on oxfmt, but it also removes a useful cross-editor fallback (especially for contributors not using VSCode/oxc). Would you consider keeping a minimal `.editorconfig` for basics like indent/newline consistency?

michal-sanoma (approved):

.oxfmtrc.json
jpzwarte: It was, it makes the code more readable imo, and it's also the default of prettier/oxfmt.

package.json
jpzwarte: `oxfmt` is so quick that you won't notice. If we don't do this, the the formatter get's overridden by the compile step.

PR #3272, opened 2026-05-05
fix(combobox): reset filtered results on blur when no option is selected

### LINKED ISSUE #3268: [COMBOBOX] Retains filter value

The search value of a `sl-combobox` is retained when leaving the control. Use these steps:

- Enter a search value, do not select this value from the list. 
- Leave the combobox by selecting an other control or use the Tab key. 
- Return to the combobox field and the list show only the result of the previous search action, but the control is empty (also see screenshot)

Image

story to reproduce this issue

```typescript
export const FilterResults: Story = {
  render: () => {
    return html`
     <sl-combobox
        filter-results
        .options=${components}
      >
      </sl-combobox>

      <sl-text-field></sl-text-field>    
      `
  }
};
```

### 👤 Your name

Frans de Boer

### 🧢 Your Product/Team

Magister

michal-sanoma:
## Summary
This PR fixes a bug in `sl-combobox` (`filter-results`) where filtering state could persist after the user leaves the field without selecting an option.

## Problem
When a user typed a search value, did not select anything, and moved focus away (click or Tab), the visible input could be reset but the options list remained filtered internally.  
After returning to the combobox, the list showed only results from the previous search although the field looked empty.

## Root Cause
On `focusout`, the component restored the text-field value but did not reset filtered option visibility.

### BEFORE

https://github.com/user-attachments/assets/c05f98fb-256a-40e5-862f-d88e5e0b7195

### AFTER

https://github.com/user-attachments/assets/8d67f603-0bef-41c4-a81c-d739d0345ec7

### DIFF
--- a/.changeset/soft-rocks-dig.md
+++ b/.changeset/soft-rocks-dig.md
@@ -1,0 +1,10 @@
+---
+'@sl-design-system/combobox': patch
+---
+
+Fix `filter-results` behavior in `sl-combobox` when leaving the field without selecting an option.
+
+Previously, typing a search value and blurring the component could clear the input while keeping the internal filtered state, so reopening the list showed only stale filtered results.
+Now, when focus leaves the component, filtered option visibility is reset to match the restored input value.
+
+Also adds a regression test for the flow: type search text -> leave combobox (click/Tab) -> return to combobox.
--- a/.changeset/soft-rocks-dig.md
+++ b/.changeset/soft-rocks-dig.md
@@ -6,5 +6,3 @@ Fix `filter-results` behavior in `sl-combobox` when leaving the field without se
 
 Previously, typing a search value and blurring the component could clear the input while keeping the internal filtered state, so reopening the list showed only stale filtered results.
 Now, when focus leaves the component, filtered option visibility is reset to match the restored input value.
-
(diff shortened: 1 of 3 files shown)


a11ymiko (approved):

.changeset/soft-rocks-dig.md
```diff
@@ -0,0 +1,10 @@
+---
+'@sl-design-system/combobox': patch
+---
+
+Fix `filter-results` behavior in `sl-combobox` when leaving the field without selecting an option.
+
+Previously, typing a search value and blurring the component could clear the input while keeping the internal filtered state, so reopening the list showed only stale filtered results.
+Now, when focus leaves the component, filtered option visibility is reset to match the restored input value.
+
+Also adds a regression test for the flow: type search text -> leave combobox (click/Tab) -> return to combobox.
```
anna-lach: I'm not sure if the information about regression test is sth important to our users, I guess it's more an information for us, so I think it should not be part of the changeset.

anna-lach (approved):

.changeset/soft-rocks-dig.md
michal-sanoma: true, removed

PR #3339, opened 2026-05-15
chore: remove the checklist package

### LINKED ISSUE #2735: [Checklist] Is it worth the cost?

The checklist was a really nice idea, but there are doubts to its usefulness. A poll in Slack resulted in 1 out of 7 people saying they've actually used it. The rest have never used it, or don't know of its existence.

The component is very different than other regular components. So there is an added cost to maintainability for it.

Let's check whether it's worth the cost to keep the component around. I suggest contacting the person who used the component.

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

jpzwarte:
Closes #2735 

This removes the source code for the `@sl-design-system/checklist` package. The published versions will remain, but there won't be any new versions published, due to lack of use.


packages/checklist/package.json
```diff
@@ -1,47 +0,0 @@
-{
-  "name": "@sl-design-system/checklist",
```
anna-lach, line 2: Since we are removing the whole package there should be no changeset, right? But how we're going to inform people that the package has been removed? I know people are not using it, but still. :)

anna-lach (approved):

packages/checklist/package.json
jpzwarte: Via release notes

PR #3365, opened 2026-05-20
fix(panel): Fix collapsed Panel content being exposed to screen readers

### LINKED ISSUE #3046: [Panel] For collapsible variants screen reader can read collapsed panel content.
## Current behavior

Screen reader users that interact with collapsed Panel can set screen reader's focus on invisible, collapsed panel content and read it. To do it user have to use "down arrow" in NVDA or "Control + Option + right arrow" in VoiceOver to read inactive elements.

https://github.com/user-attachments/assets/f71dca1d-8726-483d-9f0d-bf97379563d8

https://github.com/user-attachments/assets/f8e18bdd-7d04-4686-9186-0398f748229b

Issue occurs in each collapsible version of Panel which includes: 
1. [Collapsible](https://storybook.sanomalearning.design/?path=/story/layout-panel--collapsible)
2. [Collapsed](https://storybook.sanomalearning.design/?path=/story/layout-panel--collapsed)
3. [No Padding Content](https://storybook.sanomalearning.design/?path=/story/layout-panel--no-padding-content)
4. [Toggle Externally](https://storybook.sanomalearning.design/?path=/story/layout-panel--toggle-externally)
5. [All](https://storybook.sanomalearning.design/?path=/story/layout-panel--all)

## Suggested solution

It's recommended to hide collapsed panel's content from screen reader. Common solution is to use CSS "display: none;" or "visibility: hidden;" as it hides content both visualy and from screen reader.
But other solutions, for example this used in [Accordion Basic](https://storybook.sanomalearning.design/?path=/story/layout-accordion--basic) also should work, as long as screen reader does not announce collapsed content.

### 👤 Your name

Kacper

### 🧢 Your Product/Team

SLDS

michal-sanoma:
## Summary
This PR fixes an accessibility issue in collapsible sl-panel variants where screen reader users could navigate into visually collapsed panel content.

### Changes:
- Adds aria-hidden="true" to the panel body when a collapsible panel is collapsed
- Adds inert to prevent focus/navigation into collapsed content
- Removes both attributes again when the panel is expanded
- Adds regression tests for toggled and initially collapsed panels

### BEFORE

https://github.com/user-attachments/assets/561406a8-6ceb-4158-8fba-b34c76cc5a87

### AFTER

https://github.com/user-attachments/assets/8906f657-2d0c-4840-ab39-f61f72c06089

### DIFF
--- a/.changeset/stupid-drinks-refuse.md
+++ b/.changeset/stupid-drinks-refuse.md
@@ -1,0 +1,5 @@
+---
+'@sl-design-system/panel': patch
+---
+
+Hide collapsed panel content from assistive technology
--- a/packages/components/panel/src/panel.spec.ts
+++ b/packages/components/panel/src/panel.spec.ts
@@ -112,6 +112,39 @@ describe('sl-panel', () => {
       expect(body).to.have.attribute('aria-labelledby', 'heading');
     });
 
+    it('should hide the body from assistive technology when collapsed', async () => {
+      const body = el.renderRoot.querySelector('[part="body"]');
+
+      expect(body).not.to.have.attribute('aria-hidden');
+      expect(body).not.to.have.attribute('inert');
+
+      el.toggle(true);
+
+      await new Promise(resolve => requestAnimationFrame(resolve));
+      await el.updateComplete;
+
+      expect(body).to.have.attribute('aria-hidden', 'true');
+      expect(body).to.have.attribute('inert');
+
+      el.toggle(false);
+
+      await new Promise(resolve => requestAnimationFrame(resolve));
+      await el.updateComplete;
+
+      expect(body).not.to.have.attribute('aria-hidden');
+      expect(body).not.to.have.attribute('inert');
+    });
+
+    it('should hide initially collapsed body from assistive technology', async () => {
+      const collapsedEl = await fixture<Panel>(
+        html`<sl-panel collapsible collapsed heading="Heading">Body content</sl-panel>`
+      );
+      const body = collapsedEl.renderRoot.querySelector('[part="body"]');
+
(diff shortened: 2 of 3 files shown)


a11ymiko (approved):

.changeset/stupid-drinks-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sl-design-system/panel': patch
+---
+
+Hide collapsed panel content from assistive technology
```
anna-lach, line 5: Should we also hide the tool-bar for the screen reader users when there is nothing in it? Or it's not necessary?

Image

https://github.com/user-attachments/assets/b44d8ee4-e3c6-4461-8b68-292092bfdbf7

a11ymiko: I saw it as well and wanted to add an issue with P3 priority for that. This currently exists in our storybook in more stories (e.g. [Panel - No Padding](https://storybook.sanomalearning.design/?path=/story/layout-panel--no-padding-content)) so I didn't add it as issue with this PR. 
It's not a big issue but it'd be more clear if we present to user only containers/regions that contain at least one element. 
In this case when the tool bar is empty I would recommend either to remove it completely from DOM or to hide it with `aria-hidden` or to remove `role='toolbar'` from `sl-toobar` when it's empty. Each of this option would be as accessible so you can choose what is the best for you.

anna-lach (approved):

.changeset/stupid-drinks-refuse.md
a11ymiko: Issue created for this one: https://github.com/sl-design-system/components/issues/3369

PR #3368, opened 2026-05-21
feat(tooltip): simplify implementation

### LINKED ISSUE #3049: [Menu button] Esc key on expanded menu with tooltip closes menu not tooltip
## Current behavior

In story [Menu button With Tooltips](https://storybook.sanomalearning.design/?path=/story/actions-menu-menu-button--with-tooltips) cursor users that expand menu and after that make tooltip visible cannot dismiss tooltip with `esc` key. `Esc` key collapses expanded menu, but tooltip stays visible. It disappears only after moving cursor out of menu button. In some cases tooltip can obscure menu items, so `esc` option for dismissing it would be helpful.
This behavior may be problematic for people with limited dexterity, shaking hands and people who use magnifiers/zoom attached to their cursor.

https://github.com/user-attachments/assets/602f7d5f-5ec2-45c4-9860-c1bd6e7248b3

## Suggested solution

It's recommended to add mechanism that checks if tooltip is visible and all the time it's visible it make sure that `esc` dismisses the tooltip. 
So in situation in which tooltip is presented on expanded menu `esc` key should in first step dismiss tooltip, and second use of `esc` should collapse expanded menu.

### 👤 Your name

Kacper

### 🧢 Your Product/Team

SLDS

### LINKED ISSUE #3255: [Tooltip] Simplify tooltip architecture while preserving current behavior and use cases

Our current tooltip implementation has grown significantly in complexity because it needs to support multiple advanced scenarios at once, including:

- tooltips for components with internal focusable elements inside shadow DOM
- shared tooltips reused across multiple anchors
- tooltips used in slotted / reparented contexts such as grid bulk actions and toolbars
- lazy-created tooltips
- hover, focus, and manual opening flows
- accessibility wiring across host elements, proxy targets, and ElementInternals

Although the current implementation works for these cases, it has become difficult to understand, maintain, and safely extend. We should investigate a simplification of the tooltip architecture that keeps all currently supported behavior, but makes the implementation easier to reason about.

This task should use the Web Awesome tooltip approach as inspiration, especially the idea of moving more tooltip behavior closer to the actual component/anchor, instead of relying on complex global discovery and reparenting logic.

### Goal
Design and implement a simpler tooltip architecture that preserves all existing functionality, but reduces implementation complexity and improves maintainability.

Inspiration:
https://github.com/shoelace-style/webawesome/blob/next/packages/webawesome/src/components/tooltip/tooltip.ts

### 👤 Your name

Michał

### 🧢 Your Product/Team

_No response_

### LINKED ISSUE #3264: [Tool bar] Tooltips in Tooltips and Icon Only stories are not announced by screen readers
### 😯 Current Behavior

https://github.com/user-attachments/assets/fd365e6c-2d52-402f-ae37-bb1d0654c76f

In DevTools inside Accessibility tab there is no value provided for `aria-labelledby`:
Image

https://github.com/user-attachments/assets/72b7161c-37e6-4edb-8f72-81a29d0e0c68

Tooltip content should be announced as a name of each button.

https://github.com/user-attachments/assets/e5a12812-e015-4d4e-977b-d16d6a307dbc

Image

### 🖥️ Steps to Reproduce

1. Go to https://storybook.sanomalearning.design/?path=/story/actions-tool-bar--tooltips or https://storybook.sanomalearning.design/?path=/story/actions-tool-bar--icon-only
2. Open DevTools.
3. In Accessibility tab check if buttons with tooltip has `aria-labelledby` with value added 
4. With screen reader turned on go to buttons with tooltips. 
5. Check if tooltip content is announced as button name.

Yes

https://storybook.sanomalearning.design/?path=/story/actions-tool-bar--tooltips and https://storybook.sanomalearning.design/?path=/story/actions-tool-bar--icon-only

_No response_

### ⚖️ Impact

### Framework

Other

All

### If other, please specify.

_No response_

_No response_

### Component version

v0.2.3

### Theme version

_No response_

### 👤 Your name

Kacper

### 🧢 Your product/team

SLDS

### 🎨 Your Theme(s)

Sanoma Learning

### LINKED ISSUE #3274: [Tool bar] Tooltips in stories Tooltips and IconOnly cannot be collapsed with `Esc` key when triggered with mouse
### 😯 Current Behavior

When user expand tooltip with mouse cursor using `Esc` key should collapse that tooltip. This does not work right now. User can collapse tooltip with `Esc` only when tooltip is triggered by focus not hover.

https://github.com/user-attachments/assets/76dfaf6d-6f2b-4f30-a24a-4f2edf548eba

https://github.com/user-attachments/assets/01d44767-3212-4aaa-9213-4503c8f7f7f6

This should work like in Toggle group Basic story for example: 

https://github.com/user-attachments/assets/3cf8408e-9c89-4d4e-96d8-9160280a84f5

### 🖥️ Steps to Reproduce

1. Go to https://storybook.sanomalearning.design/?path=/story/actions-tool-bar--tooltips or https://storybook.sanomalearning.design/?path=/story/actions-tool-bar--icon-only
2. Hover over buttons that triggers tooltips
3. When tooltip is visible click `Esc` 
4. Check if tooltip disappears after clicking `Esc`

Yes

https://storybook.sanomalearning.design/?path=/story/actions-tool-bar--tooltips

_No response_

### ⚖️ Impact

### Framework

Other

All

### If other, please specify.

_No response_

_No response_

### Component version

v.0.0.14

### Theme version

_No response_

### 👤 Your name

Kacper

### 🧢 Your product/team

SLDS

### 🎨 Your Theme(s)

Sanoma Learning

### LINKED ISSUE #3275: [Tool bar] `aria-labelledby` pointing to tooltip content is not announced for stories Tooltips and Icon Only
### 😯 Current Behavior

Right now user gets no announcement with buttons names that are provided via `aria-labelledby` in stories Tool bar Tooltips and Tool bar Icon Only

https://github.com/user-attachments/assets/40b3b264-715e-4992-98ba-49a228239625

Image

Partial fix is introduced in PR #3273, but usage of `aria-label` instead of `aria-labelledby` means that whenever you change the tooltip text, you must also remember to update the `aria-label` attribute. This can lead to potential issues in the feature as tooltip would say something different than `aria-label`.

Improved version of behavior introduced in PR #3273 should only rely on `aria-labelledby` as it can be observed in [Button Icon Only story](https://storybook.sanomalearning.design/?path=/story/actions-button--icon-only):

https://github.com/user-attachments/assets/d75615ca-090e-480d-8cf3-b7930e7c61de

Image

### 🖥️ Steps to Reproduce

1. Go to https://storybook.sanomalearning.design/?path=/story/actions-tool-bar--tooltips or https://storybook.sanomalearning.design/?path=/story/actions-tool-bar--icon-only
2. With screen reader turned on check if names of buttons with tooltips are announced
3. If names are announced check if name is provided with `aria-labelleby`. To do so use browser's Accessibility tab in DevTools.

Yes

https://storybook.sanomalearning.design/?path=/story/actions-tool-bar--tooltips 

_No response_

### ⚖️ Impact

### Framework

Angular

All

### If other, please specify.

_No response_

_No response_

### Component version

v0.2.3

### Theme version

_No response_

### 👤 Your name

Kacper

### 🧢 Your product/team

SLDS

### 🎨 Your Theme(s)

Sanoma Learning

### LINKED ISSUE #3287: [CHECKBOX] tooltip breaks sl-checkbox-group

Adding `tooltip` breaks `sl-checkbox`. The tooltip is not displayed and the checkbox doesn't work anymore. 

calling  `this.#form.reportValidity();` results in an error message:
```
checkbox-group.ts:191 Uncaught TypeError: box.reportValidity is not a function
    at checkbox-group.ts:191:36
    at Array.forEach (<anonymous>)
    at CheckboxGroup.reportValidity (checkbox-group.ts:191:17)
    at form.ts:180:37
    at Array.map (<anonymous>)
    at Form.reportValidity (form.ts:180:26)
    at FormController.reportValidity (form-controller.ts:119:24)
    at #onSave (composite-form.ts:127:16)
```

Code to reproduce this issue

```typescript
        <sl-form-field label="Subscriptions">
          <sl-checkbox-group name="subscriptions" required>
            <sl-checkbox ${tooltip('test')} value="newsletter">Newsletter</sl-checkbox>
            <sl-checkbox value="promotions">Promotions</sl-checkbox>
            <sl-checkbox value="updates">Product updates</sl-checkbox>
          </sl-checkbox-group>
        </sl-form-field>
```

### 👤 Your name

Frans de Boer

### 🧢 Your Product/Team

Magister

### LINKED ISSUE #3343: [Tooltip] multiple tooltips visible at the same time

Given a scenario where you have a nested DOM structure with the same web component that uses sl-tooltip, hovering the child component triggers the tooltip in the parent component as well.

Image

My guess if we're listening for an event and we're not checking if the event is for it for `this` or a child element.

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

### LINKED ISSUE #3344: [Button] add a `tooltip` property to button

A *lot* of usecases where you want to use `<sl-tooltip>` is for `<sl-button>`s. In those scenario's the tooltip has to be a sibling of the button so use can "connect" them using `aria-labelledby`/`aria-describedby`. That has to be taken into consideration for container components like `<sl-tool-bar>` etc making them more complex and leading to a lot of bugs.

My suggestion is to add tooltip support *directly* to button: `<sl-button tooltip="Hello world">...</sl-button>`. That way the `<sl-tooltip>` can be added to the shadow DOM of the button. This way no tooltip stuff will "leak" beyond `<sl-button>`.

- [ ] Remove pointer of tooltip in Figma component

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

jpzwarte:
Closes #3344 
Closes #3287 
Closes #3275
Closes #3255 
Closes #3343 
Closes #3274 
Closes #3049 
Closes #3264 

## Overview

This PR rewrites the `sl-tooltip` component and simplifies how tooltips are used across the design system. The new implementation replaces the complex internal machinery with native browser APIs and introduces a first-class `tooltip` property on several components.

---

## `@sl-design-system/tooltip` — major rewrite

The tooltip was the component responsible for the most bug reports and had a complex internal implementation. This rewrite significantly simplifies it.

The `AnchorController`, `EventsController`, and `Tooltip.lazy()` static method have all been removed in favour of native browser `popover` and CSS Anchor Positioning APIs.

> [!NOTE]
> CSS Anchor Positioning is not yet supported in all browsers. You may need to include the [CSS Anchor Positioning polyfill](https://anchor-positioning.oddbird.net/) in your application.

**Breaking changes**

- The `TooltipOptions` interface and `Tooltip.lazy()` static method have been removed. Use the `for` attribute to link a tooltip to its anchor instead.
- The `position`, `offset`, `maxWidth`, `arrowPadding`, and `viewportMargin` properties/statics have been removed.
- `hoverShowDelay` changed from `500ms` to `150ms` and `hoverHideDelay` changed from `200ms` to `0ms`.

**New API**

- `for` — links the tooltip to one or more anchor elements by id; pass a space-separated list of ids to share a single tooltip between multiple elements. Every listed element gets the ARIA relation and the triggers, and the tooltip is positioned against the anchor that triggered it
- `type` — controls the ARIA relationship: `'label'` (`ariaLabelledByElements`, default) or `'description'` (`ariaDescribedByElements`)
- `trigger` — space-separated list of triggers: `'focus'`, `'hover'`, and/or `'click'` (default: `'focus hover'`)
- `disabled` — prevents the tooltip from showing
- `open` — shows or hides the tooltip programmatically, regardless of trigger; to check whether the tooltip is showing, use `matches(':popover-open')` instead

---

## `tooltip` property on components — minor

Previously, adding a tooltip to a component required adding a sibling `<sl-tooltip>` element manually and wiring up the correct `aria-describedby` or `aria-labelledby` relationship by hand. The new `tooltip` property improves the DX by letting you set a tooltip directly on the component, with all accessibility wiring handled automatically:

- For **icon-only buttons** the tooltip text acts as the accessible label (`aria-labelledby`).
- For **text buttons** the tooltip text acts as an accessible description (`aria-describedby`).

Affected components: `@sl-design-system/button`, `@sl-design-system/menu`, `@sl-design-system/tag`, `@sl-design-system/toggle-button`.

---

## `@sl-design-system/toggle-button` — major refactor

Refactored to use an internal `<button>` element. This improves accessibility and removes the need for manual keyboard and ARIA handling.

**Breaking changes**

- The `[pressed]`, `[icon-only]`, `[text-only]`, and `[error]` attributes have been replaced by CSS custom states (`:state(pressed)`, `:state(icon-only)`, `:state(text-only)`, `:state(error)`).
- The `shape` property type changed from `ButtonShape` to `ToggleButtonShape` (`'rect' | 'pill'`).
- The `pressed` property is no longer reflected as an attribute.
- The `label` property has been removed. Use the new `tooltip` property instead.

**New**

- CSS parts `button` and `tooltip` for styling the internal elements.
- Focus delegated to the internal `<button>` via `delegatesFocus: true`.
- ARIA attributes forwarded to the internal `<button>` via `ForwardAriaMixin`.

---

## `@sl-design-system/toggle-group` — minor

The group styles its slotted buttons, so it follows the toggle button refactor above: it now targets `::slotted(sl-toggle-button:state(pressed))` instead of the `[pressed]` attribute, and applies the border radius of the first and last button through the `--_button-border-radius` custom property rather than separate corner properties.

> [!IMPORTANT]
> `@sl-design-system/toggle-group` and `@sl-design-system/toggle-button` must be upgraded together. These styles do not match an older toggle button, and the group only manages buttons that are instances of the `ToggleButton` class it was built against, so mixing versions leaves the pressed state unstyled and breaks keyboard navigation and single-select behaviour.

---

## `@sl-design-system/checkbox` — minor

`sl-checkbox` now uses `ForwardAriaMixin` instead of `ObserveAttributesMixin` to forward ARIA to its internal `<input>`:

- Reference attributes are resolved to elements and forwarded via `ariaLabelledByElements`/`ariaDescribedByElements`, so relationships set as element references — such as an `sl-tooltip` labelling or describing the checkbox — now reach the `<input>`. Previously only the attribute value was copied, which meant a tooltip anchored to an `sl-checkbox` did not label the actual control.
- All `aria-*` attributes set on the host are forwarded, rather than a fixed list. `aria-describedby` therefore reaches the `<input>` as well; it used to stay on the host.

---

## `@sl-design-system/shared` — minor

Improved how `ForwardAriaMixin` forwards ARIA element references (such as `ariaDescribedByElements` and `ariaLabelledByElements`) to its proxy target:

- References added by other code are no longer overwritten. Components can maintain their own ARIA relationships (such as a button registering its own tooltip) without interference.
- The mixin keeps track of the references it forwarded itself, so re-forwarding replaces them instead of appending. Previously, changing an attribute like `aria-labelledby` left the old reference in place, resulting in a stale accessible name.
- Removing an `aria-*` attribute from the host is now forwarded to the target as well: plain attributes are removed and element references are cleared, leaving any references added by others intact.

---

## Other components updated to use the new tooltip implementation — minor

- `@sl-design-system/avatar`
- `@sl-design-system/breadcrumbs` — tooltips for truncated links now use the `for` attribute approach, removing the need for manual cleanup functions.
- `@sl-design-system/calendar`
- `@sl-design-system/ellipsize-text`
- `@sl-design-system/grid`
- `@sl-design-system/angular` — tooltip directive updated; API unchanged but DOM structure may have changed. The directive also no longer creates an `<sl-tooltip>` element when the tooltip text is empty, which previously gave the anchor element an empty accessible name or description; the tooltip is created as soon as the text becomes non-empty and removed again when it becomes empty.

---

## Bug fixes

- **`@sl-design-system/menu`** — Fixed: clicking the menu button a second time did not dismiss the menu. The button's click handler was firing after the popover's `toggle` event and calling `togglePopover()` on an already-closed menu. Fixed by tracking a `#popoverJustClosed` flag via the `beforetoggle` event. The menu button's focus management also changed: the menu is focused when opened via click, focus returns to the button after selecting a menu item (preserving `:focus-visible`), and `pointerdown` is suppressed while the menu is open.
- **`@sl-design-system/checkbox`** — Fixed: `sl-checkbox-group` treated any slotted element as a checkbox, so a slotted `<sl-tooltip>` broke the group (#3287). The group now only queries `sl-checkbox` elements. Also fixed the display of checkboxes with the `no-label` attribute.
- **`@sl-design-system/button-bar`** — Fixed: buttons were not given enough time to set their icon-only state.

---

## Tooling

- **`eslint-plugin-slds`** — the `button-has-label` rule now accepts the `tooltip` attribute on `sl-button` as providing an accessible name.


a11ymiko (commented):
`esc` key cannot be used to dismiss tooltip triggered by mouse for Button stories (Icon Only and Double Label). When tested main region including button was focused so `esc` should work as it works on our production right now.
`esc` works well and dismisses tooltip when it's triggered by keyboard focus.

https://github.com/user-attachments/assets/2af93520-5c20-42b7-b73d-daceefc7a1f2

packages/components/tooltip/src/tooltip.stories.ts
```diff
@@ -1,330 +1,106 @@
...
-              >${message}</sl-tooltip
-            >
-          `}
-    `;
-  }
+  args: {
+    text: 'Tooltip text'
+  },
+  render: ({ disabled, maxWidth, open, position, showHoverBridge, text, tooltip, trigger }) => html`
+    <sl-button id="button">Anchor</sl-button>
```
a11ymiko, line 71: Buttons with text in stories from group Tooltip2 use `aria-labelledby` to attach tooltip content to them. Because of that screen readers announce only tooltip's content and not inform user about visible text of the button which is a bug.

For buttons that have visible text tooltip's should be referenced using `aria-describedby` rather than `aria-labelledby` and then screen reader would announce visible text as button's accessible name and toolpi's text as additional accessible description.

Image

a11ymiko (commented):
In Edge Case Menu Button tooltip is still visible after menu is expanded and tooltip makes it difficult to use menu items.

with mouse: 

https://github.com/user-attachments/assets/5cf00303-7115-4c2a-8516-f620f987415f

with keyboard:

https://github.com/user-attachments/assets/5f0faa88-1910-461a-855c-5ce1caa665c8


Menu button's tooltip should disappear when menu is expanded either by mouse or by keyboard.

When menu is expanded tooltip on menu button may be triggered by mouse but we have to make sure tooltip is not displayed over any menu item.

a11ymiko (commented):
Tooltip2 have issues with being hoverable. Mouse user mostly cannot hover over tooltip's content without closing it which is WCAG violation:

https://github.com/user-attachments/assets/07854623-ae79-4de8-81e5-9e55bc3d2667

User should have possibility to move mouse cursor frokm Anchor to tooltip's content without closing it as it's shown in other storybook's examples: 

https://github.com/user-attachments/assets/d4ea5381-e43e-4c22-adf2-795bf1ec2ea7

a11ymiko (commented):
Tooltip2 is expanding its width with each time it's being triggered either by mouse or keyboard focus: 

https://github.com/user-attachments/assets/91168bba-3cd8-4690-9f6b-aef075439263

https://github.com/user-attachments/assets/50143007-27a9-400a-b41f-8ed9b42e1927

a11ymiko (commented):
`Disabled` tooltip2 is not working properly.

Right now mouse users and keyboard users cannot make disabled tooltip visible, but screen readers users placing their screen reader's focus on Anchor get full content of tooltip as Anchor's accessibility name.

https://github.com/user-attachments/assets/99fc24d5-6cd7-499b-a965-3fe0c4b3ada1

If I understand correctly only tooltip should be disabled in this scenario (so it won't appear on hover/on focus) but the anchor is still active. 
If that's true then anchor in that scenario should not reference tooltip via `aria-labelledby` or `aria-describedby`.

Screen readers users when placing their SC focus on Anchor should be announced only 'Anchor, button' without any tooltip's content.

packages/components/tooltip/src/tooltip.stories.ts
jpzwarte: Fixed

a11ymiko (commented):
For stories in Tooltip2 group (Basic, Hover Bridge, Positions) tooltip is not showed on focus.

https://github.com/user-attachments/assets/96df44ff-8bd2-42d2-951c-1f7a555d4b20

a11ymiko (approved):

michal-sanoma (approved):
Amazing work with this refactor! 🚀 

Small compatibility clarification: do we still want to support shared / multi-anchor tooltips?

The old implementation allowed multiple anchors to point to the same `<sl-tooltip>` via ARIA attributes, and we had coverage for that in the `Shared` story and `tooltip-shared.spec.ts`. The new `for` API appears to model a single tooltip-to-anchor relationship instead

If that change is intentional, I think it would be helpful for consumers if we mention it as a breaking change and suggest the preferred replacement pattern, wdyt?

.changeset/clean-chicken-look.md
```diff
@@ -0,0 +1,19 @@
+---
+'@sl-design-system/toggle-button': minor
+---
+
+Refactor toggle button to use an internal `<button>` element. This improves accessibility and removes the need for manual keyboard and ARIA handling.
+
+**Breaking changes**
```
Diaan, line 7: with this many breaking changes, is it still a `minor` version update?

packages/components/menu/src/menu-button.spec.ts
```diff
@@ -145,15 +147,15 @@ describe('sl-menu-button', () => {
           expect(menu).not.to.match(':popover-open');
         });
 
-        it('should not move focus when the button is clicked without being focused', async () => {
+        it('should focus the first menu item when the button is clicked', async () => {
```
Diaan, line 162: I'm not sure this is correct. When i click the button it doesn't show a focusring on the first menu item:
https://storybook-3368.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/actions-menu-menu-button--submenu

packages/components/menu/src/menu-button.stories.ts
Diaan: these need to be updated as well

packages/components/toggle-button/src/toggle-button.spec.ts
```diff
@@ -370,122 +379,51 @@ describe('sl-toggle-button', () => {
...
-      await userEvent.hover(el);
-
-      const tooltip = el.nextElementSibling;
-
-      expect(tooltip).to.exist;
-      expect(el).to.have.attribute('aria-label', 'Settings');
-      expect(el).to.have.attribute('aria-describedby', tooltip?.id);
-    });
-
-    it('should be icon-only even if one icon is missing (Errors variant)', async () => {
```
Diaan, line 433: why is this removed?

packages/components/toggle-button/src/toggle-button.ts
```diff
@@ -156,93 +153,75 @@ export class ToggleButton extends ScopedElementsMixin(LitElement) {
...
-      if (this.#tooltip instanceof Tooltip) {
-        this.#tooltip.remove();
-        this.#tooltip = undefined;
-      } else if (this.#tooltip) {
-        this.#tooltip();
-        this.#tooltip = undefined;
-      }
-    }
-
-    if (changes.has('disabled')) {
```
Diaan, line 205: why is this removed?

packages/components/tool-bar/src/tool-bar.stories.ts
```diff
@@ -440,179 +437,22 @@ export const Tooltips: Story = {
   args: {
     description: 'This example shows a tool bar with different tooltip techniques on the buttons.',
     items: () => html`
-      <sl-button aria-labelledby="tooltip-bold" fill="outline">
+      <sl-button aria-labelledby="tooltip-bold" fill="outline" id="bold">
```
Diaan: Does it still need the `aria-labelledby="tooltip-bold"`?

Diaan (approved):

packages/components/toggle-button/src/toggle-button.ts
jpzwarte: Because toggle-button now has a `<button>` inside it and also uses `ForwardAriaMixin`, so this code is no longer necessary.

packages/components/toggle-button/src/toggle-button.spec.ts
jpzwarte: Because toggle-button now has a <button> inside it and also uses ForwardAriaMixin, so this code is no longer necessary.

packages/components/menu/src/menu-button.spec.ts
jpzwarte: Works for me: when i press Shift after clicking the menu button, it shows the focus outline on the first menu item.

packages/components/menu/src/menu-button.stories.ts
jpzwarte: Done

packages/components/tool-bar/src/tool-bar.stories.ts
jpzwarte: Fixed

.changeset/clean-chicken-look.md
jpzwarte: Changed to major!

a11ymiko (commented):
Examples of tooltips on documentation website do not work right now: 

https://github.com/user-attachments/assets/2aecab60-09c9-4c25-b261-694e45da911b

a11ymiko (commented):
As this new implementation doesn't support 'shared' tooltip for multiple elements it also shouldn't be presented as example on the documentation website, right? 

Image

a11ymiko (commented):
Clicking on tooltip for menu button expands menu.

https://github.com/user-attachments/assets/ba913ae0-5725-40d3-8650-971ecda32d1c

Clicking on tooltip for menu button should do nothing as it works for Toggle button or Toggle group:

https://github.com/user-attachments/assets/7a6a88a2-e460-4a4d-8d23-e59fc24a521b

a11ymiko (commented):
Tooltips don't work in story [Popover With Tooltips](https://storybook-3368.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/overlay-popover--with-tooltips):

https://github.com/user-attachments/assets/c7802c30-01eb-4b11-bed9-744fc79285f5

a11ymiko (commented):
Tooltips don't work in story [Panel Actions With Tooltips](https://storybook-3368.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/layout-panel--actions-with-tooltips):

https://github.com/user-attachments/assets/263ed774-d4ea-40cc-88ff-363cd49cc05a

jpzwarte:
> Clicking on tooltip for menu button expands menu.

Fixed!

a11ymiko (commented):
Tooltip for [checkbox with no visible label](https://storybook-3368.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/form-checkbox-checkbox--no-visible-label) doesn't serve as accessible name of an input. Tooltip is added as `aria-labelledby` for `sl-checkbox` but NVDA is not reading that label, as it is only reading labels added for `<input>` inside `sl-checkbox.`

If tooltip info is expecte do act as a label in this case tooltip's ID should be added as `aria-labelledby` to `<input>`.

https://github.com/user-attachments/assets/9945a16a-b40c-4329-bc0f-3ee6d79af764

Image

jpzwarte:
> Tooltips don't work in story [Panel Actions With Tooltips](https://storybook-3368.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/layout-panel--actions-with-tooltips):

Fixed; i actually removed this particular story as all buttons in the panel stories now have tooltips (most previously only had `aria-label`s).

jpzwarte:
> Tooltips don't work in story [Popover With Tooltips](https://storybook-3368.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/overlay-popover--with-tooltips):

Fixed!

jpzwarte:
> Tooltip for [checkbox with no visible label](https://storybook-3368.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/form-checkbox-checkbox--no-visible-label) doesn't serve as accessible name of an input. Tooltip is added as `aria-labelledby` for `sl-checkbox` but NVDA is not reading that label, as it is only reading labels added for `<input>` inside `sl-checkbox.`

Checkbox still uses the old `ObserveAttributesMixin` and not the new `ForwardAriaMixin`; i would rather refactor this in a different PR?

jpzwarte:
> Checkbox still uses the old `ObserveAttributesMixin` and not the new `ForwardAriaMixin`; i would rather refactor this in a different PR?

Fixed here.

a11ymiko (commented):
For button on documentation website screen readers don't read its visible text, only tooltip text.

https://github.com/user-attachments/assets/ae53b822-8e44-41b0-aa45-280560e50619

https://github.com/user-attachments/assets/0fdd1ef1-1407-4c6d-873e-e811ce9e08ab

jpzwarte:
> For button on documentation website screen readers don't read its visible text, only tooltip text.

Fixed; `type="description"` was missing.

a11ymiko (approved):

.changeset/violet-baboons-crash.md
```diff
@@ -0,0 +1,32 @@
+---
+'@sl-design-system/tooltip': major
+---
+
+Rewrite of the tooltip component: the tooltip was the component responsible for the most bug reports and had a complex internal implementation. This rewrite significantly simplifies the component.
+
+The complex internal positioning logic, `AnchorController`, `EventsController`, and `lazy()` static method have been removed in favour of native browser `popover` and CSS Anchor Positioning APIs.
+
+> [!NOTE]
+> CSS Anchor Positioning is not yet supported in all browsers. You may need to include the [CSS Anchor Positioning polyfill](https://anchor-positioning.oddbird.net/) in your application.
```
anna-lach, line 10: Maybe it's worth to mention it here: https://website-3368.d3c5h2uor7jlkj.amplifyapp.com/categories/getting-started/developers/#add-polyfills as well?

.changeset/violet-baboons-crash.md
```diff
@@ -0,0 +1,32 @@
...
+
+#### Breaking changes
+
+- The `TooltipOptions` interface and `Tooltip.lazy()` static method have been removed. Use the `for` attribute to link a tooltip to its anchor instead.
```
anna-lach, line 14: So probably this part of the documentation should be removed?
https://website-3368.d3c5h2uor7jlkj.amplifyapp.com/categories/components/tooltip/code/#tooltip.lazy-helper

packages/components/tooltip/src/tooltip.ts
```diff
@@ -1,1387 +1,457 @@
...
+ * separated by spaces, so a single tooltip can serve several elements.
+ *
+ * The tooltip will automatically determine the appropriate ARIA relation to use based on the `type`
+ * property. By default, it will use `ariaLabelledByElements`, but if `type` is set to
+ * `description`, it will use `ariaDescribedByElements` instead.
  *
- * @slot default - The slot for the tooltip content.
+ * @element sl-tooltip
+ *
+ * @slot - The content of the tooltip.
```
anna-lach, line 25:
```suggestion
 * @slot default - The content of the tooltip.
```

Right now there is an empty name in the API documentation:

Image

packages/components/avatar/src/avatar.ts
```diff
@@ -46,15 +46,16 @@ export type AvatarSize = 'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl' | '4xl';
  *   picture-url="http://sanomalearning.design/avatars/lynn.png"></sl-avatar>
  * ```
  *
+ * @slot - The subheading of the avatar.
```
anna-lach, line 49:
```suggestion
 * @slot default - The subheading of the avatar.
```

Right now name in the API documentation is empty

Image

packages/components/avatar/src/avatar.ts
```diff
@@ -175,8 +181,10 @@ export class Avatar extends ScopedElementsMixin(LitElement) {
       ${this.imageOnly
         ? nothing
         : html`
-            <sl-tooltip id="avatar-tooltip">${this.displayName}</sl-tooltip>
-            <span part="name">${this.displayName}</span>
+            ${this.tooltip
+              ? html`<sl-tooltip for="name" part="tooltip">${this.displayName}</sl-tooltip>`
```
anna-lach, line 185: I know it works the same on `main`, but is it ok that we only have display name in the tooltip and no subheading? Subheading also has ellipsis in that case and right now there is no way to check the whole text there?

Image

anna-lach, line 185: I know it works the same on `main`, but is it ok that we only have display name in the tooltip and no subheading? Subheading also has ellipsis in that case and right now there is no way to check the whole text there?

Image

packages/components/ellipsize-text/src/ellipsize-text.ts
```diff
@@ -15,6 +16,8 @@ declare global {
 /**
  * Small utility component to add ellipsis to text that overflows its container. It also adds a
  * tooltip with the full text.
+ *
+ * @slot - The text to be truncated.
```
anna-lach, line 20:
```suggestion
 * @slot default - The text to be truncated.
```

packages/components/toggle-button/src/toggle-button.scss
```diff
@@ -166,21 +87,118 @@
...
+  &:is(:disabled, [aria-disabled]) {
+    --_bg-color: transparent;
+    --_bg-mix-color: transparent;
+
+    border-color: var(--sl-color-border-disabled);
+    color: var(--sl-color-foreground-disabled);
+    cursor: default;
+
+    &[aria-pressed='true'] {
+      // --_bg-color: var(--sl-color-background-disabled);
```
anna-lach: Do we still need this commented line?

packages/components/tooltip/src/edge-cases.stories.ts
```diff
@@ -0,0 +1,85 @@
...
+import '../register.js';
+
+Icon.register(faGear, faPen, faTrash, fasGear);
+
+export default {
+  title: 'Overlay/Tooltip/Edge cases'
+};
+
+export const Dialog = {
+  render: () => html`
```
anna-lach, line 22: In this case the tooltip should be shown only after closing the dialog using keyboard (also this is how it works on `main`) and now it's also shown when it's just closed using mouse. So it should be shown when a button that triggers dialog is focused again after closing the dialog (with keyboard).

Could you please improve the dialog a little bit and make it similar to that one on `main`? :) 

Image

Image

packages/components/tooltip/src/tooltip.stories.ts
```diff
@@ -1,324 +1,140 @@
...
-            ariaRelation: 'label',
-            position: 'bottom-start',
-            maxWidth: 100
-          })}>
-          <sl-icon name="face-smile" size="lg"></sl-icon>
-        </sl-button>
-      </div>
-    `;
+    text: 'Click again to dismiss',
+    trigger: ['click']
```
anna-lach, line 97: So `click` trigger is more for touchscreen users? Do I understand it right? Probably it should be described somewhere in the documentation?

packages/components/tooltip/src/tooltip.ts
```diff
@@ -1,1387 +1,457 @@
...
+    const a = anchor.getBoundingClientRect(),
+      t = this.getBoundingClientRect();
+
+    // Determine on which side of the anchor the tooltip ended up (after CSS anchor positioning
+    // and any position-try fallbacks). We then build a trapezoid whose parallel edges align with
+    // the touching edges of the anchor and the tooltip, so the user can move the pointer between
+    // the two without crossing an unhovered area.
+    let left: number, top: number, width: number, height: number, polygon: string;
+
+    if (t.bottom <= a.top) {
```
anna-lach, line 344: Do we have an issue to update `position` in Figma component? It looks like in Figma we still have arrows and more positions available like `right-bottom` and so on.

https://www.figma.com/design/CHpKrPIdXdbV2u7X8vizKI/Components-2.0?node-id=8660-65181&m=dev

packages/components/avatar/src/avatar.ts
anna-lach: Duplicate :)

packages/components/tooltip/src/tooltip.stories.ts
jpzwarte: This is more for different use cases. Like an `<sl-copy-button>` where you want to show a tooltip with "Copied!" after you click it (like we have in the new documentation website ;) )

packages/components/tooltip/src/tooltip.ts
jpzwarte: https://github.com/sl-design-system/components/issues/3574

packages/components/tooltip/src/edge-cases.stories.ts
jpzwarte: Fixed

packages/components/ellipsize-text/src/ellipsize-text.ts
jpzwarte: CleanShot 2026-08-05 at 11 14 06

By not specifying an explicit slot, CEM will automatically mark is as `(default)`.

packages/components/avatar/src/avatar.ts
jpzwarte: This is an edge case. One that i've seen happen with student names. But the subheading is of less importance here. Having 2 tooltips would be overkill imho.

jpzwarte: If you disagree, let's create a new issue for this :)

packages/components/avatar/src/avatar.ts
jpzwarte: Ah, that's a shortcoming of the current documentation website. I'll see if i can fix that. See comment below how it looks in the new website.

packages/components/avatar/src/avatar.ts
anna-lach: Yeah, but maybe it could be shown in one tooltip? Maybe it's worth to ask @arecuenco-dsgn or @a11ymiko ? :) 
Probably sth we can look into in a separate issue. :)

packages/angular/src/tooltip.directive.ts
```diff
@@ -6,40 +6,59 @@ import {
...
+    if (!text) {
       this.tooltip?.remove();
       this.tooltip = undefined;
     } else if (this.tooltip) {
-      this.tooltip();
-      this.tooltip = undefined;
+      this.tooltip.textContent = text;
+    } else {
+      this.tooltip = document.createElement('sl-tooltip');
+      this.tooltip.for = this.elRef.nativeElement.id ||= `sl-tooltip-${Math.random()
```
michal-sanoma, line 57: **question (non-blocking):** can this fallback ID, generated with `Math.random()`, cause problems with Angular SSR/hydration?

tools/eslint-plugin-slds/lib/rules/button-has-label.js
michal-sanoma: **question:** shouldn't we remove this one as `tooltip-directive.ts` is gone?

Diaan (approved):

website/src/categories/components/tooltip/usage.md
michal-sanoma: **question:** isn't pointer removed from the SCSS/template in this refactor?

.changeset/violet-baboons-crash.md
jpzwarte: Done

packages/components/avatar/src/avatar.ts
jpzwarte: #3578

packages/components/shared/src/mixins/forward-aria-mixin.ts
```diff
@@ -164,6 +176,46 @@ export function ForwardAriaMixin<
...
+    override removeAttribute(name: string): void {
+      if (observedAttributes ? observedAttributes.includes(name) : name.startsWith('aria-')) {
+        // Always remove from pending so a queued forward doesn't re-add it.
+        this.#pendingAttributes.delete(name);
+
+        // If the attribute is still present on the host, this call came from
+        // #forwardAttributes cleaning up after forwarding — the proxy was just
+        // set correctly, so don't touch it. Only clear the proxy when the
+        // attribute is already absent (i.e. an explicit external removal call).
+        if (!this.hasAttribute(name)) {
```
michal-sanoma: **suggestion:** can we add a regression test for this path? 

The case I’m worried about is: existing ARIA value has already been forwarded to the proxy target, then the host gets `setAttribute(name, newValue)` followed immediately by `removeAttribute(name)` before the MutationObserver runs. With the current `hasAttribute()` check, the proxy target can retain the stale value

.changeset/violet-baboons-crash.md
jpzwarte: Done

packages/components/tooltip/src/tooltip.stories.ts
jpzwarte: Docs added

packages/angular/src/tooltip.directive.ts
jpzwarte: Angular SSR/hydration ignores web components completely. So there are bigger issues to solve there before this becomes a thing :)

website/src/categories/components/tooltip/usage.md
jpzwarte: Removed

packages/components/shared/src/mixins/forward-aria-mixin.ts
jpzwarte: Done

tools/eslint-plugin-slds/lib/rules/button-has-label.js
jpzwarte: Done

a11ymiko (approved):

anna-lach (approved):

PR #3379, opened 2026-05-22
fix(grid): Fix empty Grid menu button column header

### LINKED ISSUE #2881: [Grid - Basic] Menu button variant has empty table header

Menu button variants have a menu button with actions in the last column, but the column is missing a descriptive header. This makes the function, and the structure of the table hard to interpret without exploration.

Table header cell, which is the last element in the table header row is empty.

### 🤔 Expected Behavior?

Table header has a text, which describes the purpose of the column.

### 😯 Current Behavior

Table header is empty and is not communicated to assistive technologies.

### 💁 Possible Solution

Add a text (e.g., only visible to screen reader users) which describes the purpose of the column. E.g., add a sr-only text "Actions" to explain the purpose of the column.

### 🖥️ Steps to Reproduce

Navigate to the header cell for the column with the functions -> Screen reader either reads "empty" or "space" (along with mentioning the table rows/columns and announcing that the cell is a table header).

Chrome

### If other, please specify.

_No response_

Windows 11

### 👤 Your name

Valtteri Järvinen

### 🧢 Your product/team

External

### 🎨 Your Theme(s)

Sanoma Learning

michal-sanoma:
## Summary

This PR fixes the Grid/Basics/MenuButton story where the final menu button column rendered an empty table header

The menu button column now provides a visually hidden Actions header so assistive technologies can announce the purpose of the column while keeping the visual layout unchanged

Also updates the Grid header styling so .visually-hidden content can  be used in any table header, not only the drag handle column

### BEFORE

Screenshot 2026-05-22 at 18 02 46

### AFTER

Screenshot 2026-05-22 at 18 03 55


a11ymiko (approved):

packages/components/grid/src/stories/basics.stories.ts
```diff
@@ -199,21 +193,19 @@ export const MenuButton: Story = {
...
           header="Person"
           .renderer=${avatarRenderer}
-          .scopedElements=${{ 'sl-avatar': Avatar }}
-        ></sl-grid-column>
+          .scopedElements=${{ 'sl-avatar': Avatar }}></sl-grid-column>
         <sl-grid-column path="email"></sl-grid-column>
         <sl-grid-column
           grow="0"
-          header=""
+          .header=${() => html`<span class="visually-hidden">Actions</span>`}
```
jpzwarte: So the solution here works, but is it intuitive enough for users? Will they remember to do this?

What if we add a `hide-header-text` boolean property instead?

```html
header="Actions"
hide-header-text
```

jpzwarte (approved):

Diaan (approved):

packages/components/grid/src/stories/basics.stories.ts
michal-sanoma: good idea, I've simplified to yr suggestion

a11ymiko (approved):

PR #3381, opened 2026-05-25
fix(date-field): enhance focus management and dialog behavior in date picker

### LINKED ISSUE #3074: [Date field] Focus can escape expanded date-field picker without collapsing the picker.
## Current behavior

User can expand [Date field](https://storybook.sanomalearning.design/?path=/story/form-date-field--basic) picker using "Select date" button. Their focus is automatically placed on some day date inside the dialog. If user uses Tab in this situation their focus goes out of dialog to the next widget on the app. In "Select day" view of calendar controls such as month and year selectors and buttons for previous and next month are skipped with tab and user may be unaware of existence of that controls.
This issue occurs only in "Select day" variant of the calendar, because in "Select month" and "Select year" using Tab inside opened dialog creates focus trap that moves it in the loop, which is desired behavior.

https://github.com/user-attachments/assets/a27e98bd-3b02-4401-a55f-5377ef18fbba

https://github.com/user-attachments/assets/f3ce0b86-7b12-4e25-b304-a07dafbdb5db

If user uses Shift+Tab while their focus is placed on date then focus is moved properly to calendar controls (month and year selectors and buttons for previous and next month), but user can go out of dialog with that shortcut aiming "Select date" button, and the rest of the page.
This issue occurs in "Select dat", "Select month" and "Select year" variants of the calendar.

https://github.com/user-attachments/assets/c000b2ae-9fef-4f5b-ab61-96c8f15fc96c

https://github.com/user-attachments/assets/f83e4f78-82e8-4219-8101-cc3691f131bb

## Proposed solution

It's recommended to use focus trap for both focus moving forward with Tab and backward with Shift+Tab on each expanded dialog. Focus trap can for each calendar view should limit elements possible to be reached with Tab only to visible calendar controls. Example: https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/#example

Another possible solution is to use `dialog` element that adds `inert` attribute to all page elements that are underneath the dialog. In that scenerio there is no need to add focus trap inside dialog, because after last element inside dialog user would place their focus on URL adres bar (or other browser control) which is known information that there is no other tabbable items on the page. More about why this works: https://css-tricks.com/there-is-no-need-to-trap-focus-on-a-dialog-element/

### 👤 Your name

Kacper

### 🧢 Your Product/Team

SLDS

anna-lach:
This pull request refactors the `sl-date-field` component to replace the use of a popover with a native dialog for the calendar picker, improving accessibility and focus management. 

**Component refactor: popover → dialog**

- Replaced all references to "popover" with "dialog" in code, comments, and tests, and removed popover-specific code and properties such as `#popoverJustClosed`, `offset`, and `viewportMargin`. The dialog is now managed using native dialog APIs and events, with improved accessibility attributes and event handling. [[1]](diffhunk://#diff-8cffd37537884231ae8731da24c3c42d23d63685216c4317b7cfb560aa33cb9cL67-L69) [[2]](diffhunk://#diff-8cffd37537884231ae8731da24c3c42d23d63685216c4317b7cfb560aa33cb9cL90-L92) [[3]](diffhunk://#diff-8cffd37537884231ae8731da24c3c42d23d63685216c4317b7cfb560aa33cb9cL108-R106) [[4]](diffhunk://#diff-8cffd37537884231ae8731da24c3c42d23d63685216c4317b7cfb560aa33cb9cL132-R116) [[5]](diffhunk://#diff-8cffd37537884231ae8731da24c3c42d23d63685216c4317b7cfb560aa33cb9cL149-R133) [[6]](diffhunk://#diff-8cffd37537884231ae8731da24c3c42d23d63685216c4317b7cfb560aa33cb9cL225-R210) [[7]](diffhunk://#diff-8cffd37537884231ae8731da24c3c42d23d63685216c4317b7cfb560aa33cb9cL242-R226) [[8]](diffhunk://#diff-8cffd37537884231ae8731da24c3c42d23d63685216c4317b7cfb560aa33cb9cL398-R394) [[9]](diffhunk://#diff-8cffd37537884231ae8731da24c3c42d23d63685216c4317b7cfb560aa33cb9cL426-R406) [[10]](diffhunk://#diff-283168ba51015418833217bcabcdc7c7192aaeda8b65f5956a677f6cc5be2cf2L141-R148)

- Updated the dialog rendering in the template to remove the `popover` attribute and related anchor positioning, and to use dialog-specific event handlers (`@close`, `@click`, `@cancel`). [[1]](diffhunk://#diff-8cffd37537884231ae8731da24c3c42d23d63685216c4317b7cfb560aa33cb9cL398-R394) [[2]](diffhunk://#diff-8cffd37537884231ae8731da24c3c42d23d63685216c4317b7cfb560aa33cb9cL426-R406)

**Accessibility and focus management improvements**

- Updated the field button's `aria-expanded` attribute and related logic to reflect the dialog's open state, and improved dialog accessibility attributes. [[1]](diffhunk://#diff-8cffd37537884231ae8731da24c3c42d23d63685216c4317b7cfb560aa33cb9cL398-R394) [[2]](diffhunk://#diff-d7fd6a020def5b7eff6855d90a5ca79f3335b0806d7b85392a2ec984932f2bf4L714-R823)


a11ymiko (commented):
Right now in Select Day keyboard focus can escape dialog and go to browser controls like URL adress bar. That is wanted behavior that is also similar to behaviors of other dialogs in SLDS.
But in Select Month and Select Tear view focus moves in endless loop between only dialog components and cannot leave dialog to be placed on browser's controls. That is not a violation but this focus order is not similar to Select Day view's and other dialogs in SLDS and that may be not easy to understand by users.
It is recommended to standardize the behavior of focus in each view so in each view user can leave dialog and place focus on browser's controls as it is now true for the Select Day view

michal-sanoma (approved):
Great work! 

One non blocking comment. On FF when we open the dialog focus is not visible at the beggining

FF:

https://github.com/user-attachments/assets/2f018d4f-ec2f-4a15-8fd0-d9c4561c82e9


Chrome:

https://github.com/user-attachments/assets/270c83eb-1298-4ffb-b1d2-f94821d26f88

Diaan (approved):

PR #3385, opened 2026-05-26
fix(group-header): add aria-expanded attribute to toggle group button for accessibility

### LINKED ISSUE #2883: [Grid - Grouping] Accordions missing expanded/collapsed state for assistive technologies

The accordion buttons in the grouped variant are missing aria-expanded attribute. Assistive technology users (especially screen reader users) must explore the surrounding elements to understand whether the element is closed or opened. 

Image

### 🤔 Expected Behavior?

When user operates the accordion button users would head "expanded" or "collapsed" (or similar depending on screen reader)

### 😯 Current Behavior

When user operates the accordion button screen readers stay silent (some might repeat the current button)

### 💁 Possible Solution

Add aria-expanded with value "true" when the accordion is open and "false" when it is closed. 

### 🖥️ Steps to Reproduce

1. Navigate to accordion button with a screen reader
2. Activate it (Enter or Space key) -> Screen reader stays silent

Chrome

### If other, please specify.

_No response_

Windows 11

### 👤 Your name

Valtteri Järvinen

### 🧢 Your product/team

External

### 🎨 Your Theme(s)

Sanoma Learning

anna-lach:
This pull request introduces an accessibility improvement to the grid component by adding the `aria-expanded` attribute to the toggle group button in the group header.

**Accessibility improvements:**
* Added the `aria-expanded` attribute to the toggle group button in the group header of the grid component to indicate its expanded/collapsed state.


a11ymiko (commented):
Right now in Select Day keyboard focus can escape dialog and go to browser controls like URL adress bar. That is wanted behavior that is also similar to behaviors of other dialogs in SLDS.
But in Select Month and Select Tear view focus moves in endless loop between only dialog components and cannot leave dialog to be placed on browser's controls. That is not a violation but this focus order is not similar to Select Day view's and other dialogs in SLDS and that may be not easy to understand by users.
It is recommended to standardize the behavior of focus in each view so in each view user can leave dialog and place focus on browser's controls as it is now true for the Select Day view

a11ymiko (approved):

a11ymiko (approved):

michal-sanoma (approved):

packages/components/grid/src/group-header.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    customElements.define('sl-grid-group-header', GridGroupHeader);
+  }
+} catch {
+  // empty
+}
+
+describe('sl-grid-group-header', () => {
+  let el: GridGroupHeader;
+
+  describe('aria-expanded attribute', () => {
```
Diaan: I don't think this should be a 'describe' level. There are scenarios like "default" where expanded is true by default, and "collapsed" where it's false. On those levels you probably want more tests (not in this story, but in general) and ONE of those tests is whether the `aria-expanded` attribute is set correctly.

packages/components/grid/src/group-header.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      it('should have aria-expanded="true" on the toggle button', () => {
+        const button = el.renderRoot.querySelector<Button>('sl-button');
+
+        expect(button).to.exist;
+        expect(getForwardedAriaAttribute(button!, 'aria-expanded')).to.equal('true');
+      });
+    });
+
+    describe('when collapsed is true', () => {
```
Diaan:
```suggestion
    describe('collapsed', () => {
```

packages/components/grid/src/group-header.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      it('should have aria-expanded="false" on the toggle button', () => {
+        const button = el.renderRoot.querySelector<Button>('sl-button');
+
+        expect(button).to.exist;
+        expect(getForwardedAriaAttribute(button!, 'aria-expanded')).to.equal('false');
+      });
+    });
+
+    describe('when clicking the toggle button', () => {
```
Diaan:
```suggestion
    describe('toggle button', () => {
```

packages/components/grid/src/group-header.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+        expect(getForwardedAriaAttribute(button!, 'aria-expanded')).to.equal('true');
+
+        button!.click();
+        await el.updateComplete;
+        await button!.updateComplete;
+
+        expect(getForwardedAriaAttribute(button!, 'aria-expanded')).to.equal('false');
+      });
+
+      it('should toggle aria-expanded back from "false" to "true" on second click', async () => {
```
Diaan:
```suggestion
      it('should toggle aria-expanded back from "false" to "true" correctly when expanding', async () => {
```

packages/components/grid/src/group-header.spec.ts
anna-lach: I've changed it, thanks. :)

Diaan (approved):

PR #3390, opened 2026-05-27
fix(grid): Grid selection - missing selected state for single select

### LINKED ISSUE #2896: [Grid - Selection] Missing selected state for single select

Currently the single select variant is missing a selected / not selected state for the rows. The multiselect variant uses a checkbox, but the single select uses a button. Users should be aware that they can only select a single row, and assistive technology users should know when an item is selected or not. 

Single select grid

### 🤔 Expected Behavior?

When user enters the table, they are aware that they can only select one at a time. Assistive technology users are aware when an item is selected and when not. 

### 😯 Current Behavior

Assistive technology users are aware, that the items are buttons, but when activating the button, nothing indicates what the action was, if anything even happened. 

### 💁 Possible Solution

As users can select and deselect the items, these are closes to either an exclusive checkbox or switch/toggle button group.  Anyways, they should communicate whether they are selected or not. 

### 🖥️ Steps to Reproduce

1. Navigate to a row using screen reader
2. Activate the row by using space/enter -> Screen reader stays silent (or repeats the button name & role depending on sr)

Chrome

### If other, please specify.

_No response_

Windows 11

### 👤 Your name

Valtteri Järvinen

### 🧢 Your product/team

External

### 🎨 Your Theme(s)

Sanoma Learning

anna-lach:


a11ymiko (approved):

packages/components/grid/src/stories/selection.stories.ts
```diff
@@ -59,9 +71,19 @@ export const Activate: Story = {
         <sl-grid-column
           grow="3"
           header="Student"
-          .renderer=${(student: Student) => html`
-            <sl-button fill="link" variant="primary">${avatarRenderer(student)}</sl-button>
-          `}
+          .renderer=${(student: Student) => {
+            const isActive = activeStudent === student;
```
Diaan, line 75: This feels like we're putting a lot of logic on the CFA side. Is there no way we can keep this inside the grid component?

Diaan (approved):

packages/components/grid/src/stories/selection.stories.ts
anna-lach: I couldn't think of any other logical solution, but maybe @jpzwarte you have an idea for that one?

michal-sanoma (approved):

packages/components/grid/src/stories/selection.stories.ts
anna-lach: As discussed on Slack, we are going to keep it in the example stories. We need to describe it in the future in the documentation website as well.

a11ymiko (approved):

PR #3398, opened 2026-06-01
fix(combobox): Fix Space key selection for select-only comboboxes

### LINKED ISSUE #3053: [Combobox] Spacebar cannot be used to select and deselect options
## Current behavior

In expanded Combobox when keyboard-only users get to option they want to select the most common way is to use `spacebar`. But now `spacebar` is used to insert space inside Combobox's input field and it's not selecting or deselecting any option. 

Options can be selected/deselected with `Enter` key but it's not that common for keyboard-only users to interact with options this way.

https://github.com/user-attachments/assets/c1ee5294-2e2f-4358-bc34-a47439bd948f

https://github.com/user-attachments/assets/a0477d6e-7193-49b7-9adf-558a14f1dd13

## Suggested solution

It's recommended to add possibility of selecting option with `spacebar` but also still suport selecting with `Enter` like in example: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-both/

### 👤 Your name

Kacper

### 🧢 Your Product/Team

SLDS

michal-sanoma:
## Summary

- Allow select-only comboboxes to select and deselect the current option with the Space key
- Keep existing editable combobox behavior unchanged, so Space still inserts a space in the input field
- Add regression tests for single-select, multi-select, and editable combobox behavior

### BEFORE

https://github.com/user-attachments/assets/2eba7e99-46ba-43e2-9a96-cd734c81025a

### AFTER

https://github.com/user-attachments/assets/f1f71d90-c3ef-4f9b-bad2-f7520ad5cfb1

### DIFF
--- a/packages/components/combobox/src/combobox.ts
+++ b/packages/components/combobox/src/combobox.ts
@@ -683,7 +683,7 @@ export class Combobox<T = any, U = T> extends ObserveAttributesMixin(
   }
 
   #onKeydown(event: KeyboardEvent): void {
-    const isSelectOnlySpace = !!this.selectOnly && (event.key === ' ' || event.key === 'Spacebar');
+    const isSelectOnlySpace = !!this.selectOnly && event.key === ' ';
 
     if ((event.key === 'Enter' || isSelectOnlySpace) && !this.focusedTag) {
       if (isSelectOnlySpace) {
--- a/packages/components/combobox/src/combobox.ts
+++ b/packages/components/combobox/src/combobox.ts
@@ -683,7 +683,13 @@ export class Combobox<T = any, U = T> extends ObserveAttributesMixin(
   }
 
   #onKeydown(event: KeyboardEvent): void {
-    if (event.key === 'Enter' && !this.focusedTag) {
+    const isSelectOnlySpace = this.selectOnly && (event.key === ' ' || event.key === 'Spacebar');
+
+    if ((event.key === 'Enter' || isSelectOnlySpace) && !this.focusedTag) {
+      if (isSelectOnlySpace) {
+        event.preventDefault();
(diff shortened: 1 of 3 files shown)


packages/components/combobox/src/combobox.ts
```diff
@@ -683,7 +683,13 @@ export class Combobox<T = any, U = T> extends ObserveAttributesMixin(
   }
 
   #onKeydown(event: KeyboardEvent): void {
-    if (event.key === 'Enter' && !this.focusedTag) {
+    const isSelectOnlySpace = this.selectOnly && (event.key === ' ' || event.key === 'Spacebar');
```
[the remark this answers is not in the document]
michal-sanoma: fixed

a11ymiko (approved):

Diaan (approved):

packages/components/combobox/src/combobox.ts
```diff
@@ -683,7 +683,13 @@ export class Combobox<T = any, U = T> extends ObserveAttributesMixin(
   }
 
   #onKeydown(event: KeyboardEvent): void {
-    if (event.key === 'Enter' && !this.focusedTag) {
+    const isSelectOnlySpace = !!this.selectOnly && (event.key === ' ' || event.key === 'Spacebar');
```
jpzwarte: I'm not sure if `"Spacebar"` is needed; afaik we don't check for that anywhere else in the design system.

jpzwarte (approved):

packages/components/combobox/src/combobox.ts
michal-sanoma: removed

PR #3410, opened 2026-06-03
Chore/3399 add vitest visual testing support to components repo

a11ymiko:


a11ymiko:
This is not necessary to have. It's better to add few scenarios to storybook so they are covered in Components project in Chromatic than to have separate project there.

PR #3461, opened 2026-06-23
[2904] feat(checkbox): add infotip support to checkbox component

### LINKED ISSUE #2904: [Checkbox] Add/append a more info button to a single checkbox

I would like to give more context or explanation to a checkbox item without adding more lines.
This would apply to single checkboxes (not a checkbox group)

### 💁 Possible Solution

By adding a info-circle or question-circle icon only button to the end of the checkbox label text. Users can read more background or contextual info on the option they can choose

Idle
Image

Hover
Image

Clicked
Image

### 💻 Examples

Microsoft Power apps, adding column to a table (see screenshots @ possible solution)

### 👤 Your name

Jeremy Talen

### 🧢 Your Product/Team

Magister

Diaan:
This pull request adds support for integrating `sl-infotip` components within `sl-checkbox` elements, allowing infotips to be included alongside checkbox labels and ensuring they are accessible via keyboard navigation. The changes update both the checkbox and checkbox group components to handle infotips properly, including focus management and keyboard interaction, and update the Storybook stories to demonstrate the new functionality.

**Support for infotips in checkboxes:**

* [`checkbox.ts`](diffhunk://#diff-e36e433fb335ea6a39ad4d4abd96dbf9304ffaec47206ecd26cbb250518fa604L3-R4): Adds a new `infotip` state property to the `Checkbox` class, renders an additional slot for infotips, and updates event handling so infotips are focusable and can be toggled with keyboard events. Also includes logic to detect and store the infotip element. [[1]](diffhunk://#diff-e36e433fb335ea6a39ad4d4abd96dbf9304ffaec47206ecd26cbb250518fa604L3-R4) [[2]](diffhunk://#diff-e36e433fb335ea6a39ad4d4abd96dbf9304ffaec47206ecd26cbb250518fa604L22-R18) [[3]](diffhunk://#diff-e36e433fb335ea6a39ad4d4abd96dbf9304ffaec47206ecd26cbb250518fa604R115-R116) [[4]](diffhunk://#diff-e36e433fb335ea6a39ad4d4abd96dbf9304ffaec47206ecd26cbb250518fa604L210-R220) [[5]](diffhunk://#diff-e36e433fb335ea6a39ad4d4abd96dbf9304ffaec47206ecd26cbb250518fa604R278-R287) [[6]](diffhunk://#diff-e36e433fb335ea6a39ad4d4abd96dbf9304ffaec47206ecd26cbb250518fa604R349-R356)

* [`checkbox-group.ts`](diffhunk://#diff-37e8fdd8868d16b9d9349a9c4d1a3c0f6f4962538332538085e1f4c8268f403dR3): Updates the roving tabindex and focus management logic to include infotip elements, ensuring infotips within checkboxes are keyboard-navigable. Adds the `Infotip` import and updates type usage accordingly. [[1]](diffhunk://#diff-37e8fdd8868d16b9d9349a9c4d1a3c0f6f4962538332538085e1f4c8268f403dR3) [[2]](diffhunk://#diff-37e8fdd8868d16b9d9349a9c4d1a3c0f6f4962538332538085e1f4c8268f403dL23-R24) [[3]](diffhunk://#diff-37e8fdd8868d16b9d9349a9c4d1a3c0f6f4962538332538085e1f4c8268f403dL67-R80) [[4]](diffhunk://#diff-37e8fdd8868d16b9d9349a9c4d1a3c0f6f4962538332538085e1f4c8268f403dR226-R241)

* [`infotip.ts`](diffhunk://#diff-077408d31e0a1b723ada841c87506314f28afee3249fd9bf6d05985ad84bccb1R101): Adds keyboard event handling so infotips can be toggled with Enter/Space, and exposes a `toggleInfotip` method for programmatic toggling. [[1]](diffhunk://#diff-077408d31e0a1b723ada841c87506314f28afee3249fd9bf6d05985ad84bccb1R101) [[2]](diffhunk://#diff-077408d31e0a1b723ada841c87506314f28afee3249fd9bf6d05985ad84bccb1L115-R132)

**Storybook and documentation updates:**

* `checkbox-group.stories.ts` and `checkbox.stories.ts`: Updates stories to demonstrate infotip usage inside checkboxes, including examples with long labels and infotips for accessibility and UX testing. [[1]](diffhunk://#diff-b148a1b81122a383558fd7bcba276539d0c11518347ea856b710331744503fa8R4) [[2]](diffhunk://#diff-b148a1b81122a383558fd7bcba276539d0c11518347ea856b710331744503fa8L112-R123) [[3]](diffhunk://#diff-65adabe4495700b3969c86198da28fca640e10d91d704fc5c9dbc8f229c0e276R250-R261)

These changes improve accessibility and interactivity for checkboxes with infotips, ensuring users can access infotip content using both mouse and keyboard.


a11ymiko (approved):
I find infotips for 'Lorem ipsum' super helpful! 😆

packages/components/checkbox/src/checkbox.ts
```diff
@@ -327,6 +336,44 @@ export class Checkbox<T = any> extends ObserveAttributesMixin(FormControlMixin(L
     this.toggleAttribute('no-label', label.length === 0);
   }
```
michal-sanoma, line 339: **question:** Shouldn't we also add logic here (similar to what was added in switch/radio) to handle async loaded label text?


```
#onLabelSlotChange(): void {
    if (this.infotip && !this.infotip.describes) {
      this.infotip.describes = this.#labelText();
    }
  }
```

packages/components/infotip/src/infotip.ts
```diff
@@ -112,10 +121,44 @@ export class Infotip extends ScopedElementsMixin(LitElement) {
...
     this.renderRoot.querySelector('sl-popover')?.togglePopover();
   }
 
+  #onKeydown(event: KeyboardEvent): void {
+    if (event.key === 'Enter' || event.key === ' ') {
+      event.preventDefault();
+      event.stopPropagation();
+      this.toggleInfotip();
+    }
+  }
```
michal-sanoma, line 143: **suggestion:** We can simplify this method and preserve native button a11y behavior. If we only call `event.stopPropagation()` for Enter/Space (without `preventDefault()` and `toggleInfotip()`), the browser will naturally trigger the `@click` handler on keypress, keeping the standard keydown/keyup timing intact

packages/components/checkbox/src/checkbox-group.ts
```diff
@@ -64,11 +65,19 @@ export class CheckboxGroup<T = any> extends FormControlMixin(LitElement) {
...
     direction: 'vertical',
-    focusInIndex: (elements: Checkbox[]) => elements.findIndex(el => !el.disabled),
-    elements: () => this.boxes || [],
-    isFocusableElement: (el: Checkbox) => !el.disabled
+    focusInIndex: (elements: Array<Checkbox | Infotip>) =>
+      elements.findIndex(el =>
+        el instanceof Checkbox
+          ? !el.disabled
+          : el.parentElement instanceof Checkbox && !el.parentElement.disabled
+      ),
```
michal-sanoma: **suggestion:** Ddin't check this manually but else-branch check for Infotip in focusInIndex appears to be redundant. When entering the group, focus should always default to the main control (checkbox/radio), which is checked first in the findIndex loop and comes earlier in the elements array

I guess sth like this should be sufficient:

```focusInIndex: (elements: Array<Checkbox | Infotip>) =>
  elements.findIndex(el => el instanceof Checkbox && !el.disabled)
```

We have a similar thing in radio-group

michal-sanoma (approved):

packages/components/infotip/src/infotip.stories.ts
```diff
@@ -7,17 +8,24 @@ import { type Meta, type StoryObj } from '@storybook/web-components-vite';
...
 import '../register.js';
 
-type Props = { content: string };
+type Props = { content: string; size: ButtonSize };
 type Story = StoryObj<Props>;
 
 Icon.register(faStrawberry);
 
 export default {
   title: 'Utilities/Infotip',
```
anna-lach: In Figma 'infotip' is part of 'Overlay', maybe we should also move it to 'Overlay' in Storybook?

packages/components/infotip/src/infotip.ts
```diff
@@ -55,6 +56,12 @@ export class Infotip extends ScopedElementsMixin(LitElement) {
     };
   }
 
+  /** The name of the element that this infotip describes. */
+  @property() describes?: string;
+
+  /** The size of the infotip button. */
+  @property({ reflect: true }) size: ButtonSize = 'md';
```
anna-lach, line 63: Could you please update the 'infotip' documentation?

Probably it would be better to show here in those examples in the code and usage tab: https://sanomalearning.design/categories/components/infotip/usage/
small size of the infotip together with label?

anna-lach: Could you please also add the API documentation there? It looks like it's missing.

packages/components/switch/src/switch.ts
```diff
@@ -202,6 +207,7 @@ export class Switch<T = any> extends ObserveAttributesMixin(
     return html`
       <slot></slot>
       <slot @slotchange=${() => this.#onLabelSlotChange()} style="display: none"></slot>
+      <slot name="infotip" @slotchange=${() => this.#onInfotipSlotChange()}></slot>
```
anna-lach, line 210: Is it a proper focus order in the `reverse` version?
Probably switch should be focused first and then infotip? @a11ymiko what's your opinion?

https://github.com/user-attachments/assets/4e97d728-24c4-4897-ad44-5e5746126d8a

anna-lach (approved):

packages/components/switch/src/switch.ts
a11ymiko: I think right now it works just fine. It's said that focus order should be logical, and that sometimes might mean that something more to the right is focused before an item more to the left in the same line. 
In this case it's logical to first get information about what switch is used for and only after that to change its state. This way it's more useful for screen reader users, and sighted keyboard users also can easily understand why focus is 'skipping' switch at first tab and goes there after infotip ;)

a11ymiko (approved):

Issue #3224, opened 2026-04-15
Tag-list/Combobox keeps long items in the ".stack" wrapper


michal-sanoma:
### 😯 Current Behavior

When I pick an item with long text in the combobox then it lands in stack wrapper and user can remove it only by opening a combobox's listbox again and unselecting it

### 🤔 Expected Behavior/possible solution?

Item should be a removable tag in overflow version

https://storybook.sanomalearning.design/?path=/story/feedback-status-tag-tag--overflow&args=removable:!true

Another important note. We always want to have at least one removable tag visible when we have a `.stack` wrapper visible

### 🖥️ Steps to Reproduce

https://github.com/user-attachments/assets/71c522e4-e673-42d2-a4e0-63070376de29

### Were you able to reproduce this in our storybook?

Yes

### If yes, provide the storybook URL

https://storybook.sanomalearning.design/?path=/story/form-combobox-multiple--rich-content

### ⚖️ Impact

This blocks our release or stops us from using this component

### Framework

Lit

### What browsers are you seeing the problem on?

Chrome

### Component version

0.1.7

### 👤 Your name

Michał

PR #3474, opened 2026-06-30
fix(combobox tag): remove fake combobox tag focus

### LINKED ISSUE #2861: [Combobox] Tag list and Combobox interaction inaccessible with screen readers

The tag list interaction is inaccessible with screen readers, with different screen readers having different experiences. The explanations here are very text heavy, **I will demo this at our meeting and we will discuss this**. A part of the problem is how the tag list itself behaves, and part of the problem is related to the tag list actions when focused on the combobox or listbox. Tag list component problems are mentioned here for now since I've not yet fully tested them and created an issue for it. (TODO: probably redo the whole thing after discussion)

Multiselectable combobox with Breadcrumbs and Button items selected

### 💁 Possible Solution

This will need a discussion.

If going by the minimum (which in this case is probably the safest option), the tag list function (which is the problematic element here) is redundant for screen reader users at the current form. The combobox itself allows for selection and deselection already. This could probably be hidden from both keyboard and assistive technology users.

I get the idea of the tag list, and it would probably be immensely helpful for many users especially with list boxes that have more than a dozen options. If the tag list function is to be preserved, it needs to be separate from the combobox (i.e., when user is focused on the combobox, the tag list is not operable in any way). This would essentially make these work as two separate functions, which would be way easier to explain to an assistive technology user, and understand as an assistive technology user.

### 🖥️ Steps to Reproduce

**NVDA**
1. Focus on the combobox
2. Select items from the listbox -> items appear as a tag list in front of the combobox/text box
3. Use left or right keys to select items in the tag list

When items are already selected and the combobox is not focused:
1. Press tab -> Focus moves to the tag list, but screen reader stays in browse mode, left or right keys continue reading characters
2. Press tab -> Focus moves to combobox
3. Press shift + tab -> Focus moves back to tab list
4. Press NVDA key (insert or caps lock, depending on selection) + space -> Focus mode activates, user can select items using left or right keys

Firefox, Chrome, Safari

### If other, please specify.

_No response_

Windows, iOS 26.0.1, Android 15

### 👤 Your name

Valtteri Järvinen

### 🧢 Your product/team

External

### 🎨 Your Theme(s)

Sanoma Learning

### LINKED ISSUE #3009: [Combobox] Tags in active comboboxes can be focused and are not read by screen readers
## Current behavior 
All active comboboxes with tags: 
1. [Group Selected](https://storybook.sanomalearning.design/?path=/story/form-combobox-multiple--group-selected) 
2. [Groups With Group Selected](https://storybook.sanomalearning.design/?path=/story/form-combobox-multiple--groups-with-group-selected)
3. [Selected](https://storybook.sanomalearning.design/?path=/story/form-combobox-multiple--selected)
4. [Stacked](https://storybook.sanomalearning.design/?path=/story/form-combobox-multiple--stacked)
5. [Value](https://storybook.sanomalearning.design/?path=/story/form-combobox-multiple--value)
6. [Virtual List](https://storybook.sanomalearning.design/?path=/story/form-combobox-multiple--virtual-list) 
7. [Virtual List With Groups](https://storybook.sanomalearning.design/?path=/story/form-combobox-multiple--virtual-list-with-groups)

All active comboboxes with tags use `<sl-tag>` to present selected options. Each `<sl-tag>` has `aria-hidden='true'` and the first one in the list has `tabindex='0'` which make whole list focusable with arrow keys. While user's focus is on each tag its name is not being announced as the element is hidden with ARIA. 
`<sl-tag-list>` inside which tags are is not hidden with `aria-hidden='true'` and use can place screen reader there using arrow keys as well. 
Both `<sl-tag-list>` and `sl-tag` shouldn't be announced with assistive technologies as its user can use listbox to get information about what is selected.

https://github.com/user-attachments/assets/379ac732-5a30-4ea3-bd3d-eb23a4409303

https://github.com/user-attachments/assets/6c51f998-03bc-409e-970b-d0d8ca6b7508

## Proposed solution
1. `sl-tag-list` should be hidden with `aria-hidden='true'` so user won't place screen reader focus on that element.
 2. All `sl-tag` should have `tabindex='-1'` or no `tabindex` at all, so keyboard focus can't be placed there. 

### 👤 Your name

Kacper

### 🧢 Your Product/Team

SLDS team

michal-sanoma:
## What changed

This PR improves keyboard and screen reader behavior for removable tags used in comboboxes and tag lists.

### Combobox

- Removed the old fake tag focus state from `<sl-combobox>`.
- Removed the `.focused` styling that visually placed focus on the whole tag.
- Left/Right arrow keys no longer move fake focus from the combobox input to selected tags.
- Up/Down arrow keys from the combobox input remain available for navigating the option list.
- When selected tags are present, the combobox uses two tab stops:
  - the first focusable removable tag button
  - the combobox input
- Once focus is inside the tag list, arrow keys move between removable tag buttons.
- After removing a tag, focus moves to the next/previous visible removable tag button when available, otherwise back to the input.

### Tag / Tag list

- Added `aria-hidden="true"` to the hidden navigation description text while keeping it referenced through `aria-describedby`.
- Added an internal `keyboardNavigation` switch for `<sl-tag-list>` so roving keyboard navigation can be disabled where needed.
- Added tests for the updated combobox tab/arrow behavior and tag-list navigation description behavior.

### DIFF
--- a/.changeset/nine-singers-worry.md
+++ b/.changeset/nine-singers-worry.md
@@ -3,4 +3,4 @@
 '@sl-design-system/tag': patch
 ---
 
-Improved keyboard and screen reader behavior for removable tags in comboboxes and tag lists. Comboboxes no longer move fake focus from the input to selected tags with Left/Right arrow keys, while tag navigation remains available once focus is inside the tag list. The hidden tag-list navigation instructions are now `aria-hidden` while still being exposed through `aria-describedby`
+Improved keyboard and screen reader behavior for removable tags in comboboxes and tag lists. Comboboxes no longer move fake focus from the input to selected tags with Left/Right arrow keys, while tag navigation remains available once focus is inside the tag list. The hidden tag-list navigation instructions are now `aria-hidden` while still being exposed through `aria-describedby`.
--- a/.changeset/nine-singers-worry.md
+++ b/.changeset/nine-singers-worry.md
@@ -1,0 +1,6 @@
+---
+'@sl-design-system/combobox': patch
+'@sl-design-system/tag': patch
+---
+
+Improved keyboard and screen reader behavior for removable tags in comboboxes and tag lists. Comboboxes no longer move fake focus from the input to selected tags with Left/Right arrow keys, while tag navigation remains available once focus is inside the tag list. The hidden tag-list navigation instructions are now `aria-hidden` while still being exposed through `aria-describedby`
--- a/packages/components/combobox/src/combobox.scss
+++ b/packages/components/combobox/src/combobox.scss
@@ -56,10 +56,6 @@ sl-tag-list {
 sl-tag {
   outline-offset: 0;
   outline-width: 0;
-
-  &.focused {
-    outline: var(--sl-size-borderWidth-default) solid var(--sl-color-border-focused);
-  }
 }
 
 sl-icon {
--- a/packages/components/combobox/src/combobox.scss
+++ b/packages/components/combobox/src/combobox.scss
@@ -46,11 +46,15 @@ sl-text-field {
 }
 
 sl-tag-list {
-  flex: 1 1 auto;
+  flex: 0 1 auto;
   max-inline-size: 66%;
   min-inline-size: 0;
   position: relative;
   z-index: 1;
+}
+
+sl-tag-list[data-stacked-active] {
+  flex-grow: 1;
 }
 
 sl-tag {
--- a/packages/components/combobox/src/combobox.spec.ts
+++ b/packages/components/combobox/src/combobox.spec.ts
@@ -1266,8 +1266,7 @@ describe('sl-combobox', () => {
         el.style.maxInlineSize = '300px';
         el.value = ['Option 1', 'Option 2', 'Option 3', 'Option 4', 'Option 5', 'Option 6'];
         await el.updateComplete;
-        await new Promise(resolve => setTimeout(resolve, 300));
-        await el.updateComplete;
+        await waitForNextFrame();
         await waitForNextFrame();
 
         const tagList = el.renderRoot.querySelector('sl-tag-list')!,
@@ -1328,8 +1327,7 @@ describe('sl-combobox', () => {
             count
           );
           await el.updateComplete;
-          await new Promise(resolve => setTimeout(resolve, 300));
-          await el.updateComplete;
+          await waitForNextFrame();
           await waitForNextFrame();
 
           const inputWidth = input.getBoundingClientRect().width,
--- a/packages/components/combobox/src/combobox.spec.ts
+++ b/packages/components/combobox/src/combobox.spec.ts
@@ -1238,6 +1238,28 @@ describe('sl-combobox', () => {
         expect(gap).to.be.lessThan(16);
       });
 
+      it('should keep one selected tag visible next to the stack counter in limited space', async () => {
+        el.style.maxInlineSize = '300px';
+        el.value = ['Option 1', 'Option 2', 'Option 3', 'Option 4', 'Option 5', 'Option 6'];
+        await el.updateComplete;
+        await new Promise(resolve => setTimeout(resolve, 300));
+        await el.updateComplete;
+        await waitForNextFrame();
+
+        const tagList = el.renderRoot.querySelector('sl-tag-list')!,
+          stackTag = tagList.renderRoot.querySelector('.stack sl-tag'),
+          selectedTags = Array.from(el.renderRoot.querySelectorAll('sl-tag')),
+          visibleSelectedTags = selectedTags.filter(
+            tag => getComputedStyle(tag).display !== 'none'
+          );
+
+        expect(stackTag).to.exist;
+        expect(stackTag).to.have.trimmed.text(
+          `+${selectedTags.length - visibleSelectedTags.length}`
+        );
+        expect(visibleSelectedTags.length).to.be.greaterThan(0);
+      });
+
       it('should not flicker when selecting many items in a limited space', async () => {
         el.style.maxInlineSize = '300px';
 
--- a/packages/components/combobox/src/combobox.spec.ts
+++ b/packages/components/combobox/src/combobox.spec.ts
@@ -942,6 +942,21 @@ describe('sl-combobox', () => {
         expect(options[1]).to.be.displayed;
         expect(options[2]).to.be.displayed;
       });
+
+      it('should ignore option navigation when filtering hides all options', async () => {
+        input.focus();
+        await userEvent.keyboard('Foo');
+        await el.updateComplete;
+
+        expect(el.items.filter(item => item.type === 'option' && item.visible)).to.have.lengthOf(0);
+
+        await userEvent.keyboard('{ArrowDown}');
+        await userEvent.keyboard('{Home}');
+        await el.updateComplete;
+
+        expect(el.currentItem).to.be.undefined;
+        expect(input).not.to.have.attribute('aria-activedescendant');
+      });
     });
 
     describe('current item on open', () => {
--- a/packages/components/combobox/src/combobox.spec.ts
+++ b/packages/components/combobox/src/combobox.spec.ts
@@ -1247,37 +1247,27 @@ describe('sl-combobox', () => {
       });
 
       it('should keep the input width bounded while adding tags in limited space', async () => {
-        vi.useFakeTimers();
+        el.style.maxInlineSize = '300px';
+        const textField = el.renderRoot.querySelector('sl-text-field') as HTMLElement;
 
-        try {
-          el.style.maxInlineSize = '300px';
-          const textField = el.renderRoot.querySelector('sl-text-field') as HTMLElement;
+        for (const count of [1, 2, 3, 4, 5, 6]) {
+          el.value = ['Option 1', 'Option 2', 'Option 3', 'Option 4', 'Option 5', 'Option 6'].slice(
+            0,
+            count
+          );
+          await el.updateComplete;
+          await new Promise(resolve => setTimeout(resolve, 300));
+          await el.updateComplete;
+          await waitForNextFrame();
 
-          for (const count of [1, 2, 3, 4, 5, 6]) {
-            el.value = [
-              'Option 1',
-              'Option 2',
-              'Option 3',
-              'Option 4',
-              'Option 5',
-              'Option 6'
-            ].slice(0, count);
-            await el.updateComplete;
-            await vi.advanceTimersByTimeAsync(300);
-            await el.updateComplete;
-            await waitForNextFrame();
+          const inputWidth = input.getBoundingClientRect().width,
+            fieldWidth = textField.getBoundingClientRect().width,
+            comboboxWidth = el.getBoundingClientRect().width;
 
-            const inputWidth = input.getBoundingClientRect().width,
-              fieldWidth = textField.getBoundingClientRect().width,
-              comboboxWidth = el.getBoundingClientRect().width;
-
-            expect(inputWidth).to.be.a('number');
-            expect(Number.isFinite(inputWidth)).to.be.true;
-            expect(inputWidth).to.be.at.most(fieldWidth + 0.5);
-            expect(comboboxWidth).to.be.at.most(300.5);
-          }
-        } finally {
-          vi.useRealTimers();
+          expect(inputWidth).to.be.a('number');
+          expect(Number.isFinite(inputWidth)).to.be.true;
+          expect(inputWidth).to.be.at.most(fieldWidth + 0.5);
+          expect(comboboxWidth).to.be.at.most(300.5);
         }
       });
 
--- a/packages/components/combobox/src/combobox.spec.ts
+++ b/packages/components/combobox/src/combobox.spec.ts
@@ -1272,16 +1272,69 @@ describe('sl-combobox', () => {
         expect(removable).to.be.true;
       });
 
-      it('should not show fake tag focus when navigating remove buttons', async () => {
-        const tags = Array.from(el.renderRoot.querySelectorAll('sl-tag')),
-          button = tags[0].renderRoot.querySelector('button');
+      it('should use the first removable tag button and input as combobox tab stops', async () => {
+        const wrapper = await fixture<HTMLDivElement>(html`
+            <div>
+              <button>Before</button>
+              <sl-combobox multiple .value=${['Option 1', 'Option 2']}>
+                <sl-listbox>
+                  <sl-option>Option 1</sl-option>
+                  <sl-option>Option 2</sl-option>
+                  <sl-option>Option 3</sl-option>
+                </sl-listbox>
+              </sl-combobox>
+              <button>After</button>
+            </div>
+          `),
+          combobox = wrapper.querySelector('sl-combobox')!,
+          input = combobox.querySelector<HTMLInputElement>('input[slot="input"]')!;
 
-        button?.dispatchEvent(
-          new KeyboardEvent('keydown', { bubbles: true, composed: true, key: 'ArrowRight' })
-        );
+        await combobox.updateComplete;
+        await waitForNextFrame();
+        await combobox.updateComplete;
+
+        const tags = Array.from(combobox.renderRoot.querySelectorAll('sl-tag')),
+          buttons = tags.map(tag => tag.renderRoot.querySelector('button'));
+
+        expect(tags).to.have.lengthOf(2);
+
+        wrapper.querySelector('button')!.focus();
+
+        await userEvent.tab();
+
+        expect(combobox.renderRoot.activeElement).to.equal(tags[0]);
+        expect(tags[0].shadowRoot?.activeElement).to.equal(buttons[0]);
+
+        await userEvent.keyboard('{ArrowRight}');
+
+        expect(combobox.renderRoot.activeElement).to.equal(tags[1]);
+        expect(tags[1].shadowRoot?.activeElement).to.equal(buttons[1]);
+
+        await userEvent.keyboard('{ArrowLeft}');
+
+        expect(combobox.renderRoot.activeElement).to.equal(tags[0]);
+        expect(tags[0].shadowRoot?.activeElement).to.equal(buttons[0]);
+
+        await userEvent.tab();
+
+        expect(document.activeElement).to.equal(input);
+
+        await userEvent.tab();
+
+        expect(document.activeElement).to.equal(wrapper.querySelector('button:last-child'));
+      });
+
+      it('should not show fake tag focus when navigating from the input with arrow keys', async () => {
+        const tags = Array.from(el.renderRoot.querySelectorAll('sl-tag'));
+
+        input.focus();
+        input.setSelectionRange(0, 0);
+
+        await userEvent.keyboard('{ArrowLeft}');
         await el.updateComplete;
 
         expect(tags.some(tag => tag.classList.contains('focused'))).to.be.false;
+        expect(document.activeElement).to.equal(input);
       });
 
       it('should stack options when there is limited space', async () => {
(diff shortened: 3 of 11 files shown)


packages/components/tag/src/tag.ts
```diff
@@ -190,7 +190,7 @@ export class Tag extends ScopedElementsMixin(LitElement) {
             </button>
             ${this.navigationDescription
               ? html`
-                  <span id="navigation-description" class="visually-hidden"
+                  <span id="navigation-description" class="visually-hidden" aria-hidden="true"
```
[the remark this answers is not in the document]
michal-sanoma: this is intended

.changeset/nine-singers-worry.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sl-design-system/combobox': patch
+'@sl-design-system/tag': patch
+---
+
+Improved keyboard and screen reader behavior for removable tags in comboboxes and tag lists. Comboboxes no longer move fake focus from the input to selected tags with Left/Right arrow keys, while tag navigation remains available once focus is inside the tag list. The hidden tag-list navigation instructions are now `aria-hidden` while still being exposed through `aria-describedby`
```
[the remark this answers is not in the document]
michal-sanoma: added

packages/components/tag/src/tag-list.ts
```diff
@@ -239,6 +246,11 @@ export class TagList extends ScopedElementsMixin(LitElement) {
     super.updated(changes);
 
     this.#syncTags();
+
+    if (changes.has('keyboardNavigation')) {
+      this.#rovingTabindexController.clearElementCache();
+    }
+
     this.#syncRovingTabindexController();
```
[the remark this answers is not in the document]
michal-sanoma: added

a11ymiko (commented):
Right now tag with tooltip with list of hidden elements is not announced properly by NVDA. Screen reader is not announcing the text inside the tooltip when focus is placed on the tag. When placed on tag with text '+3' NVDA announces "Selected options  list  with 6 items  +3" without information from the tooltip.
This is due to NVDA not announcing `aria-labelledby` on non-interactive elements and `sl-tag` has `role=listitem`.

image


There are two possible ways to fix it for NVDA without harm to VoiceOver.
1. Add text form tooltip as content of `sl-tag`. Make it hidden with CSS, but read by screen readers. Proposed 
```
<sl-tag role='listbox' tabindex='0' aria-labelledby='tooltip'>
  "+"
  "3"
  <span class='visually-hidden'>TOOLTIP TEXT</span>
</sl-tag>
```
2. If the first option does not work as wanted then `aria-labelledby` should be removed from `sl-tag` and inside of `sl-tag` there should be added new `<span role='button' aria-disabled='true' aria-labelledby='tooltip'>`. Inside of this span currently visible text of tag e.g. "+3" may be placed.

michal-sanoma:
> Right now tag with tooltip with list of hidden elements is not announced properly by NVDA. Screen reader is not announcing the text inside the tooltip when focus is placed on the tag. When placed on tag with text '+3' NVDA announces "Selected options list with 6 items +3" without information from the tooltip. This is due to NVDA not announcing `aria-labelledby` on non-interactive elements and `sl-tag` has `role=listitem`.
>
> 
> 1. Add text form tooltip as content of `sl-tag`. Make it hidden with CSS, but read by screen readers. Proposed
> 
> ```
> <sl-tag role='listbox' tabindex='0' aria-labelledby='tooltip'>
>   "+"
>   "3"
>   <span class='visually-hidden'>TOOLTIP TEXT</span>
> </sl-tag>
> ```
> 
> 2. If the first option does not work as wanted then `aria-labelledby` should be removed from `sl-tag` and inside of `sl-tag` there should be added new `<span role='button' aria-disabled='true' aria-labelledby='tooltip'>`. Inside of this span currently visible text of tag e.g. "+3" may be placed.

I've introduced this solution we discussed on Slack:

```
<sl-tag
  role="listitem"
  tabindex="0"
  aria-describedby="hidden-elements-description">
  +3
  <span id="hidden-elements-description" class="visually-hidden">
    List of hidden elements: Option 1, Option 2, Option 3
  </span>
</sl-tag>
```

packages/components/combobox/src/combobox.spec.ts
```diff
@@ -1194,6 +1194,24 @@ describe('sl-combobox', () => {
...
+    tag => getComputedStyle(tag).display !== 'none'
+  ),
+  lastTag = visibleTags.at(-1);
+
+expect(lastTag, 'expected at least one visible tag').to.exist;
+
+const inputRect = input.getBoundingClientRect(),
+  lastTagRect = lastTag!.getBoundingClientRect(),
+  gap = inputRect.left - lastTagRect.right;
+
```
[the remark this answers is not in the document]
michal-sanoma: added

packages/components/tag/src/tag.ts
```diff
@@ -160,6 +163,12 @@ export class Tag extends ScopedElementsMixin(LitElement) {
       buttonDescription = [
         this.tooltip ? 'tooltip' : undefined,
         this.navigationDescription ? 'navigation-description' : undefined
+      ]
+        .filter(Boolean)
+        .join(' '),
+      labelDescription = [
```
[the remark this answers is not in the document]
michal-sanoma: added

a11ymiko (approved):
Nice work! Huuuge improvement! 🥳

.changeset/nine-singers-worry.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sl-design-system/combobox': patch
+'@sl-design-system/tag': patch
+---
+
+Improved keyboard and screen reader behavior for removable tags in comboboxes and tag lists. Comboboxes no longer move fake focus from the input to selected tags with Left/Right arrow keys, while tag navigation remains available once focus is inside the tag list. The hidden tag-list navigation instructions are now `aria-hidden` while still being exposed through `aria-describedby`.
```
anna-lach, line 6: Could you please also update documentation? https://website-3474.d3c5h2uor7jlkj.amplifyapp.com/categories/components/combobox/accessibility/#keyboard-interactions

anna-lach: Probably this `updateInternalValidity` method should be marked as `@internal`, so we don't want to show it as part of the API?

Image

https://website-3474.d3c5h2uor7jlkj.amplifyapp.com/categories/components/combobox/code/#methods

packages/components/combobox/src/combobox.spec.ts
```diff
@@ -1194,75 +1220,80 @@ describe('sl-combobox', () => {
         expect(hostStyles.contain).to.include('inline-size');
       });
 
+      it('should keep the input caret next to the visible tags', async () => {
```
anna-lach, line 1247: There is still an issue in rich content example
https://storybook-3474.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/form-combobox-multiple--rich-content

Image

anna-lach: I missed that, but as far as I remember we always wanted to have at least one tag visible except stack tag, so sth like `[stack tag][tag 1]`, is that true @arecuenco-dsgn ? Is it still valid?

packages/components/tag/src/tag.ts
```diff
@@ -190,7 +202,7 @@ export class Tag extends ScopedElementsMixin(LitElement) {
             </button>
             ${this.navigationDescription
               ? html`
-                  <span id="navigation-description" class="visually-hidden"
+                  <span id="navigation-description" class="visually-hidden" aria-hidden="true"
```
anna-lach, line 205: Why `aria-hidden=true`? We already have it visually-hidden, so only visible for screen readers, why we want to have it set as visible for screen readers only and not visible for screen readers (by setting aria-hidden=true) at once?

packages/components/combobox/src/combobox.scss
```diff
@@ -46,20 +46,20 @@ sl-text-field {
...
   min-inline-size: 0;
   position: relative;
   z-index: 1;
 }
 
+sl-tag-list[data-stacked-active] {
+  flex-grow: 1;
+}
+
 sl-tag {
```
anna-lach, line 60: When I focus remove button there is an outline-color animation and outline color changes from black to blue, it looks a bit weird, could you please just get rid of that black outline?

https://github.com/user-attachments/assets/3350e95a-0bc2-4493-a231-432c518a3254

packages/components/tooltip/src/tooltip-shared.spec.ts
```diff
@@ -14,6 +14,21 @@ describe('sl-tooltip shared', () => {
   let tooltip: Tooltip;
 
   const waitFor = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
+  const waitForPopoverToOpen = async (popover: HTMLElement, timeout = 1500): Promise<void> => {
```
anna-lach, line 17: Please add a changeset for tooltip changes,

anna-lach, line 17: Please add a changeset for tooltip changes.

.changeset/nine-singers-worry.md
michal-sanoma: good point, I totally forgot about the docs - added

packages/components/combobox/src/combobox.spec.ts
michal-sanoma: nice catch, I fixed the caret here and added this change that we show long tag but I'm not sure if this looks good

Will need an opinion of Alvaro here. Please note that I have a task in this area on me. We can also address this issue there https://github.com/sl-design-system/components/issues/3224

Screenshot 2026-07-14 at 12 00 40

packages/components/tag/src/tag.ts
michal-sanoma: Good question. The span is intentionally both visually hidden and `aria-hidden`. We do not want it to be discoverable as standalone hidden content in browse/navigation modes, but it is still referenced from the remove button via `aria-describedby`, so it remains available in the button context

packages/components/combobox/src/combobox.scss
michal-sanoma: Wow, amazing catch! Should be better now

packages/components/tooltip/src/tooltip-shared.spec.ts
michal-sanoma: This change only stabilizes the tooltip tests by waiting for the popover to actually open instead of relying on a fixed delay. It does not change tooltip runtime behavior or public API, so I don’t think it needs a changeset

michal-sanoma: same as above

packages/components/combobox/src/combobox.spec.ts
anna-lach: So maybe it would be better to work on it in that issue (#3224) then? :)

packages/components/combobox/src/combobox.ts
```diff
@@ -737,41 +736,13 @@ export class Combobox<T = any, U = T> extends ObserveAttributesMixin(
...
     } else if (
       !this.wrapper?.matches(':popover-open') &&
       ['ArrowDown', 'ArrowUp'].includes(event.key)
     ) {
       this.#popoverOpenedViaKeyboard = true;
       this.wrapper?.showPopover();
-    } else if (['ArrowDown', 'ArrowUp', 'End', 'Home'].includes(event.key) && !this.focusedTag) {
+    } else if (['ArrowDown', 'ArrowUp', 'End', 'Home'].includes(event.key)) {
       event.preventDefault();
       event.stopPropagation();
```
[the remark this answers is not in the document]
michal-sanoma: added

packages/components/tooltip/src/tooltip-shared.spec.ts
anna-lach: Imo there should be a changeset added for any change in the SLDS packages (then we will have a new version released, with patch in that case and people should be informed there is a new version that should be updated, e.g. to get it working properly with combobox).

anna-lach: oh ok, only for tests, I see it now 🥲😁 sorry, I agree 🙂

anna-lach: So forget my previous comment. 😉

michal-sanoma: I agree that runtime changes in SLDS packages should get a changeset. In this case, though, the tooltip change is test-only: it only changes the assertion timing in `tooltip-shared.spec.ts` and does not touch tooltip source, styles, public API, or emitted package output. So I don’t think there is a user-facing tooltip change to release here

anna-lach: Yes, sure, I missed that those changes were only for tests. 😄

packages/components/combobox/src/combobox.spec.ts
michal-sanoma: For me it's ok :)

arecuenco-dsgn: We should always keep at least **one tag visible**. If the first selected tag is too long to fit, the fallback should be to show that first tag with overflow/truncation, not replace it with the counter. That way we still preserve the selected value visually inside the combobox.

I think the fallback logic should follow this rule every time:
- Always show the first tag.
- Allow [overflow](https://storybook.sanomalearning.design/?path=/story/feedback-status-tag-tag--overflow-removable&globals=theme:bingel-dc) (truncation) on the first tag when needed.
- Only show the counter for additional hidden tags.
- Make sure the fallback always preserves the **“at least 1 tag visible”** rule.

What do you think @anna-lach and @michal-sanoma?

anna-lach: I agree, but what do you mean @arecuenco-dsgn by 'first tag'? I think it should be the last tag (option) selected, so `[2][last tag selected]`. 

Or when there is only one selected, but too long to show the whole text, there should be sth like:
`[Tag too long to ...]`. 

When there are more options selected, but the last one is too long to show the shole text:
`[3][Tag too long to ...]`. 

Do I understand you correctly? :)

website/src/categories/components/combobox/accessibility.md
```diff
@@ -14,9 +14,9 @@ eleventyNavigation:
 
 |Command|Description|
 |-|-|
-|Tab|Moves the focus into/outside of the component. Use arrow down to open the drop down.|
+|Tab|Moves the focus into/outside of the component. In a multiple combobox with selected removable tags, focus moves between the first removable tag and the combobox input.|
 |Arrow up/down|Moves the active option 1 position up/down. The focus indicator loops, so when you are at the last option and press "down" it will focus on the first option.|
-|Arrow left/right|When the combobox supports multiple selection, moves the focus to the tag on the left/right, assuming the caret is already at the beginning of the input.|
+|Arrow left/right|When focus is on a removable tag in the tag list, moves focus to the previous/next removable tag. Arrow left/right does not move focus from the combobox input to selected tags.|
```
anna-lach: Actually focus moves between remove buttons in removable tags (before it was moving between tags, when tags were focusable).

website/src/categories/components/combobox/accessibility.md
```diff
@@ -14,9 +14,9 @@ eleventyNavigation:
 
 |Command|Description|
 |-|-|
-|Tab|Moves the focus into/outside of the component. Use arrow down to open the drop down.|
+|Tab|Moves the focus into/outside of the component. In a multiple combobox with selected removable tags, focus moves between the first removable tag and the combobox input.|
```
anna-lach:
```suggestion
|Tab|Moves the focus into/outside of the component. In a multiple combobox with selected removable tags, focus moves between the remove button of the first removable tag and the combobox input.|
```

anna-lach (approved):

website/src/categories/components/combobox/accessibility.md
michal-sanoma: good catch, added

website/src/categories/components/combobox/accessibility.md
```diff
@@ -14,12 +14,13 @@ eleventyNavigation:
 
 |Command|Description|
 |-|-|
-|Tab|Moves the focus into/outside of the component. Use arrow down to open the drop down.|
+|Tab|Moves the focus into/outside of the component. In a multiple combobox with selected removable tags, focus moves between the remove button of the first removable tag and the combobox input.|
 |Arrow up/down|Moves the active option 1 position up/down. The focus indicator loops, so when you are at the last option and press "down" it will focus on the first option.|
```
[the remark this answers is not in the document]
michal-sanoma: added

packages/components/tag/src/tag-list.spec.ts
```diff
@@ -561,7 +628,7 @@ describe('sl-tag-list', () => {
       expect(el.stackSize).to.equal(2);
     });
 
-    it('should hide the last tag when it does not fit in the remaining width', async () => {
+    it('should keep the last tag visible even when it does not fit in the remaining width', async () => {
       el = await fixture(html`
         <sl-tag-list stacked style="gap: 10px; padding: 0; margin: 0; border: none;">
```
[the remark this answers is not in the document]
michal-sanoma: added

packages/components/tag/src/tag-list.ts
```diff
@@ -239,6 +246,15 @@ export class TagList extends ScopedElementsMixin(LitElement) {
...
 
     this.#syncTags();
+
+    if (changes.has('keyboardNavigation')) {
+      this.#rovingTabindexController.clearElementCache();
+
+      if (!this.keyboardNavigation) {
+        this.#clearManagedTabindexes();
+      }
+    }
```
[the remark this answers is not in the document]
michal-sanoma: added

website/src/categories/components/combobox/accessibility.md
```diff
@@ -14,12 +14,13 @@ eleventyNavigation:
 
 |Command|Description|
 |-|-|
-|Tab|Moves the focus into/outside of the component. Use arrow down to open the drop down.|
+|Tab|Moves the focus into/outside of the component. In a multiple combobox with selected removable tags, focus moves between the first removable tag and the combobox input.|
```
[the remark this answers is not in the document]
michal-sanoma: added

packages/components/tag/src/tag-list.ts
```diff
@@ -254,23 +270,22 @@ export class TagList extends ScopedElementsMixin(LitElement) {
   }
 
   override render(): TemplateResult {
+    const hiddenTagsDescription = this.stacked ? this.#getHiddenTagsDescription() : '';
```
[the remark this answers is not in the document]
michal-sanoma: added

packages/components/tag/src/tag-list.ts
```diff
@@ -550,8 +580,14 @@ export class TagList extends ScopedElementsMixin(LitElement) {
     this.#syncRovingTabindexController();
   }
 
+  #clearManagedTabindexes(): void {
+    this.tags.forEach(tag => tag.removeAttribute('tabindex'));
+    this.stackTag?.removeAttribute('tabindex');
+  }
```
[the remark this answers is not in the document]
michal-sanoma: added

packages/components/combobox/src/combobox.spec.ts
arecuenco-dsgn: I meant the first Tag in the tag list, but now I'm questioning myself. Here are the behaviours:

1. If there isn't enough space, the Tag must have overflow.
`[Tag too long to ...]`

2. If there are multiple selections, we need to show at least one Tag. To show it, it must have overflow.
`[2]` `[Tag too long to ...]`

3. If there is space to show more Tags, make the last Tag with overflow.
`[2]` `[Middle Tag]``[Last Tag too long to ...]`

What do you think, @anna-lach and @michal-sanoma?

michal-sanoma: For me this sounds reasonable

anna-lach: I agree with first and second example, but third one might be hard, so I'd rather stick with that option
 `[2] [Tag too long to ...]` instead of `[2] [Middle Tag]``[Last Tag too long to ...]`

PR #3483, opened 2026-07-03
fix(select): accessibility improvements

### LINKED ISSUE #3435: [Select] Listbox in `sl-select` does not have accessible name and is not referenced with `aria-controls`
### 😯 Current Behavior

`sl-select-button` is not referencing listbox with `aria-controls`
`sl-listbox` does not have accessible name provided with `aria-labelledby` or `aria-label`

Image

`sl-select-button` should have `aria-controls='LISTBOX_ID'` to reference to listbox
`sl-listbox` should have `aria-labelledby='SELECT_LABEL_ID'` to have accessible name that is label of select element. 

### 🖥️ Steps to Reproduce

1. Go to https://storybook.sanomalearning.design/?path=/story/form-select--basic
2. Expand listbox
3. Go to Accessibility tab in Storybook add-ons and refresh tests
4. Check if 'ARIA input field name' and 'ARIA required attributes' are present in test report

Yes

https://storybook.sanomalearning.design/?path=/story/form-select--basic

_No response_

### ⚖️ Impact

### Framework

Other

All

### If other, please specify.

_No response_

_No response_

### Component version

2.1.4

### Theme version

_No response_

### 👤 Your name

Kacper

### 🧢 Your product/team

SLDS

### 🎨 Your Theme(s)

_No response_

anna-lach:


a11ymiko (approved):

packages/components/select/src/select.spec.ts
```diff
@@ -133,6 +133,12 @@ describe('sl-select', () => {
       expect(el.button).to.have.attribute('aria-labelledby', 'id');
     });
 
+    it('should set ariaControlsElements via internals on the button pointing to the listbox', async () => {
```
Diaan:
```suggestion
    it('should set ariaControlsElements on the button pointing to the listbox', async () => {
```

The 'how' doesn't really matter in the test. You just need to make sure it's there

Diaan (approved):

packages/components/select/src/select.ts
```diff
@@ -332,19 +332,61 @@ export class Select<T = any> extends ObserveAttributesMixin(
...
+        buttonAriaLabelledBy = this.button.getAttribute('aria-labelledby')?.trim(),
+        ariaLabel = hostAriaLabel || buttonAriaLabel,
+        explicitLabelledByElements =
+          (this.button.ariaLabelledByElements && [...this.button.ariaLabelledByElements]) ||
+          this.#resolveLabelledByElements(hostAriaLabelledBy || buttonAriaLabelledBy),
+        hasExplicitLabel = Boolean(ariaLabel) || explicitLabelledByElements.length > 0;
+
+      // Use element references so labeling works across the shadow boundary.
+      if (!hasExplicitLabel && labels.length) {
+        this.listbox!.removeAttribute('aria-label');
```
michal-sanoma: **suggestion:** Can we add a safety check for `this.listbox` here instead of non-null assertion? 

I mean `if (!hasExplicitLabel && labels.length && this.listbox) {`

packages/components/select/src/select.spec.ts
```diff
@@ -682,6 +688,29 @@ describe('sl-select', () => {
...
+
+      expect(labels.length).to.equal(1);
+      expect(button.ariaLabelledByElements).to.deep.equal(labels);
+    });
+
+    it('should set ariaLabelledByElements on the listbox from associated labels', async () => {
+      const select = el.renderRoot.querySelector('sl-select') as Select,
+        labels = Array.from(select.internals.labels) as Element[];
+
+      select.querySelector<SelectButton>('sl-select-button')?.click();
```
michal-sanoma: **nitpick:** clicking the select button to open the listbox is unnecessary since the listbox is rendered unconditionally

packages/components/select/src/select.ts
```diff
@@ -332,19 +332,61 @@ export class Select<T = any> extends ObserveAttributesMixin(
...
-      if (this.internals.labels.length) {
-        // Set the aria-label of the button to the concatenated text content of all labels
-        // FIXME: This is a workaround because we do not yet have access to `referenceTarget`
-        this.button.setAttribute(
-          'aria-labelledby',
-          Array.from(this.internals.labels)
-            .map(label => (label as HTMLLabelElement).id)
-            .join(' ')
-        );
+      if (this.listbox) {
```
michal-sanoma, line 346: **question:** Should we also handle later `aria-label / aria-labelledby` changes here? Since the listbox gets its accessible name manually, doing this only in `firstUpdated()` could leave it out of sync with the button

anna-lach: Good point! I've improved it. :)

michal-sanoma (approved):

PR #3495, opened 2026-07-07
Fix token import

### LINKED ISSUE #3513: Fix token import
The import from token studio and the generation of new theme css files and icon files is broken after the changes in Token studio.

This needs to be fixed

Diaan:


scripts/import-icons.js
```diff
@@ -83,13 +82,15 @@ const buildIcons = async theme => {
       icon: { style, themeIcons },
       text
     }
-  } = await import(`../packages/tokens/src/tokens/${theme}/base.json`, { with: { type: 'json' } });
+  } = await import(`../packages/tokens/src/tokens/${theme}/core.json`, { with: { type: 'json' } });
```
michal-sanoma: **question:** isn't this a dead code now?  As I understand we migrated to `base-new.json` everywhere, right?

packages/tokens/src/tokens/teacher-assistant/base-new.json
```diff
@@ -0,0 +1,301 @@
+{
```
michal-sanoma, line 1: **question:** Is it intentional that this new teacher-assistant theme is not added to `package.json`, `tsconfig.json`, and `index.ts`?

scripts/build-themes.js
```diff
@@ -80,6 +80,88 @@ StyleDictionary.registerPreprocessor({
...
+
+            if (trimmedOpacity.endsWith('%')) {
+              return `color-mix(in srgb, ${trimmedColor} ${trimmedOpacity}, transparent)`;
+            } else {
+              return `color-mix(in srgb, ${trimmedColor} calc(${trimmedOpacity} * 100%), transparent)`;
+            }
+          }
+        );
+      }
+    } else if (token && !token.isSource) {
```
michal-sanoma: **suggestion:** Not sure if this is possible, bc I don't know if we always get the same structure from token studio but stack overflow risk is here! If any parent folder has a string property (like a description), this recursion will loop infinitely on the string's characters. Let's add `typeof token === 'object'` to the else if check to prevent this

scripts/build-themes.js
```diff
@@ -168,6 +225,28 @@ StyleDictionary.registerTransform({
...
+// Transform sizes to px if they don't have a unit
+StyleDictionary.registerTransform({
+  name: 'sl/size/css/size',
+  type: 'value',
+  transitive: true,
+  filter: token => token.$type === 'size' || token.$type === 'space',
+  transform: token => {
+    const value = token.$value;
+
+    if (typeof value === 'string') {
```
michal-sanoma, line 237: **suggestion:** we check if value is a string to append px. Just to be safe, what if a token value is defined as a number? It will currently bypass this and return a unitless CSS value, which is invalid. Maybe we should defensively handle the number type as well?

Not sure if we can get a number here though

packages/tokens/src/tokens/teacher-assistant/base-new.json
Diaan: Yeah, that is done in a different branch.
This was a separate fix, one that has more priority than the teacher assistent theme

scripts/build-themes.js
Diaan: It looks like TokensStudio always exports a string. there are no numeric values in the token files.

scripts/import-icons.js
Diaan: I created a new `theme-icons.json` file, but only for magister, because they were the only one that used the overwrite function.
The tokens can't be used for icons in Figma, so they were there only for the webcomponents. It makes more sense for us to manage these overwrite icons.

michal-sanoma (approved):

a11ymiko (approved):

PR #3505, opened 2026-07-09
fix(storybook select): make sticky header background opaque in story

### LINKED ISSUE #3176: [Select] Sticky header in "Hide When Out Of View" story has transparent background which interferes other content

When user scroll down in story [Hide When Out Of View](https://storybook.sanomalearning.design/?path=/story/form-select--hide-when-out-of-view) transparent background allows text of sticky header to interfere with content that is in the layer below. 

### 🤔 Expected Behavior?

Text of sticky header should not interfere with other text on the page. 

### 😯 Current Behavior

https://github.com/user-attachments/assets/d5f16bfe-66f1-44b6-aa4b-a83d6bf08573

### 💁 Possible Solution

Changing sticky header's background from transparent to solid should fix the issue.

### 🖥️ Steps to Reproduce

1. Go to [Hide When Out Of View](https://storybook.sanomalearning.design/?path=/story/form-select--hide-when-out-of-view) story.
2. Scroll down
3. Check if text of sticky header can be displayed on top of other text

Chrome, Safari, Microsoft Edge, Firefox

### If other, please specify.

_No response_

macOS, Win11

### 👤 Your name

Kacper

### 🧢 Your product/team

SLDS

### 🎨 Your Theme(s)

Sanoma Learning

michal-sanoma:
### BEFORE

https://github.com/user-attachments/assets/0fd91d7b-8cf0-42a4-a1b5-81abe4cfc6c1

### AFTER 

https://github.com/user-attachments/assets/c0a44717-2919-4137-92cc-321b1adfec52

### DIFF
--- a/packages/components/select/src/select.stories.ts
+++ b/packages/components/select/src/select.stories.ts
@@ -487,7 +487,9 @@ export const HideWhenOutOfView: StoryObj = {
         }
 
         header {
-          background: var(--sl-color-background-subtle);
+          background:
+            linear-gradient(var(--sl-color-background-subtle), var(--sl-color-background-subtle)),
+            var(--sl-elevation-surface-base-default);
           position: sticky;
           top: 0;
           padding: 24px;


a11ymiko (approved):

packages/components/select/src/select.stories.ts
```diff
@@ -487,7 +487,9 @@ export const HideWhenOutOfView: StoryObj = {
         }
 
         header {
-          background: var(--sl-color-background-subtle);
+          background:
```
anna-lach, line 490: Maybe it would be better to just use a different color without transparency like `var(--sl-elevation-surface-raised-sunken)` and no linear-gradient will be needed in that case?

anna-lach (approved):

packages/components/select/src/select.stories.ts
michal-sanoma: this color is transparent in dark mode

https://github.com/user-attachments/assets/95749b85-4efb-4976-b2fb-954833dfb3c8

anna-lach: oh ok, so maybe it's worth to ask one of designers which one to use?

michal-sanoma: Is it looking bad now? I'm not sure if I see any benefit in changing it

michal-sanoma: discussed on slack

PR #3526, opened 2026-07-15
feat(text area): add possibility to show or hide the character count indicator

### LINKED ISSUE #3100: [Text area] Add possibility to show or hide the current and total character count indicator

Based on NPM's needs we need to add a possibility to show or hide in the text area the current and total character count indicator.

Image

First we need to make some UI research and choose the best option where to place the counter.

### 👤 Your name

SLDS Team

### 🧢 Your Product/Team

SLDS

anna-lach:


jpzwarte:
@anna-lach Minor comment: the height of the "show count" text is smaller than the validation (probably because of the triangle icon). So when it switches between the two, the layout changes (report validity button moves up/down).

a11ymiko (approved):

packages/components/text-area/src/text-area.ts
```diff
@@ -210,9 +354,277 @@ export class TextArea extends ObserveAttributesMixin(
...
+        return;
+      }
+
+      const showCountLimit = this.#getShowCountLimit();
+
+      if (showCountLimit === undefined) {
+        return;
+      }
+
+      const over = this.value.length - showCountLimit;
```
michal-sanoma: **question:** what happens if a CFA sets the value to `null` or `undefined` (for example, during a form reset)? Won't accessing `this.value.length` directly in `#getCountText()`, `#getCountState()`, and `#syncCountValidity()` throw a TypeError at runtime?

packages/components/text-area/src/text-area.ts
```diff
@@ -210,9 +354,277 @@ export class TextArea extends ObserveAttributesMixin(
...
   }
 
+  /**
+   * Sets or clears the character count custom validity error. Only clears the error if we set it
+   * ourselves, so we never accidentally overwrite a custom error set by the user.
+   */
+  override updateInternalValidity(): void {
+    if (this.#isOverLimitState) {
+      // Preserve consumer-provided custom errors when showCount did not set the current one.
+      if (this.textarea.validity.customError && !this.#countValiditySet) {
```
michal-sanoma, line 376: **question:** should we override `setCustomValidity` in this component to reset `this.#countValiditySet` to `false` when a custom error is set externally, similar to how it is implemented in `NumberField`?

michal-sanoma (approved):

PR #3530, opened 2026-07-17
fix(select): handle non-document roots when cloning selected content

### LINKED ISSUE #3528: [sl-select] rootNode.importNode is not a function
### 😯 Current Behavior

In select.js → renderSelectedContent:

```
const slotNodes = this.selectedOption.renderRoot.querySelector('slot')?.assignedNodes() ?? [];
slotNodes.forEach((node) => {
  const rootNode = node.getRootNode();
  clones.push(rootNode.importNode(node, true)); // ❌ assumes Document
});
```
`Node.getRootNode()` is not always a `Document`. During framework remount / light-DOM churn it can be a `DocumentFragment` (or another non-Document root). Only `Document` implements importNode.

Suggestion of a possible solution:

Prefer a root that always has importNode, or fall back to `cloneNode`:

```
slotNodes.forEach((node) => {
  const root = node.getRootNode();
  const doc = root instanceof Document ? root : document;
  clones.push(doc.importNode(node, true));
  // or simply: clones.push(node.cloneNode(true));
});
```

`cloneNode(true)` is enough for copying option label content into the selected-content slot and avoids depending on `getRootNode()` being a Document.

### 🖥️ Steps to Reproduce

When `sl-select [value]` is set while Angular is projecting / updating `sl-option` light DOM (e.g. navigating between sections that share the same select instance).

Snippet of similar code where the issue occurs:

```
<!-- Angular template -->
<sl-select [value]="sectionTypeId" [clearable]="true">
  @for (type of sectionTypes; track type.id) {
    <sl-option [value]="type.id">{{ type.displayName }}</sl-option>
  }
</sl-select>
```

```
// Rapid value change while options are still settling
sectionTypeId = 'type-a';
// …navigate to another section…
sectionTypeId = 'type-b'; // triggers renderSelectedContent while option nodes may not be under Document
```

Observed: console throws `rootNode.importNode is not a function` during navigation.

No

_No response_

_No response_

### ⚖️ Impact

### Framework

Angular

All

### If other, please specify.

_No response_

_No response_

### Component version

2.1.3 - 2.1.5

### Theme version

_No response_

### 👤 Your name

Jorge Villegas

### 🧢 Your product/team

ETSL

### 🎨 Your Theme(s)

_No response_

michal-sanoma:
## Summary

Fixes a crash in `sl-select` when rendering selected option content while projected option light DOM is being updated.

`renderSelectedContent()` assumed that `node.getRootNode()` always returns a `Document` and called `importNode()` on it. During framework-driven light DOM churn, such as Angular remounting or updating projected `sl-option` content, the root can be a `DocumentFragment`, which does not implement `importNode()`.

The fix keeps the scoped custom element registry behavior when available, but falls back to the node’s owner document when the current root cannot import nodes.

## Changes

- Use root-level `importNode()` only when the current root supports it.
- Fall back to `ownerDocument.importNode()` for non-document roots.
- Add a regression test for selected content rendering when a slotted node root is not a document.

### BEFORE

https://github.com/user-attachments/assets/fd630277-a117-48ba-985d-a119c9865a40

### AFTER 

https://github.com/user-attachments/assets/fe35fc62-797a-445c-96f3-08c4640c5a1a

### DIFF
--- a/packages/components/select/src/select.ts
+++ b/packages/components/select/src/select.ts
@@ -445,14 +445,14 @@ export class Select<T = any> extends ObserveAttributesMixin(
         const scopedImportNode = (rootNode as Partial<Pick<Document, 'importNode'>>).importNode;
         const ownerDocument = node.ownerDocument ?? document;
 
+        // Use a scoped importNode() implementation when available so cloned custom
+        // elements are upgraded; otherwise fall back to the owner document for
+        // non-document roots such as DocumentFragment.
         const clone =
           typeof scopedImportNode === 'function'
             ? scopedImportNode.call(rootNode, node, true)
             : ownerDocument.importNode(node, true);
 
-        // Unlike node.cloneNode(), importNode() is implemented in the
-        // scoped custom element registry polyfill, so it will upgrade
-        // the cloned node if it's a custom element.
         clones.push(clone);
       });
 
--- a/packages/components/select/src/select.ts
+++ b/packages/components/select/src/select.ts
@@ -442,11 +442,17 @@ export class Select<T = any> extends ObserveAttributesMixin(
 
       slotNodes.forEach(node => {
         const rootNode = node.getRootNode();
+        const scopedImportNode = (rootNode as Partial<Pick<Document, 'importNode'>>).importNode;
+        const ownerDocument = node.ownerDocument ?? document;
+        const importNode =
+          typeof scopedImportNode === 'function'
+            ? scopedImportNode.bind(rootNode)
+            : ownerDocument.importNode.bind(ownerDocument);
 
         // Unlike node.cloneNode(), importNode() is implemented in the
         // scoped custom element registry polyfill, so it will upgrade
         // the cloned node if it's a custom element.
-        clones.push((rootNode as Document).importNode(node, true));
+        clones.push(importNode(node, true));
       });
 
       container.replaceChildren(...clones);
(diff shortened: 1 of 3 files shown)


packages/components/select/src/select.ts
```diff
@@ -442,11 +442,17 @@ export class Select<T = any> extends ObserveAttributesMixin(
 
       slotNodes.forEach(node => {
         const rootNode = node.getRootNode();
+        const scopedImportNode = (rootNode as Partial<Pick<Document, 'importNode'>>).importNode;
+        const ownerDocument = node.ownerDocument ?? document;
+        const importNode =
+          typeof scopedImportNode === 'function'
+            ? scopedImportNode.bind(rootNode)
+            : ownerDocument.importNode.bind(ownerDocument);
```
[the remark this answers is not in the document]
michal-sanoma: added

a11ymiko (approved):

packages/components/select/src/select.ts
```diff
@@ -442,11 +442,18 @@ export class Select<T = any> extends ObserveAttributesMixin(
...
+        const ownerDocument = node.ownerDocument ?? document;
+
+        const clone =
+          typeof scopedImportNode === 'function'
+            ? scopedImportNode.call(rootNode, node, true)
+            : ownerDocument.importNode(node, true);
 
         // Unlike node.cloneNode(), importNode() is implemented in the
         // scoped custom element registry polyfill, so it will upgrade
         // the cloned node if it's a custom element.
```
Diaan: this comment is not relevant here anymore. Maybe move it above the declaration of `clone` and adjust it a bit so it explains the new situation.

Diaan (approved):

packages/components/select/src/select.ts
michal-sanoma: good idea, aligned

PR #3542, opened 2026-07-22
feat(tree): add isSelectable for mixed-selectable nodes

### LINKED ISSUE #3541: [Tree] Missing `isSelectable` function from `TreeDataSourceMapping`

Currently it is not possible to have a tree where the nodes are selectable, but the leafs aren't. Example:

```
📚 Curriculum                              [selectable]
├── 🔷 Mathematics                         [selectable]
│   ├── 🔹 Algebra I                       [selectable]
│   │   ├── 📄 Linear Equations            [leaf — not selectable]
│   │   ├── 📄 Quadratic Functions         [leaf — not selectable]
│   │   └── 📄 Systems of Equations        [leaf — not selectable]
│   └── 🔹 Geometry                        [selectable]
│       ├── 📄 Triangles & Angles          [leaf — not selectable]
│       └── 📄 Circle Theorems             [leaf — not selectable]
├── 🔷 Science                             [selectable]
│   ├── 🔹 Biology                         [selectable]
│   │   ├── 📄 Cell Structure              [leaf — not selectable]
│   │   └── 📄 Photosynthesis              [leaf — not selectable]
│   └── 🔹 Chemistry                       [selectable]
│       ├── 📄 The Periodic Table          [leaf — not selectable]
│       └── 📄 Chemical Bonding            [leaf — not selectable]
└── 🔷 Languages                           [selectable]
    └── 🔹 Spanish                         [selectable]
        ├── 📄 Basic Greetings             [leaf — not selectable]
        └── 📄 Verb Conjugation            [leaf — not selectable]
```

We should add `isSelectable` and make the above possible.

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

### LINKED ISSUE #3545: [Tree] Icons are too close to the text

Image

Turns out there is a difference between Figma and the Web component. The latter only has a 2px gap between icon and label. Figma has ~5px. The gap in the code should be changed to 4px.

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

jpzwarte:
Closes #3541
Closes #3545 

Adds an optional `isSelectable` mapping function to the tree data sources, so a tree can have some nodes selectable and others not (e.g. selectable parent categories with non-selectable leaves).

### Data source
- `isSelectable?(item)` on `TreeDataSourceMapping`; defaults to `true` when omitted.
- `select`/`selectAll` skip non-selectable nodes; a parent's selected/indeterminate state is derived from its selectable children only. A node whose children are all non-selectable keeps an independent selection state.

### Component
- Non-selectable tree nodes render no checkbox, ignore click/Enter/Space selection, and omit `aria-selected`.
- `TreeNode.selectable` now defaults to falsy (opt-in), consistent with the component's other boolean properties; the tree always passes an explicit value derived from `isSelectable`.

### Stories & tests
- Replaced the old file-tree example data with a curriculum data set that better fits the design system domain.
- New `Selectable` story demonstrates selectable categories with non-selectable leaves, including per-leaf descriptions (shown in the row and exposed as an aria description).
- Added unit tests for the data source selection logic and the tree node rendering/interaction.

### DIFF
--- a/.changeset/happy-garlics-invent.md
+++ b/.changeset/happy-garlics-invent.md
@@ -1,0 +1,5 @@
+---
+'@sl-design-system/tree': patch
+---
+
+Increase gap between tree-node icon and label to match figma
--- a/.changeset/slow-needles-repeat.md
+++ b/.changeset/slow-needles-repeat.md
@@ -1,0 +1,5 @@
+---
+'@sl-design-system/tree': minor
+---
+
+Whether or not a tree node is selectable is now determined by the `isSelectable` function, which defaults to returning true. This means that all nodes are selectable by default, and only an explicit `isSelectable` returning false opts out.
--- a/packages/components/tree/src/flat-tree-data-source.spec.ts
+++ b/packages/components/tree/src/flat-tree-data-source.spec.ts
@@ -590,6 +590,26 @@ describe('FlatTreeDataSource', () => {
       expect(ds.items.every(n => n.selectable === true)).to.be.true;
     });
 
+    it('should mark all nodes as selectable by default in single-select mode', () => {
+      ds = new FlatTreeDataSource<TestItem>(items, {
+        getId: ({ id }) => id,
+        getLabel: ({ name }) => name,
+        getLevel: ({ level }) => level,
+        isExpandable: ({ expandable }) => expandable,
+        isExpanded: () => true,
+        isSelected: ({ id }) => id === 111
+      });
+      ds.update();
+
+      // All nodes selectable by default (no isSelectable, single-select mode).
+      expect(ds.items.every(n => n.selectable === true)).to.be.true;
+
+      // The initial isSelected state must be honored for selectable nodes.
+      const leaf = ds.items.find(n => n.id === 111)!;
+      expect(leaf.selected).to.be.true;
+      expect(ds.selection.has(leaf)).to.be.true;
+    });
+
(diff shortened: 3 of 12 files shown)


a11ymiko (approved):

michal-sanoma:
**Question:** I was playing with custom example in Storybook in order to understand the fix better

1. Is the scenario like this possible (I mean non-selectable parent with selectable children)? If yes, shouldn't "Grandparent" be selected when both children are selected?
2. Should we add "parent" to `dataSource.selection` array?


https://github.com/user-attachments/assets/e4b36f5c-4f28-4b83-a5d3-eb840971bd7e

jpzwarte:
> 1. Is the scenario like this possible

I can think of scenarios where:
- *only* the leafs are selectable
- *everything* but the leafs are selectable

Having nodes in between be not selectable feels extremely weird and bad from a UX perspective?

michal-sanoma (approved):

michal-sanoma:
> > 1. Is the scenario like this possible
> 
> I can think of scenarios where:
> 
> * _only_ the leafs are selectable
> * _everything_ but the leafs are selectable
> 
> Having nodes in between be not selectable feels extremely weird and bad from a UX perspective?

Just a note for my future self. We decided on a daily discussion that this example from video with non-selectable node between two selectable ones is not a realistic/needed one for our CFAs

PR #3544, opened 2026-07-23
feat: add rules to ensure all form controls have accessible labels

### LINKED ISSUE #2003: [sl-text-field] ARIA label not set if not used in sl-form

`aria-describedby` label is not set if `<sl-text-field>` is used without an enclosing `<sl-form>`

### 🤔 Expected Behavior?

aria labels should be set by the developer?

### 😯 Current Behavior

-

### 💁 Possible Solution

- is it possible to create an ESlint rule to inform the developer of a missing ARIA label it this situation?

- update documentation at https://sanomalearning.design/categories/components/button/accessibility/#wai-aria

### 🖥️ Steps to Reproduce

-

Chrome

### If other, please specify.

_No response_

WSL

### 👤 Your name

Frans de Boer

### 🧢 Your product/team

Magister

### 🎨 Your Theme(s)

Magister

### LINKED ISSUE #2536: [Eslint Plugin] Create ES lint rule that checks if your form control has an accessible name
### ✅ Tasks

- [ ] Create ES lint rule that checks if your form control has an accessible name.
- [ ] Add 'Support' section on documentation website underneath 'Utilities' Section
- [ ] Create documentation for Eslint including this new feature

### 👤 Your name

DR

### 🧢 Your Product/Team

_No response_

anna-lach:
Fixes:
- #2536 
- #2003


a11ymiko (approved):

packages/components/checkbox/src/checkbox-group.stories.ts
```diff
@@ -177,7 +178,7 @@ export const CustomValidity: Story = {
       };
 
       return html`
-        <sl-checkbox-group @sl-validate=${onValidate} required>
+        <sl-checkbox-group @sl-validate=${onValidate} aria-label="Pick an option" required>
```
Diaan: this is not really a good example of a label. Change it into something an option can be the answer for. Something like "Your favorite number"

packages/components/checkbox/src/checkbox-group.stories.ts
```diff
@@ -207,7 +208,7 @@ export const CustomAsyncValidity: Story = {
...
```
Diaan: same

packages/components/grid/src/filter.ts
```diff
@@ -125,13 +125,16 @@ export class GridFilter<T = any> extends ScopedElementsMixin(LitElement) {
...
+
       return html`
         <sl-select
           @sl-change=${this.#onSelectChange}
           @sl-clear=${this.#onClear}
-          .placeholder=${msg(str`Filter by ${this.#getFilterHeaderValue()}`, {
-            id: 'sl.grid.filterByValue'
-          })}
+          aria-label=${filterLabel}
+          .placeholder=${filterLabel}
```
Diaan, line 137: is having the same text for label and placeholder a good idea?
@a11ymiko what is your view on what is a good placeholder text?

packages/components/grid/src/group-header.ts
```diff
@@ -68,6 +68,7 @@ export class GridGroupHeader extends ScopedElementsMixin(LitElement) {
             <div part="checkbox">
               <sl-checkbox
                 @sl-change=${this.#onChange}
+                aria-label=${msg('Select group', { id: 'sl.grid.selectGroup' })}
```
Diaan: not a good label text. (see also  comment on checkbox-group.stories.ts)

Diaan (approved):

packages/components/grid/src/filter.ts
a11ymiko: For a11y the best placeholder text is none tbh 😅 I think that may be good to include in our documentation that we don't recommend adding placeholders and why it's not a good idea. 

But we still have to include it in DS and then the best placeholder text is: "Example: EXAMPLE ANSWER" or slightly less accessible: "e.g. EXAMPLE ANSWER". This is information that is not crucial, so it can disappear without harm, it's different from the label, and 'Example'/'e.g.' text at its beginning makes it less prone to be seen as already populated with valid data.

Using text 'Example' instead of 'e.g.' makes it more accessible for people with cognitive disabilities and people for whom english is not their mother thong. This should also be translated to other languages as the whole word and not an abbreviation.

packages/components/grid/src/group-header.ts
a11ymiko: If `id.grid.selectGroup` is responsible for spawning the name of a group into that label then just small change is needed: not "Select group..." but "Group, { id: 'sl.grid.selectGroup' }" will be good, because user will get information that this selects at least one or more items in a grid (because it's a group), will get information about unique name of the group, and still will be informed that this control is used to 'select' because of it's role, so no need to repeat it in the label text with word "Select".

tools/eslint-plugin-slds/lib/label-rule-helpers.js
```diff
@@ -0,0 +1,436 @@
...
+  if (!expression) {
+    return names;
+  }
+
+  switch (expression.type) {
+    case 'Identifier':
+      names.add(expression.name);
+      break;
+    case 'CallExpression':
+      if (expression.callee.type === 'Identifier') {
```
michal-sanoma, line 65: **question:** will this case cover template helpers like `${this.renderInput()}`?

tools/eslint-plugin-slds/lib/label-rule-helpers.js
```diff
@@ -0,0 +1,436 @@
...
+
+  return ids;
+};
+
+export const hasMeaningfulContent = (node, analyzer, sourceCode) => {
+  if (node.type === 'text') {
+    return node.data.trim() !== '';
+  }
+
+  if (node.type !== 'tag') {
```
michal-sanoma, line 297: **question:** should we also create a separate check for comments?

tools/eslint-plugin-slds/lib/index.js
```diff
@@ -1,23 +1,62 @@
 import { buttonHasLabel } from './rules/button-has-label.js';
+import { checkboxGroupHasLabel } from './rules/checkbox-group-has-label.js';
```
michal-sanoma, line 2: **question:** do I get it right that these rules will run also for all `**/*.spec.ts` files? If yes, do we want to lint them and add a lot of mock `aria-labels` there?

tools/eslint-plugin-slds/lib/rules/combobox-has-label.js
```diff
@@ -0,0 +1,121 @@
...
+  if (node.type === 'Literal' && typeof node.value === 'string') {
+    return node.value;
+  }
+
+  if (node.type === 'TemplateLiteral') {
+    // Skip template literals with expressions/interpolations (e.g., `<sl-combobox aria-label="${label}">`)
+    // because dynamic attribute values might provide the required label.
+    // Analyzing only the static quasis would lose the interpolated parts and create false positives.
+    if (node.expressions.length > 0) {
+      return null;
```
michal-sanoma: **suggestion:** I see that we accepted here Copilot's fix but it's probably too broad

We can have sth like this which passes validation but should fail:
```
element.innerHTML = `
  <h3>${unrelatedTitle}</h3>
  <sl-combobox></sl-combobox> 
`;
```

since `${unrelatedTitle}` is present, `getRawTemplateContent` returns `null`, and the missing label on `<sl-combobox>` is never flagged

tools/eslint-plugin-slds/lib/label-rule-helpers.js
anna-lach: It doesn't cover such cases, but I'm not sure if it's really necessary.

tools/eslint-plugin-slds/lib/index.js
anna-lach: Good question 🙂 I'll turn them off for `spec.ts` files.

tools/eslint-plugin-slds/lib/rules/combobox-has-label.js
anna-lach: Good catch!

packages/components/grid/src/filter.ts
anna-lach: Discussed with Kacper - we will keep both `aria-label` and `placeholder` the same in this case. Placeholder can be helpful for sighted users. It can also happen that `placeholder` will not be read by screen readers, so better to have the information sometimes read twice than nothing.

michal-sanoma (approved):

PR #3548, opened 2026-07-23
feat(select): add `fill` property

### LINKED ISSUE #3519: [Select] Add ghost fill
**Ghost type**
- [ ] Add `fill` property with `outline` as the default and `ghost` as an option

**To do**
- [x] create ghost type select in figma (use copy of current select component)
- [x] publish in figma
- [x] discuss dev during refinement

jpzwarte:
Closes #3519 

This adds a new `fill` property to `<sl-select>` with values `"outline"` (default) and `"ghost"`. When ghost is set, the border becomes transparent.

### DIFF
--- a/.changeset/two-dots-happen.md
+++ b/.changeset/two-dots-happen.md
@@ -1,0 +1,5 @@
+---
+'@sl-design-system/select': minor
+---
+
+Add a fill property and values "outline" (default) and "ghost"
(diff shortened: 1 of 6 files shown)


a11ymiko (commented):
Fill Ghost is not working in stories: 
1. Embedded Components
2. Custom styling
3. Option styling
4. No Visible Label
5. Word break
6. Display Inline Block
7. Custom Validity
8. Custom Async Validity
9. Hide When Out of View

In this stories user can choose `fill: ghost` using Storybook controls but that does not change how select is being displayed.

a11ymiko (approved):

michal-sanoma (approved):

PR #3560, opened 2026-07-29
feat(grid): Improve drag and drop behavior for grouping, improve `ArrayListDataSource` item order handling

### LINKED ISSUE #2875: [Grid] Grid exmple is missing in Angular storybook

We need to add an example (maybe basic at the beginning) in the Angular storybook with grid wrapper.

### 👤 Your name

Anna Sobczak

### 🧢 Your Product/Team

SLDS

anna-lach:
This pull request adds Storybook documentation and examples for the **Angular** `Grid` wrapper. It also introduces documentation files and minor improvements to Storybook preview behavior.


a11ymiko:
Once opened in isolation mode [Angular > Grid > Basic > Menu Button](https://storybook-3560.d3c5h2uor7jlkj.amplifyapp.com/angular/iframe.html?id=wrappers-grid-basics--menu-button-story&viewMode=story&refId=angular) story is being having strange horizontal scroll and its menu buttons are rendered partially out of view. 

For [Grid > Basic > Menu Button](https://storybook-3560.d3c5h2uor7jlkj.amplifyapp.com/iframe.html?id=grid-basics--menu-button&viewMode=story) there is no similar issue.

Screenshot was captured on Chrome on macBook screen (14-inch, 3024px × 1964px), but issue looks the same for other browsers:
image

a11ymiko (commented):
[Angular > Grid > Drag and Drop > On Top story](https://storybook-3560.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/angular_wrappers-grid-drag-and-drop--on-top) does not allow user to reorganize the grid. Rows can be dragged but even when they are dropped on spots that are highlighted it has no effect on final grid.
The same issue is present on our production storybook: [Grid > Drag and Drop > On Top story](https://storybook.sanomalearning.design/?path=/story/grid-drag-and-drop--on-top)

https://github.com/user-attachments/assets/c89fccfd-2fcc-40c3-8e02-9bdc386bcb8d

a11ymiko (commented):
Checkboxes for selecting whole group in [Angular > Grid > Drag and Drop > Grouping](https://storybook-3560.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/angular_wrappers-grid-drag-and-drop--grouping) should have labels that are the name of a group with word "group" added at the end of a label e.g. "Allergist group" or "Allergist (1) group" - option without number of elements in the group is better as a label, but may be harder to add I believe?

This way user will get information that this checkbox is used to select the whole group and gets info about group name.

Image

a11ymiko (commented):
In Safari [Angular > Grid > Scrolling > Vertical](https://storybook-3560.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/angular_wrappers-grid-scrolling--vertical) and [Angular > Grid > Scrolling > Both](https://storybook-3560.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/angular_wrappers-grid-scrolling--both) stories have longer scroll than necessary and user may scroll down with hope to find some more components there but it's empty.

Scroll should be restricted only to the height of grid.

https://github.com/user-attachments/assets/ef9229d3-3eb8-4c02-8a72-2468ae4f95d0

a11ymiko (commented):
Drag and drop feature does not work in [Angular > Grid > Drag and drop > Groping](https://storybook-3560.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/angular_wrappers-grid-drag-and-drop--grouping) story. User can drag each row with person's data but dropping it inside or outside of it's group has no effect. 
Groups cannot be dragged and dropped, but I assume it's right?

https://github.com/user-attachments/assets/395b0911-9412-45ee-a355-5dd461ddc248

anna-lach:
> [Angular > Grid > Drag and Drop > On Top story](https://storybook-3560.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/angular_wrappers-grid-drag-and-drop--on-top) does not allow user to reorganize the grid. Rows can be dragged but even when they are dropped on spots that are highlighted it has no effect on final grid. The same issue is present on our production storybook: [Grid > Drag and Drop > On Top story](https://storybook.sanomalearning.design/?path=/story/grid-drag-and-drop--on-top)
> 
>  Screen.Recording.2026-08-04.at.16.24.46.mov

Probably it would be better to fix it in a separate story, it looks like the behaviour doesn't work properly.

anna-lach:
> > [Angular > Grid > Drag and Drop > On Top story](https://storybook-3560.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/angular_wrappers-grid-drag-and-drop--on-top) does not allow user to reorganize the grid. Rows can be dragged but even when they are dropped on spots that are highlighted it has no effect on final grid. The same issue is present on our production storybook: [Grid > Drag and Drop > On Top story](https://storybook.sanomalearning.design/?path=/story/grid-drag-and-drop--on-top)
> > Screen.Recording.2026-08-04.at.16.24.46.mov
> 
> Probably it would be better to fix it in a separate story, it looks like the behaviour doesn't work properly.

New issue created: https://github.com/sl-design-system/components/issues/3584

jpzwarte:
The title of the PR is what will be visible as the commit msg after merging the PR. Can you change the title to reflect the new/improved features wrt grid?

a11ymiko (approved):

packages/components/data-source/src/array-list-data-source.ts
```diff
@@ -61,16 +65,77 @@ export class ArrayListDataSource<T = any> extends ListDataSource<T> {
...
+    const targetGroupId = isListDataSourceGroupItem(relativeItem)
+      ? relativeItem.id
+      : relativeItem.groupId;
+
+    if (targetGroupId === undefined || targetGroupId === item.id) {
+      return;
+    }
+
+    const groupIds = this.items
+        .filter((listItem): listItem is ListDataSourceGroupItem<T> =>
```
michal-sanoma: **issue:** didn't check yet in SB example but isn't this going to cause data loss when filtering is active?

If a filter is active, `this.items` will only contain visible groups. When we rebuild `this.#mappedItems` by mapping over `orderedGroupIds` at the end of the method, won't all items in the filtered-out groups be permanently deleted from the source array?

michal-sanoma: yeah, seems like a problem

https://github.com/user-attachments/assets/76b0136d-bcbe-43ee-ac8e-eefd94ee00a1

anna-lach: Good catch! Should be fixed now.

michal-sanoma (approved):

PR #3570, opened 2026-08-03
fix(grid): prevent rows shrinking on ellipsized text hover

### LINKED ISSUE #3555: [Grid] Rows shrinking on hover
### 😯 Current Behavior

When the field content doesn't fit the column of the grid then the row height is shrinking on hover

The row height stays the same

### 🖥️ Steps to Reproduce

Image

No

_No response_

_No response_

### ⚖️ Impact

### Framework

Angular

Chrome

### If other, please specify.

_No response_

_No response_

### Component version

0.7.4

### Theme version

_No response_

### 👤 Your name

Szymon Ambroziak

### 🧢 Your product/team

MAX

### 🎨 Your Theme(s)

_No response_

michal-sanoma:
## Summary

Fix grid rows shrinking when hovering ellipsized cell content.

`sl-ellipsize-text` lazily adds an `sl-tooltip` when truncated text is hovered or focused. The grid cell selector treated that tooltip as custom cell content and removed the cell's block padding, which could reduce the entire row height.

## Changes

- Exclude `sl-tooltip` from the grid cell selector that removes block padding for custom content.
- Add a regression test to ensure ellipsized grid cells keep their padding when a tooltip is added

### BEFORE

https://github.com/user-attachments/assets/783462a2-68c8-425d-b67d-1755496cabd2

### AFTER

https://github.com/user-attachments/assets/7c788fef-7855-43d9-aaee-9fd2a9d43ff8

### DIFF
--- a/packages/components/grid/src/stories/basics.stories.ts
+++ b/packages/components/grid/src/stories/basics.stories.ts
@@ -112,7 +112,7 @@ export const EllipsizeTextWithCustomContent: Story = {
     const files: ResourceFile[] = [
       {
         empty: null,
-        name: 'Leerdolenoverzicht Of Course havo-vwo bovenbouw very very long filename.pdf',
+        name: 'Some name which has extremely long and very very long filename.pdf',
         type: 'pdf'
       },
       {
--- a/packages/components/grid/src/stories/basics.stories.ts
+++ b/packages/components/grid/src/stories/basics.stories.ts
@@ -10,7 +10,7 @@ import { Icon } from '@sl-design-system/icon';
 import { MenuButton as MenuButtonComponent, MenuItem } from '@sl-design-system/menu';
 import { Tooltip } from '@sl-design-system/tooltip';
 import { type Meta, type StoryObj } from '@storybook/web-components-vite';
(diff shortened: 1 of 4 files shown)


a11ymiko (approved):

packages/components/grid/src/stories/basics.stories.ts
```diff
@@ -101,6 +107,136 @@ export const EllipsizeText: Story = {
   `
 };
 
+export const EllipsizeTextWithCustomContent: Story = {
+  render: () => {
+    const files: ResourceFile[] = [
+      {
+        empty: null,
+        name: 'Leerdolenoverzicht Of Course havo-vwo bovenbouw very very long filename.pdf',
```
anna-lach: Maybe it would be better to keep those names in example in English to have it consistent with other stories? But that's a small thing. :)

anna-lach (approved):

packages/components/grid/src/stories/basics.stories.ts
michal-sanoma: sure, fixed :)

PR #3571, opened 2026-08-04
chore: replace package build setup with tsdown

### LINKED ISSUE #3182: [Maintenance] Replace esbuild build scripts with tsdown
## Summary

Replace the custom `scripts/build-packages.js` (and potentially `scripts/build-styles.js`) with [tsdown](https://tsdown.dev/), a modern TypeScript bundler built on top of Rolldown/oxc that natively supports monorepos via its experimental `workspace` option.

## Motivation

The current build pipeline has grown into a collection of separate tools and scripts that each solve one part of the problem — and they add up:

- **esbuild** (`build-packages.js`) for JS transpilation
- **tsc** (`tsc --build tsconfig.all.json`) for `.d.ts` declaration files
- **sass** + **stylelint** + custom wrapping (`build-styles.js`) for SCSS → Lit CSS templates
- **Wireit** to orchestrate and cache all of the above

The two most impactful pain points:

**1. The `tsc --build tsconfig.all.json` step is slow and blocks everything**

Every local build and every Storybook startup requires a full type declaration pass across all 50+ packages via TypeScript project references. This is the bottleneck in both `yarn build` and the `start` task. tsdown's `dts: true` generates declarations per-package as part of the same pass as JS compilation — and with `isolatedDeclarations: true`, it can use Oxc for this, which is significantly faster.

**2. Storybook requires a separate build step today — it shouldn't**

Currently `yarn start` (Storybook dev) depends on `build:components` and `build:types` completing first, because Storybook needs compiled `.js` files. But `@storybook/web-components-vite` processes TypeScript directly via Vite — it doesn't need pre-compiled output at all. The only reason those steps are prerequisites is that esbuild/tsc are also the mechanism for producing the files. With tsdown handling distribution builds only, the Storybook dev workflow becomes: run the SCSS pipeline once, then launch Storybook — no separate JS or type build step needed.

**3. Unified tooling**

Replacing two tools (esbuild + tsc) with one (tsdown) simplifies the Wireit task graph, reduces the number of moving parts, and gives a single place to configure build behavior across all packages.

## Proposed change

Replace `build-packages.js` with a root-level `tsdown.config.js`. A PoC using the experimental `workspace` option already works:

```js
import { defineConfig } from &#39;tsdown&#39;

export default defineConfig({
  workspace: &#39;packages/*&#39;,
  entry: [
    &#39;src/index.ts&#39;,
    // { &#39;register.js&#39;: &#39;src/register.ts&#39; }
  ],

  inputOptions(inputOptions, format) {
    if (inputOptions.cwd.endsWith(&#39;bar&#39;)) {
      inputOptions.input = { index: &#39;src/index.ts&#39; };
    }
    return inputOptions;
  },

  exports: {
    devExports: true,
    customExports(exports) {
      exports[&#39;./register.js&#39;] = &#39;./src/register.ts&#39;;
      return exports;
    }
  },
  platform: &#39;browser&#39;,
  unbundle: true,
  sourcemap: true,
  dts: true,

  deps: {
    skipNodeModulesBundle: true,
  },
})
```

tsdown expects entry points under `src/`, so `index.ts` and `register.ts` need to move from the package root into `src/` as a prerequisite.

We can replace the current `.scss` → `.scss.ts` → `.scss.js` workflow with `import styles from &#39;./foo.css&#39; with { type: &#39;css&#39; };` and have that work with vite and tsdown through plugins. This has already been implemented in the `docs/website-v2` branch for the documentation components.

Benefits over the current esbuild script:
- **Built-in DTS generation** (`dts: true`) — could eliminate the separate `tsc --build` step for declarations
- **Native workspace support** — no need for a glob + esbuild loop per package type
- **`exports` management** — tsdown can manage `package.json` exports fields (including `./register.js`) automatically, reducing per-package boilerplate
- **`devExports`** — auto-adds `./src/*` exports for dev/Storybook consumption without manual config
- **Rolldown/oxc speed** — significantly faster than esbuild for larger codebases

## Open questions / tasks

- [ ] Move `index.ts` and `register.ts` from each package root into `src/` and update all internal imports accordingly
- [ ] Validate the `workspace` glob covers all package types (`packages/components/*`, `packages/locales`, `packages/themes`)
- [ ] Handle the `register.ts` entry point per package (currently excluded from the glob in `build-packages.js` via the negation pattern but published as `register.js`)
- [ ] Decide whether to keep `tsc --build` for type checking or rely solely on tsdown&#39;s `dts: true`
- [ ] Check whether wireit incremental build (`files`/`output` tracking) can be preserved or replaced with tsdown&#39;s own watch/cache
- [ ] Test against all package types, including `packages/locales` (generated code) and `packages/themes`

### LINKED ISSUE #3569: [Angular] Upgrade to version 22

New major version is out (since June); we should upgrade.

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

jpzwarte:
Closes #3182 
Closes #3569 

Replaces the esbuild package build with [tsdown](https://tsdown.dev) and upgrades the Angular wrappers to Angular 22.

### Packaging

- Output moves from the package root to `dist/`; the packages are now resolved entirely through `exports`. `main`, `module` and `types` are dropped — `exports` already points at the JavaScript, and TypeScript picks up the adjacent `.d.ts` from there.
- `exports` and `publishConfig.exports` are **generated by tsdown** from its entry list — change the build config, not the exports map by hand.
- Sources (including `index.ts` / `register.ts`) move into `src/`. Styles move from `.scss` to `.css`, imported with `with { type: 'css' }`;

### Angular

Angular 21 → 22, builders moved to `@angular/build`, `^22.0.0` added to the peer range alongside 19–21.

### Fixes

- `sideEffects` still listed `register.js`, which matches neither `src/register.ts` nor `dist/register.js` under the new layout. Bundlers were silently tree-shaking away every `import '@sl-design-system/*/register.js'`, leaving the custom elements unregistered.
- Declare dependencies that were missing from `package.json` across 18 packages, plus missing peers (`@open-wc/scoped-elements`, `@lit/localize`, `lit`).
- Restore the `format-date/format.js`, `grid/utils.js` and `locales/*.xlf` subpath exports that the generated exports map had dropped.
- Add `isDevMode()` (`@sl-design-system/shared/dev-mode.js`), replacing the `/// <reference types="vite/client" />` in `icon` and `combobox` that leaked an empty `*.css` ambient module into every consumer's TypeScript program.
- Migrate `examples/lit` to `.css` as well, and teach the website's esbuild config to handle `with { type: 'css' }`.


a11ymiko (approved):

packages/components/accordion/tsconfig.json
```diff
@@ -1,8 +0,0 @@
-{
```
michal-sanoma, line 1: **issue:** we deleted all individual `tsconfig.json` files from each component directory but we still reference them in `tsconfig.all.json`

Image

.changeset/full-feet-sink.md
```diff
@@ -0,0 +1,62 @@
...
+'@sl-design-system/tooltip': patch
+'@sl-design-system/tree': patch
+'@sl-design-system/virtual-list': patch
+---
+
+Build the package with tsdown
+
+The build has moved from esbuild to [tsdown](https://tsdown.dev). The public API is unchanged, but the published layout is different: compiled output now lives in `dist/` instead of the package root, and the package is resolved entirely through `exports`. The `main`, `module` and `types` fields have been dropped, since `exports` already points at both the JavaScript and, alongside it, the type declarations.
+
+Bundlers and TypeScript setups that understand `exports` (`moduleResolution: bundler`, `node16` or `nodenext`) need no changes. The legacy `moduleResolution: node` is no longer supported.
```
michal-sanoma: **question:** Isn't this a "breaking change"?

website/esbuild.config.js
```diff
@@ -1,11 +1,45 @@
 import esbuild from 'esbuild';
 import gzipPlugin from '@luncheon/esbuild-plugin-gzip';
 import { minifyHTMLLiteralsPlugin } from 'esbuild-plugin-minify-html-literals';
+import { readFile } from 'node:fs/promises';
```
michal-sanoma, line 4: **question:** Do you also have failing `yarn start` command locally in `website`? I guess we need to improve TS support in this dir

Image

website/esbuild.config.js
```diff
@@ -1,11 +1,45 @@
...
+    build.onResolve({ filter: /\.css$/ }, args => {
+      if (args.with?.type !== 'css') {
+        return null;
+      }
+
+      return { path: resolvePath(args.resolveDir, args.path), namespace: 'css-sheet' };
+    });
+
+    build.onLoad({ filter: /.*/, namespace: 'css-sheet' }, async args => {
+      const css = await readFile(args.path, 'utf8');
```
michal-sanoma, line 29: **question:** there is this note in the description:

`Styles move from .scss to .css, imported with with { type: 'css' }; nesting is flattened at build time, so no browser-support change`

In which place we do this flattening? I have a feeling that we ship css as is, without any CSS processors 

Image

Image

vitest.config.ts
```diff
@@ -1,10 +1,18 @@
...
 import { defineConfig } from 'vitest/config';
 
 export default defineConfig({
+  plugins: [importCssSheet()],
   test: {
-    onConsoleLog: log => !(log.startsWith('Lit is in dev mode') || log === 'null'),
+    onConsoleLog: log => {
+      return !(
+        log === 'null' ||
+        log == '[Error: ResizeObserver loop completed with undelivered notifications.]' ||
```
michal-sanoma, line 12: **praise:** nice one!

package.json
```diff
@@ -402,40 +286,45 @@
...
+    "@storybook/addon-vitest": "^10.5.7",
+    "@storybook/web-components": "^10.5.7",
+    "@storybook/web-components-vite": "^10.5.7",
     "@types/chai-datetime": "^1.0.0",
     "@types/chai-dom": "^1.11.3",
-    "@types/sinon": "^21.0.1",
+    "@types/react": "^19.2.18",
+    "@types/sinon": "^22.0.0",
     "@types/sinon-chai": "^4.0.0",
+    "@typescript/native-preview": "7.0.0-dev.20260707.2",
```
michal-sanoma, line 300: **thought:** interesting how much it will speed up our builds

packages/components/accordion/tsconfig.json
jpzwarte: Deleted

.changeset/full-feet-sink.md
jpzwarte: Removed since it's not accurate anyway.

website/esbuild.config.js
jpzwarte: Fixed

website/esbuild.config.js
jpzwarte: Removed since it's not true :)

.changeset/full-feet-sink.md
michal-sanoma: I mean we don't support `moduleResolution: node` anymore and for some users it's a `major` change and we mark it as `patch`

Correct me if I'm wrong but I understand it this way that removal of these fields is related to this:
```
"main": "./index.js",
"module": "./index.js",
"types": "./index.d.ts",
```

and if somebody still wants to use it as before it's not possible bc afaik `moduleResolution: node` ignores `exports` field

website/esbuild.config.js
michal-sanoma: Please remove also from PR's description :)

michal-sanoma: One extra thing which comes to my mind. As I understand we did it before, I mean flattening. So if we change this now it's probably also a "breaking change". I now that support of this feature is wide across browsers but still

.changeset/full-feet-sink.md
jpzwarte: Yesterday i asked Claude this and it actually tested it using code. Turns out that, yes, you are right in that those fields are related to `moduleResolution: node`, other stuff made `node` module resolution break. So in practice, it didn't work.

website/esbuild.config.js
jpzwarte: Flattening was probably a side effect of the sass compiler, but looking at https://caniuse.com/css-nesting, i don't think we need to worry (or mention) it explicitly.

michal-sanoma (approved):

PR #3595, opened 2026-08-13
fix(toggle-group): use group aria role by default

### LINKED ISSUE #2912: [Toggle group] Use role="group" for a set off toggles

The Toggle group currently has a role="region". This is typically reserved for important sections of a website, typically consisting of a set of meaningful content, like other landmarks (header, main, navigation etc.). While not semantically totally wrong, using role="group" would be more appropriate for a set of toggles.

### 🤔 Expected Behavior?

Users understand that this is a set of interface elements.

### 😯 Current Behavior

Users might understand this as an major section of content.

### 💁 Possible Solution

Change the role="region" to role="group". Preferably add a aria-label to the example in the storybook.

### 🖥️ Steps to Reproduce

Screen reader:
1. Enter the group of toggles using a screen reader -> Screen reader identifies it as a region

Chrome

### If other, please specify.

_No response_

Windows 11

### 👤 Your name

Valtteri Järvinen

### 🧢 Your product/team

External

### 🎨 Your Theme(s)

Sanoma Learning

michal-sanoma:
## Summary

- Change the default semantic role of `sl-toggle-group` from `region` to `group`
- Preserve explicitly provided roles, such as `role="toolbar"`
- Add regression tests for the default role and explicit role handling

## Why

`role="region"` exposes the toggle group as a landmark, which is intended for important page sections. A toggle group is a set of related controls, so `role="group"` better communicates its purpose to assistive technologies

### BEFORE
Screenshot 2026-08-13 at 14 43 00

### AFTER
Screenshot 2026-08-13 at 14 43 44

### DIFF
--- a/packages/components/toggle-group/src/toggle-group.ts
+++ b/packages/components/toggle-group/src/toggle-group.ts
@@ -93,8 +93,9 @@ export class ToggleGroup extends LitElement {
   override connectedCallback(): void {
     super.connectedCallback();
 
-    // https://twitter.com/LeonieWatson/status/1545788775644667904
-    this.setAttribute('role', 'region');
+    if (!this.hasAttribute('role')) {
+      this.setAttribute('role', 'group');
+    }
   }
 
   override updated(changes: PropertyValues<this>): void {
--- a/packages/components/toggle-group/src/toggle-group.ts
+++ b/packages/components/toggle-group/src/toggle-group.ts
@@ -50,6 +50,9 @@ export class ToggleGroup extends LitElement {
     isFocusableElement: (el: ToggleButton) => !el.disabled
   });
 
+  /** @internal */
+  readonly internals = this.attachInternals();
+
   /**
    * If set, will disable all buttons in the group.
    *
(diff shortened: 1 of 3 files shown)


packages/components/toggle-group/src/toggle-group.ts
```diff
@@ -93,8 +93,9 @@ export class ToggleGroup extends LitElement {
   override connectedCallback(): void {
     super.connectedCallback();
 
-    // https://twitter.com/LeonieWatson/status/1545788775644667904
-    this.setAttribute('role', 'region');
+    if (!this.hasAttribute('role')) {
+      this.setAttribute('role', 'group');
+    }
```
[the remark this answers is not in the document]
michal-sanoma: fixed

a11ymiko (approved):

packages/components/toggle-group/src/toggle-group.ts
```diff
@@ -93,8 +96,7 @@ export class ToggleGroup extends LitElement {
   override connectedCallback(): void {
     super.connectedCallback();
 
-    // https://twitter.com/LeonieWatson/status/1545788775644667904
-    this.setAttribute('role', 'region');
+    this.internals.role = 'group';
```
anna-lach, line 99: Do we really need ElementInternals for that?

michal-sanoma: yup, we need this. It's explained by copilot above

anna-lach (approved):

PR #3603, opened 2026-08-16
feat(popover): position with CSS anchor positioning

### LINKED ISSUE #3602: [Popover] Refactor to use CSS Anchor Positioning

- remove `anchor` property in favor of `popovertarget` or `command="show-popover"`
- remove `position` property in favor of plain CSS: `position-area: top;`
- remove `AnchorController` and replace with plain CSS

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

jpzwarte:
Closes #3602 

Replace the floating-ui based positioning with CSS anchor positioning. The browser now keeps the popover in place and flips it to the opposite side when there is not enough room, so the per-frame repositioning loop is gone.

The `position` property is reflected as an attribute so all twelve positions can be styled from CSS, and the offset, arrow padding and viewport margin moved from static JS properties to the `--_offset`, `--_arrow-padding` and `--_viewport-margin` custom properties. Room for the arrow is reserved next to the anchor, on top of the offset.

The popover also anchors itself to the element that invoked it, either via `popovertarget` or via the `source` passed to `showPopover()`/`togglePopover()`, and the anchor can now be set up entirely from CSS with `anchor-name` / `position-anchor`. Infotip uses that to anchor its popover to the trigger button in its own shadow root.

### DIFF
--- a/.changeset/olive-crabs-guess.md
+++ b/.changeset/olive-crabs-guess.md
@@ -1,0 +1,14 @@
+---
+'@sl-design-system/popover': minor
+---
+
+The popover now positions itself against the element that invoked it. When a popover is opened by
+an invoker — either via the `popovertarget` attribute or by passing a `source` to `showPopover()`
+or `togglePopover()` — that element becomes the popover's anchor:
+
+```js
+popover.showPopover({ source: button });
+```
+
+When there is no invoker, the anchor is left untouched, so popovers using the `anchor` attribute or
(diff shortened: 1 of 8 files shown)


jpzwarte:
Closed automatically when the branch was renamed from `feat/3602-popover-refactoring` to `feature/3602-popover-refactoring`. Continued in #3614.

Issue #2903, opened 2025-12-17
[Radio and checkbox] Description per item


jtalen:
### Provide a general summary of the feature here

I would like to have the opportunity to add extra (hint) text to each radio button list item. This way I can explain a bit more about every option the user has.

This is an example from Microsoft Power apps:

<img width="390" height="204" alt="Image" src="https://github.com/user-attachments/assets/d8e8ba9f-3fd7-4c2d-bb22-8569e9e38ef5" />

### 🤔 Expected Behavior?

N/A

### 😯 Current Behavior

N/A

### 🔦 Context

N/A

### 💻 Examples

Microsoft Power apps. For instance when adding a column to a table

<img width="1701" height="975" alt="Image" src="https://github.com/user-attachments/assets/1c26fdeb-ca4d-4694-8ecc-713d5e2ebece" />

### 👤 Your name

Jeremy Talen

### 🧢 Your Product/Team

Magister

jpzwarte:
This is the same as #3099. It should not be just for radio button, but for checkbox and switch as well.

DanielleRameau:
Design: https://www.figma.com/design/CHpKrPIdXdbV2u7X8vizKI/Components-2.0?node-id=17355-261973&t=NTpPm5h507CVLsJ6-4

PR #3604, opened 2026-08-16
feat(switch): refactor switch

### LINKED ISSUE #3415: [SWITCH] Tooltip does not work on disabled sl-switch
### 😯 Current Behavior

Tooltip directive does not work on a disabled `sl-switch `

should show tooltip if switch control has a disabled state.

### 🖥️ Steps to Reproduce

please use the story below to reproduce the problem

```typescript
export const Disabled: Story = {
  args: {
    example: ({ alignSelf, justifySelf, message }) => html`
      <div
        style=${styleMap({
          'align-self': alignSelf,
          display: 'inline-flex',
          gap: '1rem',
          'justify-self': justifySelf
        })}>
        <sl-switch ${tooltip(message)} disabled>Disabled switch</sl-switch>
        <sl-button ${tooltip(message)} aria-disabled="true">Disabled (ARIA only) button</sl-button>
      </div>
    `
  }
};
```

No

_No response_

see 🖥️ Steps to Reproduce*

### ⚖️ Impact

### Framework

Lit

All

### If other, please specify.

_No response_

_No response_

### Component version

1.1.8

### Theme version

_No response_

### 👤 Your name

Frans de Boer

### 🧢 Your product/team

Magister

### 🎨 Your Theme(s)

_No response_

jpzwarte:
## Summary

Refactors `<sl-switch>` so the switch itself is the form associated element, and adds a `tooltip` property so a disabled-looking switch can explain why it is disabled.

- The switch no longer adds an `<input>`, `<label>` and `<style>` to its light DOM. It uses `ElementInternals` for the form value and validity, and renders an `<input type="checkbox" role="switch">` in its shadow DOM as the control for keyboard and assistive technology, following the [WAI-ARIA Authoring Practices](https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-checkbox/).
- ARIA attributes are forwarded to that input by `ForwardAriaMixin`, the same as `<sl-button>` and `<sl-checkbox>`. This makes `aria-disabled` work: the switch looks disabled and cannot be toggled, but stays focusable so a tooltip can explain why. Fixes #3415.
- New `tooltip` property, the same way `<sl-button>` has one. Hovering the switch centers the tooltip on the switch as a whole; focusing the toggle centers it on the toggle.
- New `toggle(force?)` method to toggle the switch from JavaScript, since the switch no longer listens for clicks itself. It behaves like a user toggling it: emits `sl-change` and marks the switch dirty.
- New `description` slot, linked with `aria-describedby`, plus `checked` / `has-description` / `no-label` custom states and `container`, `description`, `label`, `tooltip` and `wrapper` CSS parts. Part of #2903
- Adds a `getSlottedText()` utility to `@sl-design-system/shared` for reading the text assigned to a slot.

### Labelling a control that lives in the shadow DOM

Because the input moved into the shadow DOM, the `<label for="...">` that `<sl-label>` renders no longer reaches it: label association binds to the host, and an IDREF cannot cross a shadow boundary. `ForwardAriaMixin` now forwards the `data-label-id` attribute `<sl-label>` sets on the control as if it were `aria-labelledby`, resolving it against the host's root and assigning it to the target as an element reference, which does cross the boundary.

The attribute is left on the host, since `<sl-label>` manages it. `forwardedElementsStorage` is now keyed by attribute name rather than by target property, so `aria-labelledby` and `data-label-id` both contribute to `ariaLabelledByElements` instead of overwriting each other.

### Breaking changes

See [the changeset](https://github.com/sl-design-system/components/blob/fix/3415-disabled-switch-tooltip-v2/.changeset/tame-spiders-invite.md) for the full list. In short: `input` now returns the shadow DOM input and is internal, the `input` slot is gone, `formControlElement` is the switch itself (so `<label for>` points at the `<sl-switch>` id), `checked` is no longer reflected — use `sl-switch:state(checked)` instead of `[checked]` — and `click()` no longer toggles the switch, use `toggle()` instead.


packages/components/switch/src/switch.ts
```diff
@@ -178,185 +198,236 @@ export class Switch<T = any> extends ObserveAttributesMixin(
...
           </div>
         </div>
+
+        <label id="toggle" part="toggle">
+          <input
+            aria-checked=${Boolean(this.checked).toString()}
+            aria-describedby=${describedBy || nothing}
+            aria-labelledby=${labelledBy || nothing}
+            .checked=${!!this.checked}
+            ?disabled=${this.disabled}
```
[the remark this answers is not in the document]
jpzwarte: I've created #3625 to track this.

a11ymiko (commented):
[Icon only switch](https://storybook-3604.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/form-switch--icons) should have invisible label added via `aria-label` or `aria-labelleby` should reference invisible text, not an empty wrapper. Example label text: 'Dark mode'.

a11ymiko (commented):
Is it possible to have tooltip displayed in different positions if user hover over label or switch? I mean if user hover over label then tooltip appears under/above middle of label and when user hover over switch it is placed above/under switch? 

Right now when user hovers over label or switch the tooltip appears in the middle of space taken by both label and switch which in storybook example means that it is very far from both label and switch. This may be problematic for magnifier users who are very likely to miss this tooltip. But when tooltip would appear closer to hovered label and switch then probability of showing tooltip inside magnified content is much higher. 

https://storybook-3604.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/form-switch--aria-disabled

Image

a11ymiko (commented):
Is the change in size of gap between label and switch deliberate? 

https://www.chromatic.com/test?appId=646dd42469efabe23410bb59&id=6a857224a811de8c1226e868 

Image

packages/components/shared/src/mixins/forward-aria-mixin.ts
```diff
@@ -197,17 +219,17 @@ export function ForwardAriaMixin<
...
               if (elementsProp?.endsWith('Elements')) {
                 const forwarded = forwardedElementsStorage.get(this),
-                  previous = forwarded?.get(elementsProp) ?? [],
+                  previous = forwarded?.get(name) ?? [],
                   current =
                     (target as unknown as Record<string, Element[] | null>)[elementsProp] ?? [],
                   remaining = current.filter(el => !previous.includes(el));
 
                 (target as unknown as Record<string, Element[] | null>)[elementsProp] =
                   remaining.length ? remaining : null;
```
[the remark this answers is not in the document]
jpzwarte: Fixed

packages/components/switch/src/switch.ts
```diff
@@ -178,185 +198,236 @@ export class Switch<T = any> extends ObserveAttributesMixin(
...
-    input.role = 'switch';
+  #onWrapperClick = (event: MouseEvent): void => {
+    event.preventDefault();
+    event.stopPropagation();
 
-    input.checked = !!this.checked;
-    input.setAttribute('aria-checked', this.checked ? 'true' : 'false');
+    const path = event.composedPath();
+    if (this.infotip && path.includes(this.infotip)) {
+      return;
```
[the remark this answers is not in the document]
jpzwarte: Fixed

jpzwarte:
@a11ymiko 
- icon only: the `<input>` has the tooltip as the label
- fixed: tooltips are now either anchored to the wrapper or the toggle
- fixed: there was unneeded whitespace due to a grid-template bug

a11ymiko (commented):
Could you change tooltip text in [Icons story](https://storybook-3604.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/form-switch--icons) to "Dark mode"?  

Then it would be more aligned with accessible real-world example because it'll be informing better about current state of dark mode. Information "Dark mode, on, switch" is more clear than "Click to toggle between light and dark mode., on, switch"

a11ymiko (approved):

packages/components/switch/src/switch.stories.ts
```diff
@@ -23,140 +30,168 @@ export default {
...
-      ?checked=${checked}
-      ?disabled=${disabled}
-      ?reverse=${reverse}
-      size=${ifDefined(size)}
-      .value=${value}>
-      ${text}
-    </sl-switch>
-  `
+    component: () => html`
+      <sl-switch
```
anna-lach, line 128: Isn't it weird that the tooltip anchor is the switch text or switch element? Probably there should be just one anchor? Now it looks a bit weird when it disappears and appears again?

https://github.com/user-attachments/assets/f0e3c5bd-cefd-4e05-ba88-50b4747d0895

packages/components/switch/src/switch.ts
```diff
@@ -136,34 +168,23 @@ export class Switch<T = any> extends ObserveAttributesMixin(
...
+    this.setFormControlElement(this);
+  }
+
+  override disconnectedCallback(): void {
+    this.#eventController.abort();
+
+    super.disconnectedCallback();
   }
 
   formAssociatedCallback(): void {
```
anna-lach, line 191: Probably should be marked as `@internal`? Now it's visible on the website (without description).

Image

packages/components/switch/src/switch.ts
anna-lach, line 196: Probably should be marked as `@internal`? Now it's visible on the website (without description).

Image

packages/components/switch/src/switch.ts
```diff
@@ -35,31 +33,40 @@ declare global {
...
 
-let nextUniqueId = 0;
-
 /**
  * A toggle switch.
  *
- * ```html
- * <sl-switch>Foo</sl-switch>
- * ```
+ * @customelement sl-switch
```
anna-lach, line 39: I wanted to take a look at `switch` example also in Angular storybook, but it looks like there is sth wrong? Stories aren't loading anymore?

Image

packages/components/switch/src/switch.stories.ts
a11ymiko: @anna-lach I requested change from only one anchor (in the middle between label and switch) to how it behaves currently 😅 here is my reasoning: https://github.com/sl-design-system/components/pull/3604#pullrequestreview-4971668979

anna-lach: Oh ok, I missed that. 😆

jpzwarte: This is what Kacper and i decided on :)

packages/components/switch/src/switch.ts
jpzwarte: Hmm, i suppose. Proper solution is to automatically exclude them from the docs i think. Created #3632

packages/components/switch/src/switch.ts
jpzwarte: Fixed

packages/components/switch/src/switch.ts
jpzwarte: Added the `@internal` jsdoc in the mean time.

packages/components/switch/src/switch.ts
jpzwarte: Fixed; try again after CI etc.

anna-lach (approved):

Issue #3598, opened 2026-08-14
[Tooltip] When disappearing Tooltip filckers on the anchor element in Safari


a11ymiko:
### 😯 Current Behavior

Before disappearing Tooltip changes position and displays on the anchor element for quick moment. Additionally when user hover the tooltip's anchor then press `esc` to dismiss tooltip and not change position of mouse cursor the tooltip reappears after a while. So user has no reliable option that allow to dismiss tooltip without changing mouse cursor position.

This happen for Toggle button, Toggle group and each story in Tooltip > Edge Case group.

This happens both on hover and on focus.

This happens only in Safari.

**Toggle button**

https://github.com/user-attachments/assets/5f0118a9-df23-4a29-a3a5-53ef4eb38ef1

**Toggle group**

https://github.com/user-attachments/assets/21622ebf-f9c9-436d-9818-03cbf3ab2622

**Edge case > Dialog**

https://github.com/user-attachments/assets/16e7c130-1b45-4f9b-8bda-88a37a727d18

**Edge case > Disabled buttons**

https://github.com/user-attachments/assets/921ddc35-688e-4dcd-bb10-3935fd35abbe

### 🤔 Expected Behavior/possible solution?

Tooltip should disappear without changing position. This works fine in Safari for [Button > Icon Only story](https://storybook.sanomalearning.design/?path=/story/actions-button--icon-only):

https://github.com/user-attachments/assets/ec8e1b48-347d-47d5-b781-852e7411fcd0

### 🖥️ Steps to Reproduce

1. Go to Toggle button story: https://storybook.sanomalearning.design/?path=/story/actions-toggle-button--basic
2. Verify is Tooltip changes position before disappearing.
3. Go to Toggle group story: https://storybook.sanomalearning.design/?path=/story/actions-toggle-group--basic
4. Verify is Tooltip changes position before disappearing.
5. Retake verification step for all stories in Tooltip > Edge Cases group

### Were you able to reproduce this in our storybook?

Yes

### If yes, provide the storybook URL

https://storybook.sanomalearning.design/?path=/story/actions-toggle-group--basic

### ⚖️ Impact

Just a visual issue / needs some polishing

### Framework

Other

### What browsers are you seeing the problem on?

Safari

### Component version

2.0.1

### 👤 Your name

Kacper

### 🧢 Your product/team

SLDS

### 🎨 Your Theme(s)

Sanoma Learning

anna-lach:
Please also make sure it works in Angular as well.

[...]

PR #3608, opened 2026-08-17
feat(tooltip): Angular `slTooltip` directive - add support for setting `type`

### LINKED ISSUE #2928: [Tooltip] It's not possible to set configuration options for the tooltip using Angular tooltip directive

Currently, it's not possible to set configuration options (like position, maxWidth etc.) for the tooltip using Angular tooltip directive. 

Please also take a look how it's solved in the Lit Tooltip directive. It's possible to set such configuration options there.

### 👤 Your name

Anna Sobczak

### 🧢 Your Product/Team

SLDS

anna-lach:
This PR adds support for specifying the tooltip `type` (such as `label` or `description`) via the `slTooltip` directive input in the Angular package. The directive now accepts either a string or an object with `text` and optional `type`, improving accessibility and flexibility. The Angular documentation and stories have been updated to reflect the new API and usage patterns.

**Angular Tooltip directive improvements:**

* The `slTooltip` directive now supports an object input `{ text, type }` in addition to a string, allowing tooltips to use different ARIA semantics (`label` or `description`). 

**Angular documentation and story updates:**

* Updated the tooltip documentation and storybook stories to demonstrate the new input format, usage examples, and global styling guidance for tooltips created by the directive.
* Clarified usage of the `for` and `type` attributes for tooltips in overlay stories, aligning with the new directive capabilities.
* Updated the tooltip usage example in `wrappers-overlay.stories.ts` to use the `for` attribute instead of `aria-describedby`, aligning with the latest recommended approach.


a11ymiko (commented):
Tooltip content is added as `aria-labelledby` to the button with text. This is not correct because text of button is not being announced by this. 

Tooltip content should be added as `aria-describedby`. and then screen reader would announce both text from the button (as button's label) and tooltip's content (as button's description).

This issue is present in Angular > Components > Tooltip as well as Angular > Wrappers > Tooltip.

Image

a11ymiko (commented):
Before tooltip disappears it changes position to display on top of button. This happens only in Safari. 
This is similar bug to: https://github.com/sl-design-system/components/issues/3598

https://github.com/user-attachments/assets/0a28529d-700b-41c6-897e-f751bb6e1212

a11ymiko (commented):
In Angular tooltip documentation first time `slTooltip` is mentioned it is styled as code with use. But in 'Directive' section it is styled as normal text:

Image

anna-lach:
> Before tooltip disappears it changes position to display on top of button. This happens only in Safari. This is similar bug to: #3598
> 
>  Screen.Recording.2026-08-18.at.08.24.53.mov

I think it would be the best to work on it in the https://github.com/sl-design-system/components/issues/3598 issue.

anna-lach:
> Tooltip content is added as `aria-labelledby` to the button with text. This is not correct because text of button is not being announced by this.
> 
> Tooltip content should be added as `aria-describedby`. and then screen reader would announce both text from the button (as button's label) and tooltip's content (as button's description).
> 
> This issue is present in Angular > Components > Tooltip as well as Angular > Wrappers > Tooltip.
> 
> Image

Good catch! It should be fine now.

Diaan (approved):

PR #3612, opened 2026-08-18
feat(shared): add ElementInternalsMixin and @cssState decorator

### LINKED ISSUE #3610: [Shared] Add `ElementInternalsMixin` and `@cssState` decorator

Currently we manually manage `this.attachInternals()` and `this.internals.states.add(...)`/`this.internals.states.delete`(...)`

This issue proposes 2 things:
1. `ElementInternalsMixin` that handles calling `attachInternals()` and accessing the `ElementInternals` instance through an `elementInternals` getter
2. `@cssState()` custom decorator that uses the `elementInternals` getter to set the css custom state

This helps reduce the amount of custom code we have for this.

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

jpzwarte:
## Summary

Adds an `ElementInternalsMixin` and a `@cssState` decorator to `@sl-design-system/shared`, uses them
across the components that work with `ElementInternals` and custom CSS states, and gives every
shared mixin its own entry point. Closes #3610.

- **`ElementInternalsMixin`** attaches the `ElementInternals` and exposes them as `elementInternals`,
  so the component, its subclasses, other mixins and its tests all have one standardized way of
  getting to them. It replaces the `readonly internals = this.attachInternals()` line every form
  associated element had. A mixin instead of a decorator, because a decorator cannot add a member to
  the type of the class in either the legacy or the standard mode; a mixin can.
- **`@cssState`** keeps a custom CSS state in sync with a boolean property, accessor or getter,
  replacing the hand written `if (changes.has('x')) { states.add/delete }` blocks in `updated()`.
  The state name defaults to the dasherized property name, and `{ invert: true }` sets the state
  while the property is falsy. It works with both the legacy (`experimentalDecorators`) and the
  standard TC39 decorators, and uses a single controller per element no matter how many properties
  it decorates. States are applied in `hostUpdate()`, so before render.

```ts
import { cssState } from '@sl-design-system/shared/decorators/css-state.js';
import { ElementInternalsMixin } from '@sl-design-system/shared/mixins/element-internals.js';

class MyElement extends ElementInternalsMixin(LitElement) {
  // Sets the `checked` state; style it with `my-element:state(checked)`
  @property({ type: Boolean }) @cssState() checked?: boolean;

  // Sets the `no-label` state while `hasLabel` is falsy
  @state() @cssState('no-label', { invert: true }) hasLabel = false;

  // A getter works as well, for a state derived from other properties
  @cssState('has-name')
  get hasName(): boolean {
    return this.hasLabel || this.hasAccessibleName();
  }
}
```

### Entry points

Every mixin now has its own entry point and lost the `-mixin` suffix in its file name, the same way
`ElementInternalsMixin` and `@cssState` do. The `@sl-design-system/shared/mixins.js` barrel has been
removed, and the mixins are no longer exported from `@sl-design-system/shared` either:

```ts
import { ForwardAriaMixin } from '@sl-design-system/shared/mixins/forward-aria.js';
import { LocaleMixin } from '@sl-design-system/shared/mixins/locale.js';
import { ObserveAttributesMixin } from '@sl-design-system/shared/mixins/observe-attributes.js';
```

The `Locale`, `ForwardAriaMixinInterface` and `ObserveAttributesMixinInterface` types moved along
with the mixin they belong to. `calendar`, `checkbox`, `combobox`, `format-date`, `format-number`,
`menu`, `number-field`, `select`, `text-area`, `text-field` and `tree` are updated to the new paths.

### Deprecated, not breaking

A Form Associated Custom Element should now expose its `ElementInternals` as `elementInternals`,
since that is what `FormControlMixin` reads the validity and form value from. The old name keeps
working, so nothing has to be updated in lockstep:

- `ElementInternalsMixin` also exposes the internals under a deprecated `internals` getter, so code
  that still reads `internals` on one of our components — tests, mostly — is unaffected.
- `FormControlMixin` falls back to `internals` when a form control element has no `elementInternals`,
  so a third party FACE that has not been updated yet keeps working too.

Both paths log a `console.warn` in development builds, once per element, pointing at
`elementInternals`. Support for `internals` will be removed in a future version. Because of this the
changesets are `minor` rather than `major`.

### Refactored

`ElementInternalsMixin` is applied to 15 components: `button`, `button-bar`, `checkbox-group`,
`combobox`, `date-field`, `editor`, `link`, `radio-group`, `select`, `select-button`, `switch`,
`tag`, `time-field`, `toggle-button` and `tool-bar`, plus the `child-form` example. It wraps
`LitElement` directly, so the mixins that use the internals sit above the one that provides them.

`@cssState` replaces the manual state juggling for 21 states in 10 components:

| Component | States |
| --- | --- |
| `button` | `icon-only` |
| `button-bar` | `empty`, `icon-only` |
| `date-field` | `has-focus`, `has-value`, `placeholder-shown` |
| `link` | `reversed` |
| `select-button` | `clearable`, `clear-focused`, `has-selection` |
| `switch` | `has-description`, `has-infotip`, `no-label` (inverted) |
| `tag` | `focus-visible` |
| `time-field` | `has-focus`, `has-value`, `placeholder-shown` |
| `toggle-button` | `error`, `icon-only`, `text-only` |
| `tool-bar` | `empty` |

The `toggle-button` conversion also removes a workaround: `render()` reads the `icon-only` state, so
`updated()` compared the old and new value to trigger an extra `requestUpdate()`. `@cssState`
applies in `hostUpdate()`, before render, so that extra render pass is gone.

Two things came out of the conversion that are more than a refactor:

- The `has-value` state on `sl-time-field` was documented and styled, but never actually set, so a
  time field with a value rendered it in the muted placeholder color instead of
  `--sl-color-foreground-plain`. It is now set from the `value` accessor, matching `sl-date-field`.
- The states that used to be added and removed by hand — `has-focus` in the date and time field,
  `focus-visible` in the tag, `empty` and `icon-only` in the button bar and tool bar — are now
  applied when the component updates rather than synchronously. That is still before the browser
  paints, but tests have to await the update.

Only `measuring` in `tool-bar/overflow.ts` is still set by hand; it is added and removed within a
single measurement pass, so there is no property to hang a state on.

### Reflected attributes kept

Two states that earlier commits on this branch had turned into custom states are reflected
attributes again, because they are part of the public API of the component:

- `[pressed]` on `sl-toggle-button` still reflects the `pressed` property, so `sl-toggle-group`
  keeps targeting `::slotted(sl-toggle-button[pressed])`.
- `[checked]` on `sl-switch` is reflected again.

The `clean-chicken-look`, `quiet-moons-tap` and `tame-spiders-invite` changesets are updated to
match; `[icon-only]`, `[text-only]` and `[error]` on the toggle button are still replaced by custom
states.


jpzwarte:
@claude review

a11ymiko (approved):

packages/components/time-field/src/time-field.ts
```diff
@@ -174,8 +175,11 @@ export class TimeField extends LocaleMixin(FormControlMixin(ScopedElementsMixin(
    */
   @property() placeholder?: string;
 
+  /** @internal Whether the time field has focus. */
+  @state() @cssState() hasFocus?: boolean;
```
michal-sanoma, line 179: **todo:** Small note: The PR description states that `has-focus` in `time-field/date-field`, `button-bar`, and `tag` were left untouched, but they are now migrated in the code. Let's update the PR summary to match the commits

packages/components/time-field/src/time-field.ts
```diff
@@ -237,6 +241,7 @@ export class TimeField extends LocaleMixin(FormControlMixin(ScopedElementsMixin(
   }
 
   @property()
+  @cssState('has-value')
```
michal-sanoma, line 244: **question:** shouldn't this be attached to the getter?

packages/components/time-field/src/time-field.ts
jpzwarte: Fixed

packages/components/time-field/src/time-field.ts
jpzwarte: I intially had that, but Angular has trouble with decorators on both the getter and setter. Had to change it to this, otherwise the angular compiler wouldn't succeed.

michal-sanoma (approved):

PR #3614, opened 2026-08-18
feat(popover): position with CSS anchor positioning

### LINKED ISSUE #3602: [Popover] Refactor to use CSS Anchor Positioning

jpzwarte:
Closes #3602

Replace the floating-ui based positioning with CSS anchor positioning. The browser now keeps the
popover in place and flips it to the opposite side when there is not enough room, so the per-frame
repositioning loop is gone. `sl-infotip` is rebuilt on top of that, and the components that slot an
infotip into their label are updated to match.

> [!NOTE]
> CSS Anchor Positioning is not yet supported in all browsers. You may need to include the
> [CSS Anchor Positioning polyfill](https://anchor-positioning.oddbird.net/) in your application.

## `sl-popover` (major)

The side the popover ends up on is now decided in CSS rather than in JavaScript, and the anchor is
the element that invoked it.

- The `position` property is **removed**, along with the twelve `top`/`right`/`bottom`/`left` values
  and their `-start` and `-end` variants. Set `position-area` and `position-try-fallbacks` on the
  popover instead; the default is `bottom span-all`, falling back to `flip-block`, `flip-inline` and
  `flip-block flip-inline`.
- The `anchor` attribute and the `anchorElement` property are **removed** with the `AnchorController`
  that backed them. A popover anchors itself to the element that invoked it, either through the
  [Invoker Commands API](https://developer.mozilla.org/en-US/docs/Web/API/Invoker_Commands_API) or
  through the `source` passed to `showPopover()` / `togglePopover()`. The anchor can also be set up
  entirely from CSS with `anchor-name` / `position-anchor`, which is what `sl-infotip` uses to
  anchor its popover to a trigger in its own shadow root. An `anchor-name` that is already set is
  kept — the popover appends its own name rather than replacing it.
- The `[actual-placement]` and `[anchored]` attributes are replaced by the `anchored-top`,
  `anchored-right`, `anchored-bottom` and `anchored-left` custom states. The side is also readable
  from JavaScript as the `placement` property, and a `PopoverPlacement` type is exported.
- The `Popover.offset`, `Popover.arrowPadding` and `Popover.viewportMargin` statics are **removed**.
  The gap between anchor and popover is the `--_offset` custom property (default `12px`); the space
  the arrow needs is reserved on top of it, on the side facing the anchor, so the arrow no longer
  sits in that gap. The distance to the viewport is the popover's `margin`, and there is no arrow
  padding any more — the arrow is positioned off the center of the anchor and keeps pointing at it
  when the browser pushes the popover back into view.
- `--sl-popover-max-inline-size` is **removed**; set `max-inline-size` directly. It no longer
  defaults to `80vw`.
- A `wrapper` part was added around the container and the arrow, so `container` now wraps only the
  slotted content.

The popover no longer slides along its anchor to stay in view; it flips instead.

## `sl-infotip` (minor)

`sl-infotip` renders its own `<button>` instead of an `sl-button` and opens the popover through the
Invoker Commands API, which drops the dependency on `@sl-design-system/button` along with the click
and keydown handling that went with it.

- The `size` property is **removed**. The trigger is no longer a design system button, so it has no
  sizes to pick from; it sizes itself to its icon. Use `--sl-icon-size` on the infotip to change it.
- The `toggleInfotip()` method is **removed**; the browser toggles the popover from the invoker.
- Focus is delegated to the internal `<button>` via `delegatesFocus: true`, replacing the `focus()`
  override. Calling `focus()` on the infotip still focuses the button.
- The popover is positioned by `sl-popover` with CSS anchor positioning instead of the `anchor` and
  `position` attributes it used to be given, so it opens below the button and flips when there is
  not enough room, rather than always opening above it.
- The popover content is capped at `min(80dvw, 400px)`, the icon uses
  `--sl-color-foreground-subtlest`, and the button has its own focus ring.

## Components that slot an infotip

`sl-checkbox`, `sl-radio` and `sl-switch` used to force `size="sm"` onto a slotted `sl-infotip` to
line its icon up with the label text. That property is gone, so the alignment is done in CSS
instead: the infotip is nudged up by `--sl-size-025` in the small checkbox, and down by
`--sl-size-100` in the large checkbox and the large radio. The `size="sm"` attribute is also dropped
from the infotip examples on the website.

---

Supersedes #3603, which GitHub closed when the branch was renamed from
`feat/3602-popover-refactoring` to `feature/3602-popover-refactoring`.


a11ymiko (commented):
For Checkbox and Checkbox group infotip can't be expanded using `spacebar` or `enter`. This problem exist for sighted keyboard users and screen reader users.

https://github.com/user-attachments/assets/37c60bdb-5b35-4a6f-b526-6ef1cbedb6c8

a11ymiko (commented):
Popover in Theme project in storybook is not placed correctly. Because of that some popover's are hidden by others:

https://www.chromatic.com/test?appId=6475dc9afc6964aa1e49c8df&id=6a8e8ffb6b45dd6abd0b22dd

Image

a11ymiko (commented):
I've noticed color change for infotip icon from #222222 (possibly `--sl-color-foreground-secondary-bold`?) to #59636e (`--sl-color-foreground-subtlest`).

In Figma icon color is still #222222 with token `color/foreground/secondary/bold`.

This probably should be unified between Figma and Storybook, right?
There is also difference in color for dark mode.

Storybook:

Image

Figma:

Image

a11ymiko (commented):
On documentation website in Firefox and Safari infotip is not displayed at the right place: 

https://github.com/user-attachments/assets/fb7d0bf8-1428-482a-8b70-81fe6fee88e6

https://github.com/user-attachments/assets/af1c8b7f-db2f-4183-aae1-5442cafb9370

a11ymiko (commented):
In Firefox both VoiceOver and NVDA does not announce `aria-expanded` of button. User doesn't get information about state of infotip text.


https://github.com/user-attachments/assets/44c32957-e9b0-41b7-a9cc-76e2b6f4a50c

jpzwarte:
This need a currently missing feature from the CSS Anchor Positioning polyfill, mainly supporting `this.style.positionAnchor` and `this.style.anchorName` access. There's a WIP pull request for this, but the maintainer is travelling for conferences atm. Best to merge this after the polyfill PR has been merged.

PR #3634, opened 2026-08-20
chore: migrate angular build to use vite instead of webpack

### LINKED ISSUE #3633: [Maintenance] Upgrade Angular to use vite instead of webpack

Turns out the tsdown migration "broke" something in the Angular webpack build: webpack *silently* failed to compile the .ts files. Instead it exported a blank `{}` object. #3604 includes a workaround for that. But a proper fix would be to migrate from webpack to vite.

### 👤 Your name

Jeroen

### 🧢 Your Product/Team

SLDS

jpzwarte:
Closes #3633

Migrates the Angular Storybook from the webpack-based `@storybook/angular` to the Vite-based `@storybook/angular-vite`, replacing the workaround added in #3604.

## Why

The tsdown migration (#3571) repointed every package's `exports` at TypeScript sources (`./src/*.ts`). The Angular webpack build cannot compile those from outside its own project, and instead of failing it emitted **empty modules** — every `register.ts` in the build came out as `(){}`, so `customElements.define()` never ran and no `sl-*` element was registered. `@sl-design-system/locales` was hit the same way, which is what surfaced the visible `Error: Invalid locale code`.

Vite turns this class of failure from silent into loud. While working on this PR, an intermediate attempt that resolved packages incorrectly failed the build with 48 named `MISSING_EXPORT` errors pointing at exact files and lines — precisely the feedback webpack never gave.

## Changes

**Dependencies**
- `@storybook/angular` → `@storybook/angular-vite`, plus `@analogjs/vite-plugin-angular` and `vite`
- All `@storybook/*` packages bumped 10.5.7 → 10.5.10 (the new framework's peer floor). The existing `^10.5.7` ranges already allowed this, so only the lockfile and the resolved versions changed.

**`packages/angular/angular.json`**
- Both targets repointed at `@storybook/angular-vite:start-storybook` / `:build-storybook`
- `browserTarget` removed (it pointed at itself and is not used by the new builder)
- `experimentalZoneless` → `zoneless`, following the builder schema

**`packages/angular/.storybook/main.ts`**
- `framework` switched to `@storybook/angular-vite`
- `webpackFinal` → `viteFinal`

**Stories (25 files) + `preview.ts`**
- Renderer imports (`Meta`, `StoryObj`, `StoryFn`, `moduleMetadata`, `applicationConfig`, `Preview`) repointed to `@storybook/angular-vite`, which re-exports the full API.

## Two migration details worth knowing

**`compodoc` is no longer read from `angular.json`.** The new preset reads it from `framework.options` and defaults to *on*. The existing `compodoc: false` in `angular.json` was silently ineffective, and the build shelled out to a compodoc binary that isn't installed. It is now set in `main.ts`.

**Packages are still aliased to `dist`, for a different reason than before.** Vite does not map tsconfig `paths` (webpack got that free via `tsconfig-paths-webpack-plugin`, which is what made `@sl-design-system/angular` → `./dist` work). It also only rewrites the TS `.js`-extension convention the sources import each other with (`export * from './avatar.js'` where the file is `avatar.ts`) for files it treats as project source — these arrive through node_modules symlinks outside the Angular project root. So `viteFinal` aliases every design system package to its built output, which is how an application consumes them anyway.

## Verification

Full build via `yarn workspace @sl-design-system/angular run storybook`, with the output bundle inspected:

| | webpack on `main` (with the #3604 workaround) | vite (this PR) |
|---|---|---|
| build time, like-for-like | 82s | **38.8s** |
| `sl-*` elements registered | 71 | **71** |
| stories indexed | 131 | **131** |
| build errors | 0 | **0** |

For reference, webpack *without* the #3604 workaround registered **0** elements — that is the failure this migration removes the need to work around.

Locales resolve correctly: `sourceLocale: 'en'`, `targetLocales: ['es-ES', 'it', 'nl', 'pl']`.

## Reviewer notes

- `@storybook/angular-vite` is currently in **preview** — Storybook's docs note that "APIs and defaults may change based on feedback before it is marked stable."
- Verification was done against build output and bundle contents, not by clicking through a running Storybook. A manual pass over the deploy preview is worth doing before merge.
- Requires Angular ≥ 21 and Vite ≥ 8; this repo is on Angular 22.1 and Vite 8.2, both within the framework's supported peer ranges.


a11ymiko (commented):
For [Utility > Format Date](https://storybook-3634.d3c5h2uor7jlkj.amplifyapp.com/angular/index.html?path=/story/wrappers-utility--format-date) there is error:

Image

But this is also on our production, so probably it's not in scope of this PR?

a11ymiko (commented):
[Utility > Format Number](https://storybook-3634.d3c5h2uor7jlkj.amplifyapp.com/?path=/story/angular_wrappers-utility--format-number) is empty.

In console there is this error:
"SB_PREVIEW_API_0009 (NoStoryMatchError): Couldn't find story matching 'wrappers-utility--format-number'.

- Are you sure a story with that id exists?
- Please check your stories field of your main.js config.
- Also check the browser console and terminal for error messages."

But this is also on our production, so probably it's not in scope of this PR?

a11ymiko (commented):
Storybook works well, but website has issue that does not allow to expand categories in menu. Instead of expanding categories it redirects to the page on which user is already. 

https://github.com/user-attachments/assets/175b0d4b-0749-4a3c-a202-f5f46e651326

a11ymiko:
> Member

this should be separate issue, as this PR doesn't touch website

a11ymiko (approved):

anna-lach:
> > Member
> 
> this should be separate issue, as this PR doesn't touch website

Not directly, but it looks like sth on this branch breaks it, because current version of the website works well https://sanomalearning.design/ and when we will marge it as it is on this branch it will be broken.

jpzwarte:
> Not directly, but it looks like sth on this branch breaks it, because current version of the website works well https://sanomalearning.design/ and when we will marge it as it is on this branch it will be broken.

Fixed by removing `minify-html-literals`: it failed on `virtual-list.ts`. The difference in actual size is negligible.

PR #3650, opened 2026-08-21
fix(paginator): correct mobile select aria label

### LINKED ISSUE #3629: [Paginator] Mobile version of paginator has wrong value of `aria-label`
### 😯 Current Behavior

Select used to change pages in mobile version of paginator has `aria-label` that says `X, page`, where X is current page number - 1.

Because of that, when user place focus on paginator with 1 page selected screen reader announces this element as: '1, 0, page, listbox, pop up collapsed, combo box'. This may be confusing for users.

https://github.com/user-attachments/assets/b2701799-c134-4115-8dbe-98f354a9b73b

Image

`aria-label` for that select should be changed so it no longer include 'number of current page -1'.

`aria-label='Page'` will be sufficient. Then paginator will be announced as '1, Page, listbox, pop up collapsed, combo box'.

### 🖥️ Steps to Reproduce

1. Go to https://storybook.sanomalearning.design/?path=/story/navigation-paginator-paginator--mobile
2. With screen reader turned on place keyboard focus on select
3. Verify if screen reader announce any other number than current page number

Yes

https://storybook.sanomalearning.design/?path=/story/navigation-paginator--connected

_No response_

### ⚖️ Impact

### Framework

Other

All

### If other, please specify.

_No response_

_No response_

### Component version

0.2.8

### Theme version

_No response_

### 👤 Your name

Kacper

### 🧢 Your product/team

SLDS

### 🎨 Your Theme(s)

Sanoma Learning

michal-sanoma:
## Summary

Fixes the mobile paginator select accessible name.

The paginator stores `page` internally as a zero-based index, while the displayed page number is one-based. The mobile select previously used the internal `page` value in its `aria-label`, so screen readers could announce values like `1, page` while the visible selected page was `2`.

This changes the mobile select label to `Page`, leaving the selected option text to provide the current page number

### BEFORE
before

### AFTER
after


a11ymiko (approved):

packages/components/paginator/src/paginator.ts
```diff
@@ -305,7 +305,7 @@ export class Paginator<T = any> extends ScopedElementsMixin(LitElement) {
         <sl-select
           @sl-change=${this.#onChange}
           .value=${this.page}
-          aria-label=${`${msg(str`${this.page}, page`, { id: 'sl.paginator.currentPage' })}`}
+          aria-label=${msg('Page', { id: 'sl.paginator.currentPage' })}
```
Diaan: I think the key doesn't match the actual string anymore.
Why is it "current" page?

Diaan (approved):

packages/components/paginator/src/paginator.ts
michal-sanoma: good point, changed to just `page`

PR #3660, opened 2026-08-25
feat(form): add on blur validation

### LINKED ISSUE #3113: [Form] Add on blur validation

Add on blur validation for form fields so that when a user leaves a field after entering a value, the input is checked, and an error message is shown if the value is invalid.

The request comes from Magister (Floor)

### 🤔 Expected Behaviour?

When a user **enters a value and leaves the field**, the field should validate the input if the value is already invalid. If the value is invalid, the validation message should appear immediately.

Validation on submit should still handle empty required fields, fields with dependencies on other fields, and validations that require a database or server check.

### 😯 Current Behaviour

Validation is currently focused on **form submission only**, in line with the SLDS vision. This means users may not receive feedback on invalid entries until they submit the form.

### 💁 Possible Solution

Add on-blur validation for fields where invalid input can be detected independently after the user leaves the field, without waiting for form submission.

Keep on submit validation for:
- empty required fields
- fields depending on other fields
- database or server-side validation

### 🔦 Context

A request was raised to extend the current validation approach. While SLDS UX vision supports validation on submit, there is a need to provide earlier feedback when a user has already entered a value but leaves the field with an invalid format or structure.

This would improve feedback timing and help users correct mistakes earlier in the interaction.

### 💻 Examples

- **Email field:** user types `nameexample.com` and leaves the field → show error message because `@` is missing.
- **Phone field:** user enters an incomplete or wrongly formatted value and leaves the field → show validation message on blur.
- **Password confirmation:** validation can remain on submit if it depends on another field.
- **Username availability:** validation can remain on submit if it requires a database check.

**To do**
- [ ] make sure to update all documentation + examples 

### 👤 Your name

Alvaro R.

### 🧢 Your Product/Team

SLDS

### LINKED ISSUE #3679: [Select] losing the focus clicking outside does not dispatch sl-blur
### 😯 Current Behavior

Actually `sl-blur` has stopped working on `sl-select`. Now if you click outside the focus the list is closed but `sl-blur` is not emitted I can still see focus-ring around `sl-select`. 
However if I have another component and I **TAB** or click on other component it loses the focus on the current `sl-select` and `sl-blur` is emitted.

Make the sl-select emit sl-blur event also when you click outside not to an element just on a blank space of the website.

### 🖥️ Steps to Reproduce

1. Create scenario with `sl-select` and few options.
2. Add event listener `sl-blur`.
3. Click outside in any blank empty space of the website.
4. Result: event not emitted

Yes

_No response_

```
<sl-select
        @sl-blur=${(e: Event) => {
          console.log('blur');
        }}
      >
   <sl-option>1</sl-option>
   <sl-option>2</sl-option>
      </sl-select>
<button>click </button>
```
Now click outside not on the button and `sl-blur` is not emitted.
If you open it again and TAB focus moves to button and it's emitted

### ⚖️ Impact

### Framework

Lit

Chrome

### If other, please specify.

_No response_

Windows 11

### Component version

2.2.0

### Theme version

_No response_

### 👤 Your name

Oscar Ledesma

### 🧢 Your product/team

BingelDC/Pokemon

### 🎨 Your Theme(s)

Bingel DC, Editorial Suite

anna-lach:
This PR adds new `validate-on-blur` option for SLDS forms (it makes possible to validate fields when they lose focus). It also adds Angular examples for both reactive and template-driven forms, updates checkbox and select blur behavior.

**Form Validation:**

* Added a new `validate-on-blur` option to `sl-form`, allowing fields to validate on blur (focus loss), with errors shown only after interaction. This is opt-in and does not affect default validation behavior.
* Updated `sl-select` to emit `sl-blur` after clearing and when clicking outside, ensuring consistent blur event handling.
* Improved `sl-checkbox` to handle pointer and keyboard focus more accurately, emit `blur` only when focus truly leaves.

**Angular Stories and Examples:**

* Added new Angular story components and stories (`AllFormControlsReactiveBlurComponent`, `AllFormControlsTemplateDrivenBlurComponent`) demonstrating `validate-on-blur` with both reactive and template-driven forms, including all form controls.
* Updated Angular documentation to explain how SLDS validation integrates with Angular forms and how `validate-on-blur` works with both form types.


a11ymiko (commented):
For optimal accessibility experience for keyboard and screen readers users we should modify on blur validation so it only gives information about errors in FILLED inputs or inputs that value was changed by user when they moves focus outside that input.

This will improve a11y experience because keyboard and screen readers users common behavior is to tab through all form fields when they encounter form for the first time. When they do this they will try to hear all the information that are needed to be filled in this form. Classic on blur validation makes that behavior harder because user not only will hear name and description of current form field, they will also hear information about error from previous empty field. That may be annoying and may be making getting to know new form more difficult for such users. 

That's why it's better for keyboard users to get on blur validation only for inputs that they really interacted with by changing their value or typing something inside them. User then will know to expect error message, and it'll not make their workflow more difficult.

a11ymiko (approved):

a11ymiko (commented):
For optimal UX for mouse users we decided to have validation that only displays error on required field when user focused the field, typed something inside, then deleted it and removed focus from that field. When user only place focus inside field, leaves it emptuy and moves focus avay no error message is displayed.

This behavior is aligned with NN group recommendations: "Clicking into a field and clicking out again is not always a meaningful attempt to complete it. It can happen while scanning the form, checking what is editable, or accidentally focusing a field. Showing an error at that point can feel too aggressive, especially for required fields where the user has not actually tried to provide an answer yet." https://www.nngroup.com/articles/hostile-error-messages

packages/components/checkbox/src/checkbox.ts
```diff
@@ -445,19 +447,40 @@ export class Checkbox<T = any> extends ForwardAriaMixin(
...
+      this.input.focus();
+    }
+
     this.checked = !this.checked;
     this.input.checked = this.checked;
     this.changeEvent.emit(this.formValue);
     this.updateState({ dirty: true });
     this.updateValidity();
+
+    const form = this.closest('sl-form');
```
michal-sanoma: **question:** Since we now have similar `this.closest('sl-form')` checks in `sl-checkbox`, `sl-checkbox-group`, and `sl-combobox`, I wonder if this logic might become difficult to maintain across all form controls over time?  

Maybe listening for `sl-change` inside `<sl-form>` can help here?

packages/components/checkbox/src/checkbox.ts
```diff
@@ -445,19 +447,40 @@ export class Checkbox<T = any> extends ForwardAriaMixin(
...
+
+    if (event instanceof MouseEvent && form?.validateOnBlur && this.required && !this.checked) {
```
michal-sanoma: **issue:** I see some difference in handling mouse vs keyboard for checkbox

if you uncheck a required checkbox with the mouse, the error appears immediately on click (before blur)

https://github.com/user-attachments/assets/2854effb-3dce-4115-a750-9576e1c44dfb

if you uncheck it with the keyboard (Space), it waits until you blur (Tab)

https://github.com/user-attachments/assets/9087d8b5-c205-468e-9c71-597df304c5dc

packages/components/select/src/select.ts
```diff
@@ -607,9 +624,56 @@ export class Select<T = any> extends ObserveAttributesMixin(
...
+    if (this.#blurEmitted) {
+      return;
+    }
+
+    this.#blurEmitted = true;
+    this.blurEvent.emit();
+    this.updateState({ touched: true });
+  }
+
+  #listenForOutsidePointer(): void {
```
michal-sanoma, line 642: **question:** I see that it validates just after clicking clear button, shouldn't it wait for the Tab?

https://github.com/user-attachments/assets/95fb18c0-65ef-4feb-993f-bffd4cb1cb86

packages/components/form/src/form.ts
```diff
@@ -215,10 +246,22 @@ export class Form<T extends Record<string, any> = Record<string, any>> extends L
...
+
     // Allow the control to unregister itself; this is necessary because by the
     // time `disconnectedCallback` is called, the control has already
     // been removed from the DOM; so any events emitted will never reach the form.
     event.detail.unregister = () => {
+      const onControlBlur = this.#controlBlurListeners.get(control);
+
+      if (onControlBlur) {
+        control.removeEventListener('sl-blur', onControlBlur);
+        this.#controlBlurListeners.delete(control);
```
michal-sanoma: **suggestion:** we probably should also add a cleanup in `disconnectedCallback()`

packages/components/form/src/form.ts
```diff
@@ -202,6 +215,24 @@ export class Form<T extends Record<string, any> = Record<string, any>> extends L
     return element instanceof HTMLElement && 'formControlElement' in element;
   }
 
+  #isEmptyValue(value: unknown): boolean {
+    return value == null || value === '' || (Array.isArray(value) && value.length === 0);
+  }
+
+  #validateControlOnBlur(control: HTMLElement & FormControl): void {
```
michal-sanoma, line 237: **issue:** not sure why but sometimes it validates `onChange` instead `onBlur`

https://github.com/user-attachments/assets/48695a15-a815-4baa-8a97-2012f6edab7e

packages/components/checkbox/src/checkbox.ts
anna-lach: Yes, this behaviour was discussed with @arecuenco-dsgn and @a11ymiko, can you please explain a bit more why it's done this way?

packages/components/select/src/select.ts
anna-lach: Yes, this behaviour was discussed with @arecuenco-dsgn and @a11ymiko, can you please explain a bit more why it's done this way?
Clicking a clear button we treat as an action, so in that case the invalid state should be shown right after an action, right @arecuenco-dsgn ?

packages/components/form/src/form.ts
anna-lach: If there is an error, for the first time it's validated on blur, but when the field was already checked and a user is making another error then it works on change (similar to report validity and after that it works on change).

packages/components/checkbox/src/checkbox.ts
a11ymiko: A11y wise it's wanted keyboard on blur behavior, as error message gets displayed when user moves to the next field. It's made this way, because it has higher chances oh being noticed by user when error message is automatically read on change of focus than on change of field value. 

When it comes to the difference between keyboard and mouse behavior I don't recall the argument for the mouse behavior, could you explain it @arecuenco-dsgn?

packages/components/select/src/select.ts
a11ymiko: User needs to move focus 'outside' select in order to use clear button, and it's also an action, because of that there it's valid to show error message after clicking clear button. Clicking clear button also moves focus to the select field, so when there is an error message already displayed this is a good opportunity to announce the message to the user. Because there is a difference in how `aria-describedby` is being announced on item that just received focus vs. on item that was focused before and is still focused when `aria-describedby` is added.

Using the select ant then removing selected option is not that easy to do all by mistake, so in this case it's a good thing to inform user about the error as soon as it's detected.

michal-sanoma: Thx for the clarification! If immediate validation on clear is the desired UX, we should probably update two places in this PR to reflect that:

1. In `packages/components/form/src/form.spec.ts`, there is currently a test expecting the opposite:
`it('should not validate a required select immediately after clearing')`

2. The story description in `select.stories.ts` currently states:
"Select an option and clear it, then leave the field to see the required validation on blur."

packages/components/form/src/form.ts
michal-sanoma: Ah, I see what you mean about switching to `onChange` after an error is reported! But in this video, this was actually the very first time typing into the field (it never had an error before)

The reason it switched to `onChange` immediately is because I tabbed through the empty field earlier while navigating the form

packages/components/checkbox/src/checkbox.ts
arecuenco-dsgn: With mouse interaction, the user can select and then deselect a required checkbox with clicks. After deselecting, the user can immediately move to another field, so showing the inline validation message makes sense once the control is left unchecked/invalid.

With keyboard interaction, the user reaches the checkbox with Tab and can toggle it with `space` or `enter`, but focus remains on the checkbox. In that case, we should not show the validation message immediately, because the user is still inside the control and may continue interacting with it.

Expected behaviour:
- Mouse: show the inline message once the user unchecks the required control and it's unchecked/invalid.
- Keyboard: do not show the inline message while focus remains on the control.
- Keyboard: show the inline message only after the user tabs away and the control is still unchecked/invalid.

So the rule stays consistent: show validation when the field is left invalid, not while the user is still focused on it.

packages/components/form/src/form.ts
anna-lach: Hmm ok, this way, looks like a bug then, I'll look into it, thanks.

packages/components/checkbox/src/checkbox.ts
michal-sanoma: Thanks @arecuenco-dsgn!

Good point, but right now mouse click shows the error immediately upon unchecking, while the user is still actively focused on the checkbox (before leaving it)

if we remove the immediate mouse validation check and let both mouse and keyboard validate on blur, it will actually achieve the exact behavior you described: showing the message only once the user leaves the field

packages/components/form/src/form.ts
anna-lach: Should be fixed now..

michal-sanoma (approved):

a11ymiko (commented):
Complete list of rules for interactions that triggers `on blur` validation on each components: https://sanoma-learning.atlassian.net/wiki/spaces/SUD/pages/1900609746/Interaction+-+Form+Controls+On-Blur+Validation

a11ymiko (approved):
Validation is aligned with rules for both mouse users and keyboard users.

PR #3672, opened 2026-08-28
fix(menu): make menu items with `aria-disabled` focusable

### LINKED ISSUE #3621: [Menu] Make disabled options focusable

Right now menu options use both `disabled` and `aria-disabled` attributes for disabling option for keyboard and mouse. Options disabled with this method are not accessible via keyboard with standard navigation.  

It's better to use only `aria-disabled` and make disabled options accessible via keyboard. Disabled options still won't be possible to be activated by mouse or keyboard but they will be focusable with standard keyboard navigation. Then user of assistive technologies could get information about every option in menu much easier. 

This approach is recommended by W3C: https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_disabled_controls

### 👤 Your name

Kacper

### 🧢 Your Product/Team

SLDS

anna-lach:
Menu item with `aria-disabled="true"` is now focusable.

Breaking changes:

- `disabled` property for `sl-menu-item` has been removed. Use `aria-disabled="true"` instead to disable a menu item.


a11ymiko (approved):

packages/components/menu/src/menu-button.spec.ts
```diff
@@ -35,10 +35,6 @@ describe('sl-menu-button', () => {
       menu = el.renderRoot.querySelector('sl-menu') as Menu;
     });
 
-    it('should not be disabled', () => {
```
michal-sanoma, line 38: **question:** it looks like this test was accidentally deleted. The `disabled` property was removed from `sl-menu-item`, but `sl-menu-button` still supports the `disabled` property, right?

packages/components/tabs/src/tab-group.ts
```diff
@@ -315,7 +315,7 @@ export class TabGroup extends ScopedElementsMixin(LitElement) {
                       menuItem => html`
                         <sl-menu-item
                           @click=${() => this.#onMenuItemClick(menuItem.tab)}
-                          ?disabled=${menuItem.disabled}>
+                          aria-disabled=${menuItem.disabled ? 'true' : 'false'}>
```
michal-sanoma: **suggestion:** minor consistency suggestion: in `toolbar` we use `ifDefined(isDisabled ? 'true' : undefined)` so the attribute is omitted when not disabled instead of outputting `aria-disabled="false"`, how about doing the same here? Unless you want this `aria-disabled=false` explicitly here

michal-sanoma (approved):

packages/components/menu/src/menu-button.spec.ts
anna-lach: Good catch! :)

packages/components/tabs/src/tab-group.ts
anna-lach: Good suggestion, thanks, I've changed it.

PR #3683, opened 2026-09-02
fix(form): announce validation messages when it's shown

### LINKED ISSUE #3680: [sl-form] Error messages should use sl-announcer
### 😯 Current Behavior

When screen reader user validate form field either using on-submit or on-blur validation the error message is not automatically announced by SR without going back to invalid field.

That makes filling the form more difficult, as user has to go back to each field in order to check if there are any errors.

`sl-error` content should be automatically announced by screen readers when error message gets displayed in form.

Probably the easiest way for that is to use `sl-announcer` and add there all the error messages when their appear? This should work smoothly for small number of errors but should be also tested on [Form - All](https://storybook.sanomalearning.design/?path=/story/form-form--all) when all the errors are submitted at once. Expected behavior is that user gets to hear all the error messages at once if keyboard focus stays on submit button.

### 🖥️ Steps to Reproduce

1. With screen reader turned on go to: https://storybook.sanomalearning.design/?path=/story/form-form--all
2. Submit empty form
3. Verify if screen reader automatically reads all the error messages without moving keyboard focus from submit button.

Yes

https://storybook.sanomalearning.design/?path=/story/form-form--all

_No response_

### ⚖️ Impact

### Framework

Lit

All

### If other, please specify.

_No response_

_No response_

### Component version

1.4.2

### Theme version

_No response_

### 👤 Your name

Kacper

### 🧢 Your product/team

SLDS

### 🎨 Your Theme(s)

Sanoma Learning

anna-lach:
Add `announce-errors` to `<sl-form>` to control validation announcements.

The `announce-errors` is enabled by default. Set `announce-errors="false"` when using aggregated error messaging (for example with `<sl-inline-message>`) to avoid duplicate screen reader announcements.


a11ymiko (commented):
For macOS Firefox for Form All when all errors are displayed at once user hears 'Hint text Please choose an option from the list' before all the error messages are announced. This seem to be hint and error text for the Select, that is being announced. If there is no error for Select, then the similar text but for Combobox Multiple gets announced. And if there is also no error then Combobox Single gets its hint and error message announced twice.

There should be only announcement of each error, without this 'Hint text Please choose an option from the list' or other text at the beginning.

https://github.com/user-attachments/assets/c93618b6-0d6b-4457-9d2a-fd487156d9b8

a11ymiko (commented):
Right now there is a timing issue that interfere with VoiceOver + Safari and NVDA + Edge/Firefox. 
Because there is no delay between moving focus away from invalid field and adding error message to `sl-announcer`:
1. Safari struggles to read error messages at all. Most of the times error message is not automatically announced, because it's started to being announce in the time when focus is leaving the field, and it ends being announced when focus is placed on the next element. So there are only miliseconds when this error may be announced, and it's being interupted right after that.
2. NVDA reads the error announcement BEFORE the name of focused field. This could be treated as a good behavior, but it's not aligned with standard, frequently seen behavior when first focused field gets announcement, and after that the error message for previous field is announced. 

This could be fixed by adding small delay, e.g. 250ms, before error message is added to `sl-announcer`. This was tested locally, and it proved to work better on Safari and Windows browsers.

https://github.com/user-attachments/assets/b43813e8-d75e-4dd4-ba1b-534b241ce7fb

a11ymiko (commented):
For fields that are not Date field and Time field NVDA in `on blur` validation announces automatically: 
1. Name of field and error message AND
2. Hint text and visible, red error message
The 2. part of announcement is not important and makes this announcement too long without any reason.

The 2. part of announcement should be skipped, so when error occurs NVDA announces only "NAME: error message" without any additional info.

https://github.com/user-attachments/assets/6011541d-11fb-4425-b7d8-ce0280dc0a17

.changeset/heavy-bats-appear.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sl-design-system/form': minor
```
Diaan, line 2: Maybe explain what "field-level validation announcements" are. 
From this info it is not clear that the fact that screenreaders announce the error the moment it is shown on screen is newly added in this PR.

And for scenarios where you aggregate the errors (1 inline message with all the errors) you should turn of the  announcement otherwise you get duplicate notifications with screenreaders.

packages/angular/stories/generated/forms.mdx
```diff
@@ -42,10 +42,12 @@ So instead:
 
 - Just **tabbing through** a field shows **no error** (field info is heard without noise)
 - If you **type, clear, and then blur**, the error **shows** (user intended to interact with the field)
-- **Mouse actions** like unchecking show error **immediately** (clear user intent to change the value)
+- **Mouse actions** like unchecking or clearing a value show the error when the control **loses focus**
 
 This way, keyboard/screen reader users can freely tab through a form to learn its structure, and only receive error feedback when they've actually modified a field. Untouched required fields still validate on form submit.
 
+By default, validation messages are announced by the live-region announcer. If your form also renders an aggregated error message (via `<sl-inline-message>`), set `announce-errors="false"` on `<sl-form>` to avoid duplicate screen reader announcements.
```
Diaan:
```suggestion
By default, validation messages are announced by the live-region announcer. If your form renders an aggregated error message (via `<sl-inline-message>`), set `announce-errors="false"` on `<sl-form>` to avoid duplicate screen reader announcements.
```

Why `also`? You can't have both, right? This describes the scenario with an aggregated error.

a11ymiko (approved):

Diaan (approved):

a11ymiko (approved):

Diaan (approved):

PR #3684, opened 2026-09-03
fix(search-field): make clear text button keyboard accessible

### LINKED ISSUE #3132: [Search field] "Clear text" button is not keyboard accessible
## Current behavior 

Keyboard user can place their focus only inside search `input` but cannot place it on "Clear text" `button` because it has `tabindex=-1`. Screen reader user can place screen reader's focus on this button using arrows but not using `Tab`. 
This behavior is mimicking `input type='search'` in which "Clear" `button` cannot be reached via keyboard nor screen reader. But this button is not presented in some browsers like Firefox and its behavior is not considered accessible: https://design-patterns.service.justice.gov.uk/components/search/#accessibility-issues  

Image

## Suggested solution

It's recommended to change `tabindex` value from `-1` to `0` and make sure focus indicator is visible and it has sufficient color ratio of at least 3:1 between background.  That should make button fully accessible for keyboard users.

### 👤 Your name

Kacper

### 🧢 Your Product/Team

SLDS

anna-lach:
Fixes #3132


a11ymiko (commented):
Maybe we should add update to our documentation website about clear button being focusable and why we decided to do that? 

I can prepare that content for Accessibility tab if you think that is useful 😉

a11ymiko (commented):
Focus indicator on clear button in pill shaped search field looks like it's ,oved to the left and is not aligned with button: 

Image

For other buttons in pill shaped fields focus indicator looks like: 

Image

a11ymiko (commented):
For large size of serach field focus indicator on clear button is not as high as form field:

Image

a11ymiko (commented):
For pill shaped large search field focus indicator on clear button is too small:

Image

anna-lach:
> Maybe we should add update to our documentation website about clear button being focusable and why we decided to do that?
> 
> I can prepare that content for Accessibility tab if you think that is useful 😉

@a11ymiko  Good point! That would be helpful, thanks. 🙂

a11ymiko (approved):

.changeset/nine-ideas-jump.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sl-design-system/search-field': patch
+---
+
+Improve clear-button accessibility in `sl-search-field`.
+
+Make the clear button focusable, allow clearing with Enter/Space when focused.
```
michal-sanoma: **suggestion:** I would mention here also `Escape`

website/src/categories/components/search-field/accessibility.md
```diff
@@ -17,17 +24,16 @@ Here's an overview of the common keyboard interactions associated with a search
 
 |Command|Description|
 |-|-|
-|Tab|When focus is outside the search field, moves focus to the input container. If focus is on the input container, moves focus to the end enhancer (if provided, and interactive).|
+|Tab|When focus is outside the search field, moves focus to the input container. If focus is on the empty input container, moves focus to the end enhancer (if provided, and interactive). If the input contains a value, moves focus to the clear button.|
 |Esc|Clear search value, when is an input.|
```
michal-sanoma: **suggestion:** not yr change but would be cool to improve this. Looks like a typo or misspelled sentence, how about `Clears the search value when focus is in the input or on the clear button, and returns focus to the input.`?

packages/components/search-field/src/search-field.spec.ts
```diff
@@ -135,6 +157,49 @@ describe('sl-search-field', () => {
...
+
+    it('should focus the input when Escape is pressed on the clear button', async () => {
+      const button = el.renderRoot.querySelector<HTMLButtonElement>('button');
+
+      button?.focus();
+      await userEvent.keyboard('{Escape}');
+      await el.updateComplete;
+
+      expect(el.value).to.equal('');
+      expect(document.activeElement).to.equal(el.querySelector('input'));
```
michal-sanoma, line 202: **suggestion:** I would propose to add the same check to `Space` and `Enter` tests

packages/components/search-field/src/search-field.spec.ts
```diff
@@ -60,20 +60,42 @@ describe('sl-search-field', () => {
       expect(button).to.contain('sl-icon[name="circle-xmark"]');
     });
 
+    it('should have a focusable clear button', () => {
```
michal-sanoma, line 63: **question:** since the root issue reported in #3132 was specifically about keyboard users pressing Tab to reach the clear button, could we add a test that starts from the input and tabs into the clear button using `userEvent.tab()`?

michal-sanoma (approved):

PR #3692, opened 2026-09-07
fix(grid): support custom header renderers in sort columns

### LINKED ISSUE #3682: [Grid - Sorting] Custom header renderer not possible on a SortColumn
### 😯 Current Behavior

The Storybook page of "Grid / Basics - Header" shows an example on how to pass a custom renderer function into a `sl-grid-column`. See https://github.com/sl-design-system/components/blob/9d4d49be891a01f1c36895fe816ae451689dfc33/packages/components/grid/src/stories/basics.stories.ts#L260-L268

```
      <sl-grid-column
        path="firstName"
        .header=${() => html`
          <span>First name</span>
          <sl-icon id="info-icon" name="info"></sl-icon>
          <sl-tooltip for="info-icon">Some information about the first name</sl-tooltip>
        `}
        .scopedElements=${{ 'sl-icon': Icon, 'sl-tooltip': Tooltip }}>
      </sl-grid-column>
```

One would expect `sl-grid-sort-column` to behave similair, but it does not. It will treat the passed renderer as a string.

Root cause is GridSortColumn overriding the default renderHeaderRow method of SortColumn, which not takes passing a function into account: https://github.com/sl-design-system/components/blob/9d4d49be891a01f1c36895fe816ae451689dfc33/packages/components/grid/src/sort-column.ts#L77-L100

I would expect this to work, similar to a normal sl-grid-column:

```
      <sl-grid-sort-column
        path="firstName"
        .header=${() => html`
          <span>First name</span>
          <sl-icon id="info-icon" name="info"></sl-icon>
          <sl-tooltip for="info-icon">Some information about the first name</sl-tooltip>
        `}
        .scopedElements=${{ 'sl-icon': Icon, 'sl-tooltip': Tooltip }}>
      </sl-grid-sort-column>
```

### 🖥️ Steps to Reproduce

See description above

No

_No response_

See above for code

### ⚖️ Impact

### Framework

Other

All

### If other, please specify.

All

_No response_

### Component version

0.7.5

### Theme version

_No response_

### 👤 Your name

Jorn Nolles

### 🧢 Your product/team

Magister

### 🎨 Your Theme(s)

_No response_

michal-sanoma:
## Summary

Fix custom header renderers in `sl-grid-sort-column`

Previously, `GridSortColumn` interpolated the `header` property directly. When a renderer function was provided, Lit treated the function as a value instead of invoking it

The sortable column now delegates header rendering to the inherited `renderHeaderLabel()` method, matching the behavior of `sl-grid-column`

### AFTER 
Screenshot 2026-09-07 at 13 41 01

### DIFF
--- a/packages/components/grid/src/stories/sorting.stories.ts
+++ b/packages/components/grid/src/stories/sorting.stories.ts
@@ -1,6 +1,8 @@
 import { Avatar } from '@sl-design-system/avatar';
 import { ArrayListDataSource } from '@sl-design-system/data-source';
 import { type Student, getStudents } from '@sl-design-system/example-data';
+import { Icon } from '@sl-design-system/icon';
+import { Tooltip } from '@sl-design-system/tooltip';
 import { type Meta, type StoryObj } from '@storybook/web-components-vite';
 import { html } from 'lit';
 import '.././register.js';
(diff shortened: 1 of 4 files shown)


a11ymiko (approved):

packages/components/grid/src/stories/sorting.stories.ts
```diff
@@ -42,6 +44,30 @@ export const Basic: Story = {
...
+      <code>header</code> property of <code>sl-grid-sort-column</code> to a renderer function. If
+      you use custom elements in the header, add them to the column's
+      <code>scopedElements</code> property.
+    </p>
+    <sl-grid .items=${students}>
+      <sl-grid-sort-column
+        path="firstName"
+        .header=${() => html`
+          <span>First name</span>
+          <sl-icon id="first-name-info" name="info"></sl-icon>
```
anna-lach: I know it's just an example. but maybe it would be better to use `sl-infotip`? Right now the icon is not focusable and tooltip is not available for keyboard users.

anna-lach (approved):

packages/components/grid/src/stories/sorting.stories.ts
michal-sanoma: good point, added

PR #3693, opened 2026-09-07
fix(grid): prevent unintended row dragging from cell content

### LINKED ISSUE #3667: [Grid - Basic] Mouse user can drag and drop each row in variants that shouldn't allow drag and drop
### 😯 Current Behavior

When mouse user click and hold on avatar image the row becomes draggable. Then user can drag this row to the end of grid and drop it below the last row (which then looks like a copy of dragged row).

This was recorded for [Grid - Basics Basic story](https://storybook.sanomalearning.design/?path=/story/grid-basics--basic) but can be reproduced in other Grid modules that are not Drag and Drop.

https://github.com/user-attachments/assets/d7826fc3-ef97-408d-90b6-eb253095d7cc

Only Drag and Drop module of Grid should allow to change positions of rows?

### 🖥️ Steps to Reproduce

1. Go to https://storybook.sanomalearning.design/?path=/story/grid-basics--basic
2. Click and hold on avatar image 
3. Drag the row to the end of grid
4. Drop the row below the last row of a grid
5. Verify if dragged row become the last one and if it's not visible in its first place

Yes

https://storybook.sanomalearning.design/?path=/story/grid-basics--basic

_No response_

### ⚖️ Impact

### Framework

Lit

All

### If other, please specify.

_No response_

_No response_

### Component version

0.7.5

### Theme version

_No response_

### 👤 Your name

Kacper

### 🧢 Your product/team

SLDS

### 🎨 Your Theme(s)

Sanoma Learning

michal-sanoma:
## Summary

Prevents native draggable elements inside grid cells, such as avatar images, from unintentionally starting row drag and drop

Row dragging now starts only when the current mouse or touch gesture originates from a drag handle. The per-gesture state is cleared on release, cancellation, window blur, or when the actual drag begins

This also prevents native draggable content from bypassing the guard after an unused drag-handle press and inside permanently draggable group rows

The fix additionally corrects restoring the first row when it is dragged outside the grid. Previously, `.at(-1)` incorrectly treated the last item as the first row’s previous sibling, causing the first row to move to the end

### BEFORE

https://github.com/user-attachments/assets/64df69d0-e76a-420b-8362-013b51359bf2

### AFTER

https://github.com/user-attachments/assets/fa05bf05-6474-4914-8a60-faeca31375f9

### DIFF
--- a/packages/components/grid/src/grid.spec.ts
+++ b/packages/components/grid/src/grid.spec.ts
@@ -1,6 +1,10 @@
 import { type Button } from '@sl-design-system/button';
 import '@sl-design-system/button/register.js';
-import { ArrayListDataSource, type ListDataSourceGroupItem } from '@sl-design-system/data-source';
+import {
+  ArrayListDataSource,
+  type ListDataSourceGroupItem,
+  isListDataSourceDataItem
+} from '@sl-design-system/data-source';
 import '@sl-design-system/menu/register.js';
 import { type ToolBar } from '@sl-design-system/tool-bar';
 import '@sl-design-system/tooltip/register.js';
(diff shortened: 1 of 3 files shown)


packages/components/grid/src/grid.ts
```diff
@@ -850,12 +850,19 @@ export class Grid<T = any> extends ScopedElementsMixin(LitElement) {
   };
 
   #onDragStart(event: DragEvent, item: ListDataSourceItem<T>): void {
+    const row = event.currentTarget as HTMLTableRowElement;
+
+    // Native draggable elements inside a cell, such as images, also emit drag events. Only start
+    // dragging a grid row after the drag handle has explicitly made the row draggable.
+    if (!row.draggable) {
```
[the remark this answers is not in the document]
michal-sanoma: fixed

a11ymiko (approved):

packages/components/grid/src/grid.spec.ts
```diff
@@ -184,6 +188,200 @@ describe('sl-grid', () => {
...
+          composed: true,
+          dataTransfer: new DataTransfer()
+        })
+      );
+
+      expect(groupRow).to.have.attribute('draggable', 'true');
+      expect(dragStartSpy).not.to.have.been.called;
+    });
+
+    it('should restore the first row when it is dragged outside the grid', async () => {
```
Diaan:
```suggestion
    it('should restore the dragged row to its original position when it is dragged outside the grid', async () => {
```

packages/components/grid/src/grid.spec.ts
```diff
@@ -184,6 +188,200 @@ describe('sl-grid', () => {
...
+            { firstName: 'Alice', lastName: 'Johnson' }
+          ]}>
+          <sl-grid-drag-handle-column></sl-grid-drag-handle-column>
+          <sl-grid-column path="firstName"></sl-grid-column>
+        </sl-grid>
+      `);
+
+      await waitForGridToRenderData(el);
+
+      const firstRow = el.renderRoot.querySelector<HTMLTableRowElement>('tbody tr')!,
```
Diaan: maybe rename this to `draggedRow` as well?

Diaan (approved):

packages/components/grid/src/grid.spec.ts
michal-sanoma: added

packages/components/grid/src/grid.spec.ts
michal-sanoma: good point, changed

PR #3709, opened 2026-09-10
fix(date-field): improve focus behavior in the calendar dialog after month/year selection

### LINKED ISSUE #3135: [Date field] Focus is placed on previous month's day if month is not starting from Monday
## Current behavior 

If keyboard user use change month view or change year view in calendar dialog their focus is placed on the first Monday in the calendar. 
Most of the time focus is placed on date in the month preceding the one being selected. 
That may be confusing because if user choose to go to June 2025 their focus is placed on July 30th 2025.

https://github.com/user-attachments/assets/2e1ea884-4ee6-47d8-ad20-d54453f95979

## Suggested solution

It's recommended to add mechanism that places focus on:

- first day of selected month if this day is active
- first selectable day of selected month if the first day of the month is disabled

### 👤 Your name

Kacper

### 🧢 Your Product/Team

SLDS

anna-lach:
Fixes #3135


packages/components/date-field/src/date-field.spec.ts
```diff
@@ -568,6 +571,147 @@ describe('sl-date-field', () => {
       expect(monthView?.shadowRoot?.activeElement).to.have.attribute('aria-current', 'date');
     });
 
+    it('should focus the first day of the selected month after choosing a month', async () => {
```
a11ymiko, line 574: Is it worth to add similar test but for the year instead of month?

a11ymiko (approved):

packages/components/date-field/src/date-field.ts
```diff
@@ -1036,6 +1049,59 @@ export class DateField extends LocaleMixin(
     }
   }
 
+  #focusFirstSelectableDayOfDisplayedMonth(): void {
```
michal-sanoma, line 1052: **question:** @a11ymiko  how this should behave? Should focus stay on 14th or come back to 1st in this flow?

https://github.com/user-attachments/assets/2cf10d84-d010-42be-b0af-67ca7b1cf7fc

packages/components/date-field/src/date-field.ts
```diff
@@ -1128,6 +1194,31 @@ export class DateField extends LocaleMixin(
     });
   }
 
+  async #syncCalendarFocusAfterModeSwitch(): Promise<void> {
```
michal-sanoma, line 1197: **question:** I see this also on main already but can we create an issue for this? Seems that sometimes u need to pick month twice to change it effectively

https://github.com/user-attachments/assets/e19f13eb-bb30-4e3b-a784-e6537e3c6f07

Diaan (approved):

packages/components/date-field/src/date-field.ts
a11ymiko: This should return focus to selected date in such case. Because then user will get info, that 14th is still selected, and this might get missed if we return focus to 1st instead.

packages/components/date-field/src/date-field.ts
a11ymiko: On it, I'll create issue for that 🫡 ✅

michal-sanoma (approved):

packages/components/date-field/src/date-field.spec.ts
anna-lach: Test added.

